PR #4, opened 2016-03-22
Problem: No travis support

betatim:
This adds basic support for travis and running tests with `nose`


betatim:
@MechCoder can you enable travis for the repo so we can see that this works?

MechCoder:
Thank you!

MechCoder:
Umm the build seems to fail. Will try to figure out

PR #5, opened 2016-03-23
Change output type to scipy.OptimizeResult type

MechCoder:
cc: @betatim @glouppe

### DIFF
--- a/skopt/gp_opt.py
+++ b/skopt/gp_opt.py
@@ -94,17 +94,18 @@ def gp_minimize(func, bounds=None, search="sampling", random_state=None,
 
     Returns
     -------
-    res: OptimizeResult
+    res: OptimizeResult, scipy object
         The optimization result returned as a OptimizeResult object.
         Important attributes are
         ``x`` - float, the optimization solution,
         ``fun`` - float, the value of the function at the optimum,
         ``models``- gp_models[i]. the prior on the function fit at
                        iteration[i].
-        ``func_vals`` - the minimum function value attained at the ith
-                        iteration.
-        ``x_iters`` - the value of ``x`` corresponding to the minimum
-                      function value at the ith iteration.
+        ``func_vals`` - the function value at the ith iteration.
+        ``x_iters`` - the value of ``x`` corresponding to the function value
+                      at the ith iteration.
+        For more details related to the OptimizeResult object, refer
+        http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
     """
     rng = np.random.RandomState(random_state)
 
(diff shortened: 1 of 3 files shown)


skopt/gp_opt.py
```diff
@@ -94,19 +94,17 @@ def gp_minimize(func, bounds=None, search="sampling", random_state=None,
...
-        d["func_vals]: d["func_vals"][i] corresponds to the minimum
-        function value in iteration i.
+    res: OptimizeResult
+        The optimization result returned as a OptimizeResult object.
+        Important attributes are
+        ``x`` - float, the optimization solution,
+        ``fun`` - float, the value of the function at the optimum,
+        ``models``- gp_models[i]. the prior on the function fit at
+                       iteration[i].
+        ``func_vals`` - the minimum function value attained at the ith
```
MechCoder: Should we just return the `function_value` at every iteration instead of the cumulative minimum?

skopt/gp_opt.py
```diff
@@ -157,8 +155,8 @@ def gp_minimize(func, bounds=None, search="sampling", random_state=None,
     func_ind = np.argmin(func_val)
     x_val = x[func_ind]
     best_func_val = func_val[func_ind]
-    d = {}
-    d["models"] = gp_models
+    res = OptimizeResult
```
glouppe: shouldnt this be `OptimizeResult()`?

skopt/gp_opt.py
glouppe: Yes, it may be best to instead collect the sampled `x` and their values. (In this way, the min at iteration `i` can be reconstructed)

skopt/gp_opt.py
```diff
@@ -94,19 +94,17 @@ def gp_minimize(func, bounds=None, search="sampling", random_state=None,
...
-        Function minimum.
-
-    d: dict
-        d["models"]: List of GP models of len maxiter. d["models"][i]
-        corresponds to the GP model fit in iteration i.
-        d["x_iters"]: d["x_iters"][i] corresponds to the "best" x in
-        iteration i.
-        d["func_vals]: d["func_vals"][i] corresponds to the minimum
-        function value in iteration i.
+    res: OptimizeResult
```
glouppe: Shall we also set a `success` and `status` flag?

skopt/gp_opt.py
```diff
@@ -94,19 +94,17 @@ def gp_minimize(func, bounds=None, search="sampling", random_state=None,
...
+        The optimization result returned as a OptimizeResult object.
+        Important attributes are
+        ``x`` - float, the optimization solution,
+        ``fun`` - float, the value of the function at the optimum,
+        ``models``- gp_models[i]. the prior on the function fit at
+                       iteration[i].
+        ``func_vals`` - the minimum function value attained at the ith
+                        iteration.
+        ``x_iters`` - the value of ``x`` corresponding to the minimum
+                      function value at the ith iteration.
```
glouppe: I would add a note pointing to https://github.com/scipy/scipy/blob/master/scipy/optimize/optimize.py#L75-L130

glouppe:
Besides my comments regarding documentation, +1 for merge! Thanks :)

skopt/gp_opt.py
MechCoder: I'm not sure this is necessary right now, since `n_iter > maxiter` is the only reason it can stop

MechCoder:
@glouppe Addressed your comments. I'll merge when Travis is green?

glouppe:
Yup, thanks :)

PR #9, opened 2016-03-23
[WIP] GBTRegressor does not provide uncertainty estimates

betatim:
Started a simple wrapper for regressors like GBTRegressor to provide uncertainty estatimates as well
as central predictions.

Any ideas on how to test that this indeed gives the central and 68% confidence interval (right terminology??) predictions? Best idea so far is to use a very large number of samples on a simple problem and check that the predicted standard deviation is approximately equal to the std used when sampling.

Todo:
* [x] documentation
* [ ] check that GP uses same definition of uncertainty
* [x] smarter tests

### DIFF
--- a/skopt/gbt.py
+++ b/skopt/gbt.py
@@ -7,11 +7,37 @@ from sklearn.utils import check_random_state
 
 class GBTQuantiles(BaseEstimator, RegressorMixin):
     def __init__(self, quantiles=[0.16, 0.5, 0.84], random_state=None):
+        """Predict several quantiles with one estimator
+
+        This is a wrapper around `GradientBoostingRegressor`'s quantile
+        regression that allows you to predict several `quantiles` in
+        one go.
+
+        Parameters
+        ----------
+        quantiles : array-like, optional
+            Quantiles to predict. By default the 16, 50 and 84%
+            quantiles are predicted.
+
+        random-state : int, RandomState instance, or None (default)
+            Set random state to something other than None for reproducible
+            results.
+        """
         self.quantiles = quantiles
         self.random_state = random_state
 
     def fit(self, X, y):
-        """Fit one regressor for each quantile"""
+        """Fit one regressor for each quantile.
+
+        Parameters
+        ----------
+        X : array-like, shape = [n_samples, n_features]
+            Training vectors, where n_samples is the number of samples
+            and n_features is the number of features.
+
+        y : array-like, shape = [n_samples]
+            Target values (real numbers in regression)
+        """
         rng = check_random_state(self.random_state)
         self.regressors_ = [GradientBoostingRegressor(loss='quantile',
                                                       alpha=a,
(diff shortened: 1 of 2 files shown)


skopt/tests/test_gbt.py
```diff
@@ -0,0 +1,35 @@
...
+
+def truth(X):
+    return 0.5 * np.sin(1.75*X[:, 0])
+def constant_noise(X):
+    return np.ones_like(X)
+def varying_noise(X):
+    # noise level depends on value of `X`
+    return np.abs(2.5-X)/2.5 + 0.1
+
+def sample_noise(X, std=0.2, noise=constant_noise):
```
betatim: should pass in `rng`/use `check_random_state`

skopt/tests/test_gbt.py
```diff
@@ -0,0 +1,38 @@
...
+
+rng = np.random.RandomState(324)
+
+def truth(X):
+    return 0.5 * np.sin(1.75*X[:, 0])
+
+def constant_noise(X):
+    return np.ones_like(X)
+
+def sample_noise(X, std=0.2, noise=constant_noise):
```
betatim: use `random_state` pattern instead of relying on global `rng`

glouppe:
I am not fond of the nomenclature, std != uncertainty != quantile.

betatim:
I'm flexible, but out of ideas at the moment for a good (short) name.

`GBTQuantileRegressor`?

`GBTQuantiles(quantiles=list_of_quantiles)` with `per_quantile_prediction = predict()`. You set all the quantiles you want estimated and we return each single one? In this case `GBTQuantiles([0.18, 0.5, 0.84])`

MechCoder:
I'm planning to review my GradientBoosting know-how, this weekend. Can have a look after that.

betatim:
If you need this merge it, I'm away till Monday night without internet

skopt/gbt.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def fit(self, X, y):
+        """Fit one regressor for each quantile"""
+        rng = check_random_state(self.random_state)
+        self.regressors_ = [GradientBoostingRegressor(loss='quantile',
+                                                      alpha=a,
+                                                      random_state=rng)
+                            for a in self.quantiles]
+        for rgr in self.regressors_:
+            rgr.fit(X, y)
```
MechCoder: return self

MechCoder:
@glouppe @betatim 
How do we compute the standard deviation of the predictions given the quantiles? There is no assumption about the conditional distribution of Y given X right?

betatim:
The last item on the to do list: "check definition of uncertainty" is something I'd like to punt to #23, instead of waiting to merge this till we converge.

MechCoder:
This looks ok to me.

betatim:
:chart_with_upwards_trend:

glouppe:
Would be good to expose the `base_estimator` in this newly introduced class. Default params of `GradientBoostingRegressor` are very likely to give poor results.

PR #10, opened 2016-03-23
Make gp_acquisition_func public

MechCoder:
A minor pull request to make `_acquistion_func` public. Also renamed it as `gp_acquisition_func` since the actual acquisition function depends on the distribution of the posterior

### DIFF
--- a/skopt/gp_opt.py
+++ b/skopt/gp_opt.py
@@ -8,16 +8,53 @@ from sklearn.gaussian_process.kernels import Matern
 from scipy import stats
 from scipy.optimize import OptimizeResult, fmin_l_bfgs_b
 
-def _acquisition_func(x0, gp, prev_best, func, xi=0.01, kappa=1.96):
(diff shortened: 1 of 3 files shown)


skopt/gp_opt.py
```diff
@@ -8,16 +8,53 @@
...
+    ----------
+    x0 : array-like
+        Values where the acquistion function should be computed.
+
+    gp: GaussianProcess model
+        The fit sklearn gaussian process estimator that approximates
+        the function.
+
+    prev_best: float, optional
+        The previousbest value over which we want to improve.
```
betatim: -> previous best

betatim:
Does it need to be `gp_...` it would be useful also for a tree based model instead of a GP. As long as both provide a central value and uncertainty.

MechCoder:
Addressed. Please merge if happy.

MechCoder:
@betatim Can I merge this? Sorry for rushing, but I'm refactoring some more code..

PR #12, opened 2016-03-24
Some cleanup

glouppe:
Nothing serious, just some code cleanup.

### DIFF
--- a/skopt/tests/test_gp_opt.py
+++ b/skopt/tests/test_gp_opt.py
@@ -1,63 +1,42 @@
-from math import pi, cos
-
 import numpy as np
+from math import pi, cos
 from sklearn.utils.testing import assert_less
 
 from skopt.gp_opt import gp_minimize
 
-a = 1
-b = 5.1 / (4 * pi**2)
-c = 5.0 / pi
-r = 6
-s = 10
-t = 1 / (8*pi)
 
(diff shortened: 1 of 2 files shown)


skopt/tests/test_gp_opt.py
```diff
@@ -1,63 +1,42 @@
...
-def hartmann_6(x):
+
+def branin(x, a=1, b=5.1 / (4 * pi ** 2), c=5. / pi,
+           r=6, s=10, t=1. / (8 * pi)):
+    return (a * (x[1] - b * x[0] ** 2 + c * x[0] - r) ** 2 +
+            s * (1 - t) * cos(x[0]) + s)
+
+
+def hartmann_6(x,
+               alpha=np.asarray([1.0, 1.2, 3.0, 3.2]),
```
MechCoder: I"m not sure it is safe to use mutables as default arguments, but in this case it is harmless..

MechCoder:
Thanks!

skopt/tests/test_gp_opt.py
glouppe: Yeah, here it does not matter. I was also thinking maybe we should move these functions in a `benchmarks` submodule, à la `sklearn.datasets`? what do you think?

PR #20, opened 2016-03-27
Fix EI?

MechCoder:


MechCoder:
@glouppe I removed the zeros handling and allowed SciPy handling, but the weirdness remains. Would be great if you could cross-check

MechCoder:
I tried the function here (https://github.com/jmetzen/bayesian_optimization/blob/master/bayesian_optimization/acquisition_functions.py#L109) but I get a very similar graph.

figure_1-3

glouppe:
Thanks! Indeed you were right, EI is in fact very peaky, hence the observed figure. Nothing was wrong.

Issue #26, opened 2016-03-29
Expose `base_estimator` in `GBTQuantiles`


betatim:
Comment by @glouppe in #9: 

Would be good to expose the `base_estimator` in this newly introduced class. Default params of `GradientBoostingRegressor` are very likely to give poor results.

glouppe:
Also, not fan of `GBT` (sorry for being picky). Let's be explicit instead. What about `GradientBoostingQuantileRegressor`? or `GradientBoostingQuantiler`?

betatim:
No worries, I also don't like the name but couldn't come up with something good. `GradientBoostingQuantileRegressor` is my favourite so far.

betatim:
Reopened as we still need to expose `base_estimator`

PR #28, opened 2016-03-29
Rename GBTQuantiles to GradientBoostingQuantileRegressor

betatim:
This is a better name, addresses one comment from #26


MechCoder:
Looks a bit like Java, but I don't have a better opinion.

betatim:
If someone has a :bulb: we can always change it later.

PR #34, opened 2016-03-29
GBRT based minimizer

### LINKED ISSUE #24: GBRT based minimization
GBRT now returns the quantiles. We can get a naive approximation to the std by subtracting the 68th quantile from the 50th quantile and feeding it to the acquistion functions.

betatim:
Fixes #24.

This is a minimal implementation of a tree based minimiser. Right now it only supports EI, and can't deal with categorical variables.

The `# XXX` comments are where I am not sure if it is worth doing something now  (for this PR) or leaving it for later.

It looks like `_expected_improvement` should be refactored so it can be shared with the GP model. After a brief discussion with @glouppe I think it is worth keeping it like this for the moment. In the longer term we might be able to do something smarter/more appropriate that is specialised to the tree based minimizer.

`_random_point` could also be shared, and in the future will need to get smarter for tree based models to deal with categorical variables.

Todo:

* [x] #35 need something smarter than `bfgs` to optimise tree ensembles
  - [ ] ~~can we be smarter than random sampling to minimise EI?~~ punting this to later
* [x] pass all the same tests that GP passes
 - [x] bench1
 - [x] bench2
 - [x] bench3
 - [x] branin
 - [x] hart6
* [x] `GBT.predict()` API #41 
* [x] #40 move GBT to `skopt.learning` (~~maybe~~definitely)


skopt/trees.py
```diff
@@ -0,0 +1,127 @@
...
+    ----------
+    func: callable
+        Function to minimize. Should take an array of parameters and
+        return the function value.
+
+    bounds: array-like, shape (n_parameters, 2)
+        ``bounds[i][0]`` should give the lower bound of each parameter and
+        ``bounds[i][1]`` should give the upper bound of each parameter.
+
+    base_estimator: a GradientBoostingQuantileRegressor
```
MechCoder: Is gradient boosting strictly a tree based approach?
We just fit the negative gradient to the data in every iteration right.
I would mention gradient boosting somewhere in the name of the function.

skopt/trees.py
```diff
@@ -0,0 +1,127 @@
...
+
+from .gbt import GradientBoostingQuantileRegressor
+from .utils import extract_bounds
+
+
+def _expected_improvement(x, surrogate, best_y):
+    """Evaluate expected improvement for `surrogate` model at `x`"""
+    x = x.reshape(1, -1)
+    low, mu, high = surrogate.predict(x)
+    std = (high - low) / 2.
```
MechCoder: We should mention that this an approximation to the standard deviation and need not be the exact standard deviation.

MechCoder:
Are gradient boosting methods popular in bayesian optimization? As I asked before, we could maybe write a paper if we find out that they are useful in some scenarios as compared to GP or RF based approaches.

cc: @glouppe

glouppe:
> Are gradient boosting methods popular in bayesian optimization? As I asked before, we could maybe write a paper if we find out that they are useful in some scenarios as compared to GP or RF based approaches

I have actually never seen them used, but I dont see why they would not work. In general, I believe quite a few things can be explored based on tree-based methods (of any kind). SMAC is the only example I know and it is a very successful proof-of-concept, despite the rather simple usage of RF. I would only expect that it can be improved in many interesting ways :)

skopt/trees.py
betatim: I will add it to L45. For the name, how about `gbt_minimize`?

betatim:
No idea. To be honest I picked gradient boosting as in scikit-learn it supports quantile regression out of the box...and I like gradient boosting. No deeper reason.

To me it seems like it shouldn't matter what you use (trees, GPs, ...) all you need is a regression model to act as the surrogate function and how (un)certain you are about its value. If your method can do this, you are cooking with gas :fire:.

With non-parametric models like trees you probably lose the "bayesian" part, at least it becomes harder/impossible to specify a prior.

The advantage is that you don't care so much about high dimensions, categorical variables, weird ranges, etc.

MechCoder:
I can comment in a less-blunt manner after the weekend, but it would be great to have a set of benchmarks and see what improvements can be done to the existing GP methods and if the new GBRT methods are better in different scenarios.

I hope to help to contribute to that in the coming week.

MechCoder:
Basically, https://github.com/MechCoder/scikit-optimize/issues/18

skopt/trees.py
betatim: this is addressed in 757b0b9

betatim:
Would like to get this to the point where it passes all of the tests that the GP passes.

betatim:
Related to #41: I kept finding myself having to do `.T` which was a smell for "something wrong here", now it stacks horizontally so `a,b,c = r.predict(X)` works.

glouppe:
Could you take care at #40 as well? :)

betatim:
Added to the todo.

skopt/trees.py
```diff
@@ -0,0 +1,139 @@
+"""Gradient boosted trees based minimization algorithms."""
```
glouppe: For consistency, can we rename that file `gbrt_opt.py`?

betatim:
We should also discuss which parts are worth factoring out into a shared place for GP, trees, future-methods. Seems silly to have EI and friends duplicated as long as they don't do clever things specialised for the method. Yet right now they would need probably as much code to adapt minor differences as each function is in size. So maybe a starting point would be a common set of tests that ensures that various acquisition functions compute the same thing.

skopt/trees.py
```diff
@@ -0,0 +1,139 @@
...
+def _random_point(lower, upper, random_state=None):
+    """Sample a random point"""
+    rng = check_random_state(random_state)
+
+    num_params = len(lower)
+    delta = upper - lower
+    return lower + rng.rand(num_params) * delta
+
+
+def gbt_minimize(func, bounds, base_estimator=None, maxiter=100,
```
glouppe: gbt -> gbrt (this is more common acronym in ML)

skopt/gbt.py
```diff
@@ -50,4 +50,4 @@ def fit(self, X, y):
 
     def predict(self, X):
         """Predictions for each quantile."""
-        return np.vstack([rgr.predict(X) for rgr in self.regressors_])
+        return np.asarray([rgr.predict(X) for rgr in self.regressors_])
```
glouppe: This is not consistent with the scikit-learn API. In my opinion, we should see that as a multi-output estimator, hence the output shape should be `len(X), n_outputs`.

skopt/trees.py
glouppe: also, as part of #38, I added an `n_start` argument. Surely, this should be helpful here as well.

skopt/gbt.py
betatim: Switched to this as I kept having to write: `a, b, c = e.predict().T`, is there a better way to unpack things?

MechCoder: Also it is not data-dependent and not argument dependent. It am thinking it might be better to return the approximate std directly..

betatim: What do you mean with data and argument dependent? It feels useful to be able to return as many/few quantiles as you'd want to instead of limiting yourself to mean and std. However there is no usecase for this yet.

betatim:
Notes:
```python
# do not want EI below zero, happens for points that are (much)
# worse than ``best_y``
ei = np.clip(ei, 0., np.max(ei))
```
adding this doesn't actually help/do anything. This confuses me.

All benchmarks pass (with same settings as for GP), except for `branin`.

Switched to using random sampling to find the minimum of the acquisition function for the moment. Can't think of a smarter way to do it right now.

Updated the todo list.

skopt/gbt.py
MechCoder: Sorry, I had meant that the number of outputs that you get from predict depends on the `quantiles` arguments and not `y` given to fit, which seemed a bit odd.

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    rng = check_random_state(random_state)
+
+    # Bounds
+    num_params = len(bounds)
+    lower_bounds, upper_bounds = extract_bounds(bounds)
+
+    # Default estimator
+    if base_estimator is None:
+        base_estimator = GradientBoostingQuantileRegressor()
```
MechCoder: Should we keep a different default? The default `max_depth` should be less right since it is a weak leaner.

MechCoder:
OK, so I ran a simple example to see if we are able to approximate the original function. This is what I get for the default values.

    from skopt.gbt import GradientBoostingQuantileRegressor
    from sklearn.gaussian_process import GaussianProcessRegressor

    import numpy as np

    X = np.array([0, 1, 2, 3, 4, 5, 6])
    y = np.sin(X)
    X = np.reshape(X, (-1, 1))
    X_new = np.linspace(0, 6, 200)
    X_new = np.reshape(X_new, (-1, 1))


    gbqr = GradientBoostingQuantileRegressor()
    # gpr = GaussianProcessRegressor()
    gbqr.fit(X, y)
    l, mean, h = gbqr.predict(X_new)
    std = (h - l) / 2.0
    plt.plot(X_new, mean)
    plt.plot(X_new.ravel(), np.sin(X_new.ravel()))
    plt.fill_between(X_new.ravel(), mean - std, mean + std, color='darkorange',
                     alpha=0.2)
    plt.show()

gbrt

MechCoder:
The graph does not change much by varying `max_depth` and `n_estimators`. It should be interesting to see what acquisition function is used by random forests and we can get inspired by that.

MechCoder:
I'll be able to have a closer look during the weekend hopefully ....

betatim:
Reading your email just now I realise that you were worried about this plot. I misunderstood you because I thought "oh ja, nice plot" and moved on.

Which part of it makes you think there is something up?

skopt/gbrt_opt.py
MechCoder: Ooh, I just realized that the default `max_depth` is 3. Scratch this comment

MechCoder:
I was just afraid that the `mean` and `std` are piecewise constant over quite large intervals, which means the `acquisition_function` is also constant over large intervals right? (Which also explains why lbfgs does not work as we would like it to and that you had to resort to random sampling)

MechCoder:
Something like this https://github.com/scikit-learn/scikit-learn/pull/6166#discussion_r54333346 would be better

betatim:
The piecewise constant `mu` and `std` stops us from using `bfgs` etc but I am not sure what else the tree could really predict. If you asked me (the `TimRegressor`) to make a prediction given only two or three points I would probably make a constant prediction as well as there isn't anything else to go on.

https://github.com/scikit-learn/scikit-learn/pull/6166#discussion_r54333346 still has small areas of constant prediction. I was wondering how "big" a constant prediction is "big" in terms of `bfgs`. Say we make a constant prediction over 0.1, 1 or 10% of the range of a variable, at some point the prediction will start appearing "continous" to `bfgs` even though in reality it is still piecewise constant. Wondering what that length scale is. Or if we will always be stuck because of this.

Do you know what arXiv:1211.0906 does about this problem?

betatim:
Added some examples to help debugging which are mostly copy&paste from @MechCoder's examples for GP. Plots here:

bench3


branin

Fixed a random state that wasn't fixed, now `hart6` works but `branin` fails. Increasing the number of iterations to ~200 is enough to have it converge though.

skopt/gbt.py
```diff
@@ -50,4 +50,4 @@ def fit(self, X, y):
 
     def predict(self, X):
         """Predictions for each quantile."""
-        return np.vstack([rgr.predict(X) for rgr in self.regressors_])
+        return np.asarray([rgr.predict(X) for rgr in self.regressors_]).T
```
glouppe: Can you rename this file to `gbrt.py`?

skopt/tests/test_gbt.py
```diff
@@ -22,6 +24,7 @@ def sample_noise(X, std=0.2, noise=constant_noise,
     rng = check_random_state(random_state)
     return np.array([rng.normal(0, std*noise(x)) for x in X])
 
+
 def test_gbt_gaussian():
```
glouppe: Same here, could you rename all occurences of `gbt` to `gbrt`?

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+
+def _expected_improvement(X, surrogate, best_y, xi=0.01):
+    """Evaluate expected improvement for `surrogate` model at `x`
+
+    Parameters
+    ----------
+    X : array-like
+        Values at which to evaluate the acquisition function.
+
+    best_y: float, optional
```
glouppe: We named this `y_opt` in `gp_minimize`

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+    for i in range(1, maxiter + 1):
+        rgr = clone(base_estimator)
+        # only the first i points are meaningful
+        rgr.fit(Xi[:i, :], yi[:i])
+        models.append(rgr)
+
+        # `rgr` predicts constants for each leaf which means that the EI
+        # has zero gradient over large distances. As a result we can not
+        # use gradient based optimisers like BFGS, use random sampling
+        # for the moment.
```
glouppe: Could you add an `n_points` argument for controlling the number of points in random search, as in `gp_minimize`?

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+    if base_estimator is None:
+        base_estimator = GradientBoostingQuantileRegressor(random_state=rng)
+
+    # Record the points and function values evaluated as part of
+    # the minimization
+    Xi = np.zeros((maxiter + 1, num_params))
+    yi = np.zeros(maxiter + 1)
+
+    # Initialize with a random point
+    Xi[0] = _random_point(lower_bounds, upper_bounds, random_state=rng)
```
glouppe: Would be worth also adding an `n_start` argument to start from several random points (this is often much better to avoid local minima).

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+    best_x = Xi[0].ravel()
+    yi[0] = best_y = func(Xi[0])
+
+    models = []
+
+    # XXX should there be an early stopping criterion?
```
glouppe: Nope, not for now

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,154 @@
...
+    improvement = best_y - xi - mu[mask]
+    exploit = improvement * stats.norm.cdf(improvement / std[mask])
+    explore = std[mask] * stats.norm.pdf(improvement / std[mask])
+    ei[mask] = exploit + explore
+
+    ei = -ei # we are being used in a minimizer
+    return ei
+
+
+def _random_point(lower, upper, n_points=1, random_state=None):
```
glouppe: Is this function really necessary? We inlined it in `gp_minimize` by doing `X = lb + (ub - lb) * rng.rand(n_points, n_params)` instead.

betatim: Needed it at several points and was planning ahead for when we have categorical variables as well so I made a function out of it.

skopt/gbrt_opt.py
betatim: ✅

betatim:
Renamed `gbt.py -> gbrt.py` but then remembered we wanted to move the contents to `skopt.learning`. Should it be `skopt.learning.gbrt.GradientBoost...`? I would prefer `skopt.learning.GradientBoost...`. When that is done, all things from the to do list should be ticked off.

glouppe:
> Should it be skopt.learning.gbrt.GradientBoost...? I would prefer skopt.learning.GradientBoost....

Create `learning/gbrt.py` and then exposes whatever is needed in `learning/__init__.py`, so that `skopt.learning.GradientBoost...` works.

betatim:
Travis is happy, Tim is happy, @glouppe happy too?

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,173 @@
...
+    xi: float, default 0.01
+        By how much must a new point improve over the previous best
+        value.
+    """
+    X = np.asarray(X)
+    if X.ndim == 1:
+        X = np.expand_dims(X, axis=0)
+
+    # low and high are assumed to be the 16% and 84% quantiles
+    low, mu, high = surrogate.predict(X).T
```
MechCoder: What do you think about adding a `return_std` to `Quantile.predict` option that returns just the `mean` and `std`? In that way we can easily refactor this into the existing implementation and add "LCB" support.

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,173 @@
...
+        The regressor to use as surrogate model
+
+    maxiter: int, default 100
+        Number of iterations used to find the minimum. This corresponds
+        to the total number of evaluations of `func`. If `n_start` > 0
+        only `maxiter - n_start` iterations are used.
+
+    n_start: int, default 10
+        Number of random points to draw before fitting `base_estimator`
+        for the first time. If `n_start < maxiter` this degrades to
```
MechCoder: Should this be `n_start = maxiter`?

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,173 @@
...
+
+    if maxiter == 0:
+        raise ValueError("Need to perform at least one iteration.")
+
+    n_start = min(n_start, maxiter)
+
+    Xi[:n_start] = _random_point(
+        lower_bounds, upper_bounds, n_points=n_start, random_state=rng)
+    best_x = Xi[:n_start].ravel()
+    yi[:n_start] = [func(xi) for xi in (Xi[:n_start])]
```
MechCoder: Is this a Python 3 thing? First time I'm seeing this syntax.

skopt/gbrt_opt.py
betatim: Oups. Actually should be `n_start > maxiter`

skopt/gbrt_opt.py
betatim: Just a list comprehension no? Though there are spurious parentheses around `Xi[:n_start]`. Will remove those

skopt/gbrt_opt.py
MechCoder: In that case, we should raise an error, no?

skopt/gbrt_opt.py
```diff
@@ -0,0 +1,173 @@
...
+    if maxiter == 0:
+        raise ValueError("Need to perform at least one iteration.")
+
+    n_start = min(n_start, maxiter)
+
+    Xi[:n_start] = _random_point(
+        lower_bounds, upper_bounds, n_points=n_start, random_state=rng)
+    best_x = Xi[:n_start].ravel()
+    yi[:n_start] = [func(xi) for xi in Xi[:n_start]]
+    best_y = np.min(yi[:n_start])
```
MechCoder: `best_x` should `Xi[np.argmin(yi[:n_start])]` right?

skopt/gbrt_opt.py
betatim: Yes, because reimplementing the acquisition functions is silly. No, because we could do something smarter when you can estimate the quantile directly instead of having to make an assumption about things being gaussian.

MechCoder: Oh, we can keep the existing implementation by setting `return_std` to False by default, if that worries you

MechCoder: Just to clarify, I'm suggesting to keep both implementation, one that returns the quantiles by default and the other that returns the `mean` and `std` when return_std is set to True.

If you want I can do that refactoring separately after this has been merged.

betatim: Aha! I hadn't even thought of such a behaviour. It is a good idea. We'd have to check the user asked for the right quantiles to be estimated when they use `return_std=True`.

I was preoccupied with the thought that for LCB you'd only need to estimate one quantile instead of all three.

New PR?

skopt/gbt.py
betatim: ✅

MechCoder:
We can merge after Travis passes. But here are some thoughts. For things like `lbfgs`, we would need smooth functions. Or else in this case, `lbfgs` would degrade to random sampling since the gradient is zero.

Here is what they do in the SMAC paper and claim it gives smooth decision surfaces, but I don't grasp the intuition quickly. Let us say a split point is `x(j, k)`, any point in between the next point `x(j, k)` and `x(j, k+1)` also is a split point, so they sample this split point uniformly between `x(j, k)` and `x(j, k+1)`

This has been described in page 17 of the paper (http://arxiv.org/pdf/1211.0906v2.pdf) just in case my interpretation is wrong.

MechCoder:
Merging for now! Thanks :-)

betatim:
Thanks for all the comments and help!

glouppe:
:beers:

Issue #37, opened 2016-04-04
Add probability of improvement acquisition function


glouppe:

glouppe:
Fixed in #38

MechCoder:
Is this useful for practical purposes or is it more of a completion thing?

Issue #36, opened 2016-04-04
Extend test suite


glouppe:
Before implementing any more things, we should really extend the test suite with more thorough tests. At the moment, I cant even minimize a 1D parabola with the default parameters of `gp_minimize`... 

(and I dont even understand why it fails... so many things to adjust :/)

We might want to look at other packages for good defaults.

glouppe:
Fixed in #38

Issue #35, opened 2016-04-01
Methods to optimize the acquisition function


MechCoder:
Right now we support `lbfgs` and `random sampling`. What are some other methods to optimize the acquistion function?

MechCoder:
@glouppe @betatim What are your thoughts on this?

betatim:
Is there some prior art/literature on using random sampling? It seem like a really strange strategy to use.

One more thing to put on the list would be something like `bfgs` but with multi start (start from `n` random points and only keep the best minimum of all `n` attempts)

glouppe:
+1 for several random restarts added to l-bfgs. 

Other than that, I have seen the "DIRECT" algorithm been used and cited in several BO papers/packages. 
https://hosting.cs.vt.edu/hpcs2008/HPCS08DIRECTtut.pdf

glouppe:
... and here is the commonly used Python implementation https://pypi.python.org/pypi/DIRECT/

betatim:
From today's arxiv: http://arxiv.org/abs/1604.00169 from the abstract it sounds like this does away with acquisition functions all together. Could be OT here but also maybe not.

glouppe:
Partially fixed as part of #38

betatim:
Trees predict a constant for each leaf. This means `bfgs` (and other gradient based) optimisers give up. Pointers to smarter ways for finding the minimum of an ensemble of trees welcome. This is needed for #34.

glouppe:
I am closing this, as this issue is too vague without a clear objective. 

See #74 for ongoing work on local search (one of the many possible solvers).

PR #38, opened 2016-04-04
[MRG] Add more tests

glouppe:
* [x] Add more tests #36 
* [x] Added random restarts as in #35 
* [x] Fixed acquisition functions
* [x] Rename UCB to LCB, since for minimization we are dealing with the lower bound. 
* [x] Added probability of improvement #37

### DIFF
--- a/skopt/benchmarks.py
+++ b/skopt/benchmarks.py
@@ -1,6 +1,40 @@
-"""A collection of benchmark problems"""
+"""A collection of benchmark problems."""
 
 import numpy as np
+
+
+def bench1(x):
+    """A benchmark function for test purposes.
+
+        f(x) = x ** 2
+
+    It has a single minima with f(x*) = 0 at x* = 0.
+    """
+    return np.asscalar(x ** 2)
+
+
+def bench2(x):
+    """A benchmark function for test purposes.
+
+        f(x) = x ** 2 if x < 0
+               (x-5) ** 2 - 5 othewise.
+
+    It has a global minima with f(x*) = -5 at x* = 5.
+    """
+    if x < 0:
+        return np.asscalar(x ** 2)
+    else:
+        return np.asscalar((x - 5) ** 2 - 5)
+
+
+def bench3(x):
+    """A benchmark function for test purposes.
+
+        f(x) = sin(5*x) * (1 - tanh(x ** 2))
+
+    It has a global minima with f(x*) ~= -0.9 at x* ~= -0.3.
+    """
+    return np.asscalar(np.sin(5 * x) * (1 - np.tanh(x ** 2)))
 
 
 def branin(x, a=1, b=5.1 / (4 * np.pi**2), c=5. / np.pi,
@@ -12,8 +46,9 @@ def branin(x, a=1, b=5.1 / (4 * np.pi**2), c=5. / np.pi,
 
     More details: <http://www.sfu.ca/~ssurjano/branin.html>
     """
-    return (a * (x[1] - b * x[0]**2 + c * x[0] - r)**2 +
-            s * (1 - t) * np.cos(x[0]) + s)
+    return np.asscalar(a * (x[1] - b * x[0] ** 2 + c * x[0] - r) ** 2 +
+                       s * (1 - t) * np.cos(x[0]) + s)
+
 
 def hart6(x,
           alpha=np.asarray([1.0, 1.2, 3.0, 3.2]),
--- a/skopt/benchmarks.py
+++ b/skopt/benchmarks.py
@@ -16,8 +16,8 @@ def bench1(x):
 def bench2(x):
     """A benchmark function for test purposes.
 
-        f(x) = x ** 2 if x < 0
-               (x-5) ** 2 - 5 othewise.
+        f(x) = x ** 2           if x < 0
+               (x-5) ** 2 - 5   otherwise.
 
     It has a global minima with f(x*) = -5 at x* = 5.
     """
--- a/skopt/dummy_opt.py
+++ b/skopt/dummy_opt.py
@@ -1,11 +1,12 @@
 import numpy as np
-from scipy.optimize import OptimizeResult
 
+from scipy.optimize import OptimizeResult
 from sklearn.utils import check_random_state
 
-from skopt.utils import extract_bounds
+from .utils import extract_bounds
 
-def dummy_minimize(func, bounds=None, maxiter=1000, random_state=None):
+
+def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
     """
     Sample each parameter uniformly within the given bounds.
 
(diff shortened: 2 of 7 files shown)


glouppe:
Hmm, it seems much easier to make tests pass without scaling the features. At least, it can minimize a trivial parabola :x 

What are your thoughts @MechCoder ?

glouppe:
Hmm, I now have the 3 new benchmarks passing. The remaining failing tests are on `branin` and `hart6`. Shall we hardcode a better default for them in the unit test?

glouppe:
Not included here, but I just MinMaxScaling the input values fed to the GP. This seems to lead to much less numerical issues. However, optimization still gets stuck in local optima and/or is slow to converge.

skopt/gp_opt.py
```diff
@@ -171,22 +170,20 @@ def gp_minimize(func, bounds, base_estimator=None, acq="UCB", xi=0.01,
...
     # Bounds
-    num_params = len(bounds)
-    lower_bounds, upper_bounds = zip(*bounds)
-    upper_bounds = np.asarray(upper_bounds)
-    lower_bounds = np.asarray(lower_bounds)
-    lbfgs_bounds = np.tile((0, 1), (num_params, 1))
+    n_params = len(bounds)
+    lb, ub = zip(*bounds)
+    ub = np.asarray(ub)
+    lb = np.asarray(lb)
```
betatim: :arrow_right: `utils.extract_bounds(bounds)`

glouppe:
Some progress, only tests for `hart6` are now failing!

glouppe:
WIP -> MRG

Tests should now pass :) Please review!

skopt/gp_opt.py
```diff
@@ -55,37 +55,30 @@ def acquisition(X, model, x_opt=None, method="UCB", xi=0.01, kappa=1.96,
...
 
     # Evaluate acquisition function
     if method == "UCB":
         values = mu - kappa * std
 
     elif method == "EI":
-        improvement = x_opt - mu
-        exploit = improvement * stats.norm.cdf(improvement / std)
+        improvement = mu - x_opt - xi
+        exploit = std * improvement * stats.norm.cdf(improvement / std)
```
glouppe: @betatim this should be proofchecked

betatim: From Shahriari et al "Taking the Human out of the loop" eqn 44 I think it should be just `improvement * stats.norm.cdf(improvement / std)`.

I agree on L68 where the order of `mu` and `x_opt` has been swapped.

Do you have a reference with some details on the ideas behind subtracting `xi`?

glouppe: Indeed, you are right. 

For `xi`, it comes from Brochu 2010, eqn 4.

glouppe: But I may be wrong, because Brochu considers the maximization problem, and consider the maximization of the acquisition. Hence this really needs to be double-checked.

glouppe:
I did some more changes. This time we are good! See the first comment for a summary of the changes.

skopt/benchmarks.py
```diff
@@ -1,8 +1,42 @@
...
+
+    It has a single minima with f(x*) = 0 at x* = 0.
+    """
+    return np.asscalar(x ** 2)
+
+
+def bench2(x):
+    """A benchmark function for test purposes.
+
+        f(x) = x ** 2 if x < 0
```
betatim: insert some spaces before the if maybe? Found it a bit tricky to find the `if` after spotting the `othewise`

L20 othewise -> otherwise

glouppe: Thanks! Fixed

skopt/dummy_opt.py
```diff
@@ -43,22 +44,22 @@ def dummy_minimize(func, bounds=None, maxiter=1000, random_state=None):
     rng = check_random_state(random_state)
 
     # Bounds
-    num_params = len(bounds)
-    lower_bounds, upper_bounds = extract_bounds(bounds)
-    diff = upper_bounds - lower_bounds
+    n_params = len(bounds)
+    lb, ub = extract_bounds(bounds)
 
     # Sample each parameter uniformly between 0 and 1 and then rescale.
```
betatim: That isn't true anymore, we sample some uniform numbers and then multiply. To me this comment makes it sound like we do everything in the domain [0,1] and only at the very every end scale back to the original domain.

skopt/gp_opt.py
```diff
@@ -114,28 +117,29 @@ def gp_minimize(func, bounds, base_estimator=None, acq="UCB", xi=0.01,
...
-        Useless if acq is set to "EI"
+        Used when the acquisition is "LCB".
 
     search: string, "sampling" or "lbfgs"
         Searching for the next possible candidate to update the Gaussian prior
         with.
 
-        If search is set to "sampling", ``num_points`` are sampled randomly
+        If search is set to "sampling", ``n_points`` are sampled randomly
         and the Gaussian Process prior is updated with that point that gives
```
betatim: with that point -> with the point

skopt/gp_opt.py
```diff
@@ -171,22 +181,22 @@ def gp_minimize(func, bounds, base_estimator=None, acq="UCB", xi=0.01,
...
-            kernel=Matern(length_scale=np.ones(num_params), nu=2.5),
-            normalize_y=True, random_state=random_state)
+            kernel=(ConstantKernel(constant_value_bounds="fixed") *
+                    Matern(length_scale=np.ones(n_params),
+                           length_scale_bounds="fixed", nu=2.5) +
+                    ConstantKernel(constant_value_bounds="fixed")),
+            normalize_y=True, alpha=10e-6, random_state=random_state,
+            n_restarts_optimizer=3)
 
     # First point
```
betatim: point -> pointS

betatim:
Wait for travis then merge?

glouppe:
Yup, thanks for the review!

betatim:
Do you know why travis was unhappy? Not sure I understand why it couldn't find the PR

MechCoder:
Wow, I did not look for a few days, and the repo seems to be blowing up with activity.

Sorry for the delay guys, the last week had been long, hard and dark (and the next 1.5 years are going to be as hard), but that won't stop me ;)

betatim:
Don't worry, that is why there is more than one of us: always someone keeping the lights on :)

skopt/gp_opt.py
```diff
@@ -171,60 +182,60 @@ def gp_minimize(func, bounds, base_estimator=None, acq="UCB", xi=0.01,
...
 
     # Default GP
     if base_estimator is None:
         base_estimator = GaussianProcessRegressor(
-            kernel=Matern(length_scale=np.ones(num_params), nu=2.5),
-            normalize_y=True, random_state=random_state)
+            kernel=(ConstantKernel(constant_value_bounds="fixed") *
+                    Matern(length_scale=np.ones(n_params),
+                           length_scale_bounds="fixed", nu=2.5) +
+                    ConstantKernel(constant_value_bounds="fixed")),
```
MechCoder: Could you please explain what adding a fixed `ConstantKernel` achieves? It is not clear to me.

glouppe: Not sure exactly, but this is something I saw in https://github.com/SheffieldML/GPyOpt/blob/540fda8344a936ed315b9a929afddacbc8adf6f9/GPyOpt/methods/bayesian_optimization.py#L92

glouppe: But for sure, it makes no sense now if it is fixed... We should either remove it or set tighther bounds as explained in #45

PR #42, opened 2016-04-05
Update install.sh with latest miniconda

glouppe:


betatim:
Do we even need conda? For `everware` we use https://github.com/minrk/travis-wheels which is pretty quick. Something to think about.

PR #44, opened 2016-04-05
[MRG] Some more tests

glouppe:


glouppe:
This adds some more tests. I'll merge if Travis is happy.

skopt/tests/test_dummy_opt.py
```diff
@@ -0,0 +1,35 @@
...
+    r = dummy_minimize(func, bounds, maxiter=maxiter, random_state=1)
+    assert_less(r.fun, y_opt + margin)
+
+
+def test_dummy_minimize():
+    yield (check_minimize, bench1, 0., [[-2, 2]], 0.05, 10000)
+    yield (check_minimize, bench2, -5, [[-6, 6]], 0.05, 10000)
+    yield (check_minimize, bench3, -0.9, [[-2, 2]], 0.05, 10000)
+    yield (check_minimize, branin, 0.39, [[-5, 10], [0, 15]], 0.1, 10000)
+    yield (check_minimize, hart6, -3.32, np.tile((0, 1), (6, 1)), 0.5, 10000)
```
glouppe: Why bother making it complicated when this pass? :( @betatim @MechCoder

betatim:
Less iterations? I think that is the big selling point.

Does it make sense to add a test (or should it be a benchmark?) that puts the "clever" algorithms in a race against `dummy_minimize` to see which one converges faster? If implemented as a test you would fail if you need (statistically significant) more iterations than `dummy_minimize`.

skopt/tests/test_dummy_opt.py
MechCoder: Great question.

Well, these are "simple" functions. We should investigate complicated functions where we can gain some information from the data, for instance `LogisticRegression`, where we can choose hyperparameters close to those that have high regularization.

`LogisticRegression` might be a bad example because again we have to specify the scale in which we need to query the parameters.

MechCoder:
@betatim Yes, I think that should be the first step from now onwards before implementing new stuff.

1. We should know if we are performing better than dummy search.
2.  We should know if any of the other new methods add any value.

betatim:
Should we rename dummy -> random? In particular if it continues to be useful/competitive we might want to give it a less patronising name.

Issue #45, opened 2016-04-05
Tune the hyperparameters of the Gaussian kernel


MechCoder:
According to the talk with @glouppe on chat, in https://github.com/MechCoder/scikit-optimize/pull/38 the length scale of the kernel has been fixed to be 1.0 . We should either

1. Tune the hyperparameter by maximizing the log likelihood function and get a point estimate of this value as done before. (which does not seem to work) (or)
2. Maximize the value of the integrated acquisition function across all hyperparameters.

glouppe:
Yes, for sure we are not understanding something properly here. This needs to be investigated! 

From code in https://github.com/jmetzen/bayesian_optimization/blob/d115d6237475c832ace40bb732a12de8476ead8e/examples/plot_bayesian_optimization.py it seems @jmetzen usually sets tighter bounds.

glouppe:
It takes longer, but the following GP works:

```
        base_estimator = GaussianProcessRegressor(
            kernel=(ConstantKernel(1.0, (0.01, 1000.0)) *
                    Matern(length_scale=np.ones(n_params),
                           length_scale_bounds=[(0.01, 100)] * n_params,
                           nu=2.5) +
                    ConstantKernel(1.0, (0.01, 1000.0))),
            normalize_y=True, alpha=10e-6, random_state=random_state)
```

What do you think?

glouppe:
Let's continue the discussion in #48.

PR #46, opened 2016-04-05
[MRG] Make kernel tunable

glouppe:
* [x] Address #45 
* [x] Silence warnings


glouppe:
cc @MechCoder

MechCoder:
Thanks! But I still think the ConstantKernel can be removed, I'll try that in a different PR

PR #47, opened 2016-04-05
Remove constant kernel

MechCoder:


MechCoder:
@glouppe

glouppe:
Since tests pass, I am ok, but I am curious to understand something. If we use as kernel something like `C(np.ones(n_features)) * K + C(np.ones(n_features))`, wouldnt that automatically adjust for the offset and scaling, as you propose to do by hand in #48? 

CC: @jmetzen could you pitch in? What would you recommend?

jmetzen:
Right, a constant kernel multiplied with another kernel modulates the amplitude of the kernel. Removing it is in general not a good idea. A noise-free GP will still interpolate the observations but is not able to adjust the posterior's variance at non-observed inputs. That means, it will either underestimate or overestimate uncertainty. This is not a desirable property when using the GP as a surrogate model in Bayesian optimization.

As for adding a constant kernel: yes, it can effectively adjust an offset of the function. But it is not a clean solution, as such an offset should actually be in the GP's mean and not in its covariance function (kernel). There are also some theoretical issues with a constant offset kernel if I remember it correctly. It works well in practice however, in my experience. Alternatively, there is a parameter normalize_y for GPR, which can be used to deal with non-zero mean target functions.

BTW: The example plot_gpr_prior_posterior.py in sklearn allows visualizing the GP priors and posteriors for different kernel. It's quite useful for understanding what assumptions the different kernels can encode, in my opinion.

glouppe:
Thanks for the great explanation @jmetzen !

@MechCoder Since we use `normalize_y=True`, I would be in favour of using something like `C(np.ones(n_params)) * Matern(np.ones(n_params))` with bounds not allowed to be too small nor too large.

MechCoder:
Ah yes, of course I did not notice that the scale of each component of each element is not tuned by default and that a `ConstantKernel` has to be supplied.

PR #49, opened 2016-04-05
Input output checks for acquistion and func to be optimized

MechCoder:
Fixes https://github.com/MechCoder/scikit-optimize/issues/39


glouppe:
Thanks!

PR #50, opened 2016-04-06
Add example to show random search vs gp

MechCoder:


MechCoder:
@betatim @glouppe 

random_search

DISCLAIMER: I varied the random state till `gp_search` performs better than `random_search` :P

betatim:
:cherries: picking :)

To avoid the option of tweaking `random_state` I would suggest running the whole thing several times and plotting the mean score after each iteration and the error on the mean.

MechCoder:
I thought about that as well.

But finding the mean score at every iteration, kind of defeats the purpose right? Typically you have the budget to only run n iterations and not `n * x`, so it does depends on the random state. If we average this randomness out, we are more likely to get a better score in every iteration.

MechCoder:
Maybe, I'll give it a try.

betatim:
Doing `n` repeats isn't something you'd do in a real case. For the example it is a way to allow people to judge if there is a statistically significant advantage for GP and how big that advantage is.

If we repeat the example 100 times with 100 different seeds and ~50 times GP wins and ~50 random wins the conclusion would be that there isn't much difference. If however in 90 cases GP wins, you'd conclude that GP is really much smarter.

MechCoder:
Indeed, I agree with you!

betatim:
Wondering if average score after each iteration is the best way to visualise this or if it would be better to show the distribution of the final (after 100 iterations) score for each method.

MechCoder:
new_dummy_vs_gp

I averaged the best scores across 5 random states. Hope the new graphs are convincing. Please merge if happy.

MechCoder:
I have changed the example.

glouppe:
Thanks! However I think we should not mix the messages and illustrate only one concept at a time. What do you think of only showing how to use `gp_minimize` for hyperparameter search, without comparison with RandomizedSearchCV?

MechCoder:
Removed the comparison with dummy_search and updated just to show how to use in combination with sklearn estimator.

examples/plot_random_vs_gp.py
```diff
@@ -0,0 +1,55 @@
...
+def compute_mean_validation_score(forest_params):
+    # Hack to allow integer parameters, since the parameters
+    # sampled internally are uniform in a given range.
+    forest_params = [int(param) for param in forest_params]
+    max_depth, max_features, mss, msl = forest_params
+
+    params = {
+        'max_depth': [max_depth], 'max_features': [max_features],
+        'min_samples_split': [mss], 'min_samples_leaf': [msl]}
+    gscv = GridSearchCV(rfc, params, n_jobs=-1)
```
glouppe: Not sure why GridSearchCV is needed. `cross_val_scores` should be enough.

examples/plot_random_vs_gp.py
```diff
@@ -0,0 +1,55 @@
...
+
+from skopt.gp_opt import gp_minimize
+
+digits = load_digits()
+X, y = digits.data, digits.target
+rfc = RandomForestClassifier(random_state=10)
+
+def compute_mean_validation_score(forest_params):
+    # Hack to allow integer parameters, since the parameters
+    # sampled internally are uniform in a given range.
```
glouppe: This would worth of an issue, to discuss about the best API to define parameter types.

MechCoder: Closely related: https://github.com/MechCoder/scikit-optimize/issues/56

MechCoder:
@glouppe Addressed!

glouppe:
Merging!

PR #51, opened 2016-04-08
Add branin plot

MechCoder:


MechCoder:
branin

glouppe:
This looks good and convincing! However I am a bit lost about this PR and #50. Arent they both adding an example comparing dummy vs gp?

MechCoder:
Yes, but I added this to reproduce the results given here (http://arxiv.org/pdf/1402.0929v3.pdf)

Spearmint seems to achieve a minimum of 0.398 after 200 evaluations using the integrated acquisition function, and we achieve 0.398 by maximizing the likelihood.

MechCoder:
@glouppe @betatim Would you be able to merge this?

betatim:
Is the conclusion then to stop #50 after merging this one?

----

Can you upload the output of the branin plot function?

examples/plot_branin.py
```diff
@@ -1,54 +1,59 @@
...
-plt.xlim([-5, 10])
-plt.ylabel('X2')
-plt.ylim([0, 15])
-plt.title("Branin function values.")
-
+best_dummy_scores = np.zeros((5, 200))
+best_gp_scores = np.zeros((5, 200))
+n_iterations = range(1, 201)
+
+for random_state in range(5):
```
betatim: Is it feasible to do ten rounds or would you have to wait too long? (not super important)

examples/plot_branin.py
```diff
@@ -1,54 +1,59 @@
...
+    print(time() - t)
+    print("Best score obtained, %0.4f" % gp_model.fun)
+
+    for j in range(1, 201):
+        best_dummy_scores[random_state, j-1] = np.min(dummy_model.func_vals[:j])
+        best_gp_scores[random_state, j-1] = np.min(gp_model.func_vals[:j])
+
+mean_dummy_scores = np.mean(best_dummy_scores, axis=0)
+mean_gp_scores = np.mean(best_gp_scores, axis=0)
+std_dummy_scores = np.std(best_dummy_scores, axis=0)
```
betatim: Should we be computing the standard error on the mean? Can someone help with figuring this out?

We show the mean, and are interested in how much the mean can vary, which makes me think `sem`.

examples/plot_branin_acquisition.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+from skopt import acquisition
+from skopt import gp_minimize
+from skopt.benchmarks import branin
+
```
betatim: blank line

examples/plot_branin_acquisition.py
```diff
@@ -0,0 +1,54 @@
...
+fig = plt.gcf()
+plt.set_cmap("viridis")
+bounds = [[-5, 10], [0, 15]]
+subplot_no = 311
+
+x1_values = np.linspace(-5, 10, 100)
+x2_values = np.linspace(0, 15, 100)
+x_ax, y_ax = np.meshgrid(x1_values, x2_values)
+vals = np.asarray(list(zip(x_ax.ravel(), y_ax.ravel())))
```
betatim: Is `np.c_[x_ax.ravel(), y_ax.ravel()]` truly better or just different syntax?

examples/plot_branin.py
MechCoder: Yes, I think

examples/plot_branin_acquisition.py
MechCoder: This should there right? Separates the rest of the code from the imports.

MechCoder:
I have addressed your comments. This is the new graph.

branin

Also, the other pull request deals with using a sklearn estimator combined with `gp_minimize`. Should I remove the comparison with `dummy_minimize` in this case.

examples/plot_branin.py
MechCoder: Currently it takes ~25 seconds for every round :/

skopt/__init__.py
```diff
@@ -1,2 +1,3 @@
+from .gp_opt import acquisition
```
glouppe: Not sure what is best yet regarding the API, but I would rather not expose `gp_opt.acquisition` at the level of the package. `gp_opt.acquisition` is currently specific to GPs. 

For the example, I would rather do `from skopt.gp_opt import acquisition`

glouppe:
To make things clearer, I would do something like this:
- keep `plot_branin.py` of this PR, but rename to `plot_comparison.py`.
- keep `plot_branin_acquisition.py` of this PR, but rename to `plot_gp_minimize_2d.py`. In this example, we should show the posterior mean of for the objective function, the posterior std for the objective function and the acquisition function (only one of them). Maybe we could have these 3 plots repeated for epoch=1, epoch=N/2 and epoch=N?
- for consistency, I would rename `plot_1d_demo.py` to `plot_gp_minimize_1d.py` and show the same corresponding plots (except that they are now in 1d)
- in #50, rename `plot_random_vs_gp.py` to `plot_hyperparameter_optimization.py` and focus on using gp_minimize as a replacement of GridSearch.

skopt/__init__.py
MechCoder: Sure

MechCoder:
All right, I've addressed the first of your comments. I'll address the remaining 3 in follow up PR's

betatim:
That is a nice looking plot and thanks for addressing the comments.

PR #52, opened 2016-04-14
Lower confidence bound as a function of no. of iterations

MechCoder:
See: https://github.com/MechCoder/scikit-optimize/pull/51#issuecomment-209834106

### DIFF
--- a/examples/plot_gp_minimize_2d.py
+++ b/examples/plot_gp_minimize_2d.py
@@ -1,3 +1,25 @@
+"""
+=============================================================
+Lower confidence bound as a function of number of iterations.
+=============================================================
+
+``gp_minimize`` approximates a costly function with a GaussianProcess prior
+and instead of minimizing this costly function, it minimizes an acquisition
+function which is relatively cheaper to minimize. Before every iteration of
+``gp_minimize``, a fixed number of points are used to approximate the
+GaussianProcess prior. The acquisition function combines information
+about
+
+1. New points that are close to the previous points used to approximate the GP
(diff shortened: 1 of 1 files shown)


MechCoder:
figure_1-4

glouppe:
Hmm, maybe there is too much information. What do you think of letting the optimization run for 10 or 20 iterations and then show in 3 plots:
1) the function and its minimum
2) the mean gp value, the sampled points, the best sampled point, and the true minimum
3) the acquisition function, the sampled points, the best sampeld point, and the true minimum

1 and 2 should share the same colormap.

MechCoder:
I really think we should plot the uncertainty as well. It gives a good idea of the exploration-exploitation trade-off.

As far as the variation with n_iter goes, do you think it would be better to do it in https://github.com/MechCoder/scikit-optimize/pull/54 ?

glouppe:
> I really think we should plot the uncertainty as well. It gives a good idea of the exploration-exploitation trade-off.

okay, let's try with 4 plots then?

> As far as the variation with n_iter goes, do you think it would be better to do it in #54 ?

Yes, it might be easier to illustrate in the 1d case

MechCoder:
Sounds great!

MechCoder:
@glouppe addressed.

glouppe:
how does this look?

MechCoder:
2d

glouppe:
For all 4 plots, could you show  the sampled points (in white), the best sampled point (red?), and the true minimum (blue?)?

PR #54, opened 2016-04-14
Plot 1-D acquisition values

MechCoder:


MechCoder:
@glouppe 

gp_1d

glouppe:
Thanks for working on this! I think we should aim for something looking more like https://github.com/glouppe/talk-bayesian-optimisation/blob/master/code/fig4-3.pdf 

- I would suggest using `benchmarks.bench3` instead
- Then we should show two plots on top of each other:
1) the true function, mean value of the gp, std of the gp around the mean, sampled points
2) the acquisition function and its min

MechCoder:
@glouppe I have addressed your comments. Here are the new graphs.

gp_2
gp_1

MechCoder:
What does the red region in your graph depict around the red line?

glouppe:
> What does the red region in your graph depict around the red line?

In my example, I added noise around f(x). The red region is the 1-sigma band.

glouppe:
Looks really cool! Merging

PR #58, opened 2016-04-18
Rewrite 1d plot acquisition plot

MechCoder:


MechCoder:
@glouppe @amueller had suggested to use this plot which looks much better.

lcb

amueller:
there should be no connection between the red dots in the legend ;)

MechCoder:
lcb

MechCoder:
Not sure why matplotlib does that by default

glouppe:
Indeed, much better! To avoid clutter, could you show the legend only for epoch=1? Also maybe up to epoch=5 is enough.

MechCoder:
Done!
gp_1

glouppe:
Could you show the legend only for epoch=1? Then +1 for merge

MechCoder:
done

glouppe:
Thanks, merging :)

betatim:
Nice plotting!

PR #61, opened 2016-04-20
Lower confidence bound as a function of no.of iterations

MechCoder:


MechCoder:
@glouppe I addressed your comments. This is how it looks

2d_gp

glouppe:
Merging, thanks! We will polish that later :)

Issue #56, opened 2016-04-15
Implement weighted hamming distance kernel for categorical inputs


MechCoder:
Section 4.1.2 in http://arxiv.org/pdf/1211.0906v2.pdf

glouppe:
This should be rebased on top of master. Then I would assume that that special can be used when all dimensions are of types `Categorical`, right?

PR #62, opened 2016-04-21
[MRG] Support "all-categorical" space without one-hot-encoding in gp_minimize

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/56


MechCoder:
@betatim @glouppe This is ready for reviews. Categorical input is now supported, basically

        X = np.asarray([
        ["ham", "spam", "ted"],
        ["ham", "ted", "ted"],
        ["ham", "spam", "spam"]])
       y = rng.randn(3)
       hm = HammingKernel(length_scale=[1.0, 1.0, 1.0])

       gpr = GaussianProcessRegressor(hm)
       gpr.fit(X, y)
       gpr.predict(X)

glouppe:
General comment, can you put the kernel code under `skopt.learning`?

glouppe:
Documentation comment, could you also switch to the docstring format I used everywhere else? (This one does not render properly in Markdown)

skopt/kernels/hamming_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+from sklearn.gaussian_process.kernels import Kernel
+from sklearn.gaussian_process.kernels import NormalizedKernelMixin
+from sklearn.gaussian_process.kernels import StationaryKernelMixin
+
+class HammingKernel(StationaryKernelMixin, NormalizedKernelMixin, Kernel):
+    """
+    The HammingKernel is used to handle categorical inputs.
+
+    ``K(x_1, x_2)`` is given by \sum_{j=1}^{d}exp(-ls_j * (I(x_1j != x_2j)))
+    If all the categorical features of two samples are same then
```
glouppe: are _the_ same

skopt/kernels/hamming_kernel.py
```diff
@@ -0,0 +1,136 @@
+from math import exp
```
glouppe: is there a reason for using `math.exp` instead of `np.exp`?

glouppe:
Thanks! This is a nice first step towards support of categorical variables. This assumes that all features are categorial though, doesn't it? (i.e., we cannot mix categorical and continuous features)

MechCoder:
Actually I ran into a bug while testing the kernel gradient with respect to hyperparameters. I have fixed it in this PR and I have reported it here (https://github.com/scikit-learn/scikit-learn/issues/6701)

MechCoder:
> This assumes that all features are categorial though, doesn't it?

Yes, but we can address the mixed features type in a different PR
In 4.1.2 of http://arxiv.org/pdf/1211.0906v2.pdf, they just advise to find `K` for categorical and continuous features separately and do a pointwise multiplication.

Also, could you make another pass?

glouppe:
This should be rebased on top of master. Then I assume this special kernel can be used when all dimensions are `Categorical`, right? (Not sure what do to for mixed types)

MechCoder:
For mixed-types we can have something as described here https://github.com/scikit-learn/scikit-learn/issues/6868 

(I'll try to address that today)

MechCoder:
@glouppe This does not refactor `gp_optimize` but just adds the kernel. I'm not sure why I should rebase. Could you clarify?

glouppe:
Well indeed! I was just assuming you would also make the necessary changes to `gp_optimize`.

MechCoder:
Let us do incremental development :P

MechCoder:
I started working on the `SelectDimensionKernel` here (https://github.com/MechCoder/scikit-learn/tree/select_kernel)

MechCoder:
Just a FYI, we should push this further due to non-trivial issues in sklearn's implementation.

https://github.com/scikit-learn/scikit-learn/issues/6961
https://github.com/scikit-learn/scikit-learn/issues/6868

glouppe:
Do you wish to finish this for 0.1?

MechCoder:
(That moment when I look back and cannot believe I wrote such pathetic code) :-/

MechCoder:
I still have to write the `gradient_x` method and integrate it with the optimizer functions. Will continue on that tomorrow.

MechCoder:
@glouppe @betatim This is ready for reviews!

skopt/optimizer/optimizer.py
```diff
@@ -195,6 +196,7 @@ def tell(self, x, y, fit=True):
             only be fitted after `n_random_starts` points have been queried
             irrespective of the value of `fit`.
         """
+        is_categorical = all([isinstance(dim, Categorical) for dim in self.space.dimensions])
```
glouppe: Shouldnt this be set once at init instead? Or better, as propery of the `space` instance?

skopt/space/space.py
```diff
@@ -46,6 +46,9 @@ def check_dimension(dimension, transform=None):
         the returned Dimension instance the values are scaled
         between 0 and 1.
 
+        If transform is set to "Identity", when `transform` is called
```
glouppe: `"identity"` (without the cap)

glouppe: also the accepted values for `transform` need to be updated in the docstring just above (line 44)

skopt/space/space.py
```diff
@@ -273,10 +276,18 @@ def __init__(self, categories, prior=None):
...
             are equally likely.
+
+        * `transform` [string, "onehot" or "identity"]:
+            When `transform` is set to "identity", the transformed space
+            is the same as the original space.
         """
         self.categories = categories
-        self.transformer = CategoricalEncoder()
-        self.transformer.fit(self.categories)
+        self.transform_ = transform
```
glouppe: Not sure I like setting both `self.transform` and `self.transform_`. This is confusing. Only `self.transform` seems necessary.

glouppe:
Also Travis complains.

skopt/space/space.py
MechCoder: I did that because `self.transform` is a method.

skopt/optimizer/optimizer.py
MechCoder: I need to check once that all are dimensions are Categorical before instantiating the space, no?

The reason being that if all are Categorical, I need to use a different transform.

MechCoder: I did not realize you were commenting on the `optimizer` diff. I've addressed this.

MechCoder:
updated!

MechCoder:
After doing this, we can add different kernels operating on different dimensions in a separate PR.

glouppe:
Thanks for the changes. Merging.

MechCoder:
Thanks! Will try to implement the `SelectDimensionKernel` either tonight or after I get back in a few days.

betatim:
> Thanks! Will try to implement the SelectDimensionKernel either tonight or after I get back in a few days.

new issue so it doesn't get lost?

PR #69, opened 2016-04-24
Set encoding to UTF-8

AlexanderFabisch:
I cannot run nosetests otherwise (probably a Python 2 issue)


MechCoder:
Thanks! We don't guarantee Python 2 support, though for now.

Issue #59, opened 2016-04-20
API for non continuous inputs


glouppe:
At the moment, input values are assumed to live within a bounded continuous range. We should think about an API on how to specify integer and symbolic values as well, and what would be the consequences for the algorithms we implemented so far.

betatim:
These are two existing ways of specifying such things, `hyperopt` and `GridSearch`:
```python
space = hp.choice('a',
    [
        ('case 1', 1 + hp.lognormal('c1', 0, 1)),
        ('case 2', hp.uniform('c2', -10, 10))
    ])
```

```python
param_grid = [
  {'C': [1, 10, 100, 1000], 'kernel': ['linear']},
  {'C': [1, 10, 100, 1000], 'gamma': [0.001, 0.0001], 'kernel': ['rbf']},
 ]
```

It would be good to adopt an existing one if it isn't too awkward for us. We could reuse some of the code for the sampling etc and as a user I don't have to get used to yet another slightly different way of specifying choices.

glouppe:
Good idea indeed. I would be in favour of following as closely as possible elements from the scikit-learn API. I am not sure if it really fits though, as we would lack a way to specify the scale.

betatim:
An idea:
```python
param_grid = [
  {'C': [1, 10, 100, 1000], # categorical C
   'kernel': ['linear']},
  {'C': skopt.range(1,1000), # continuos C
   'gamma': [0.001, 0.0001], 'kernel': ['rbf']},
 ]
```
If you pass a list/tuple we assume it is categorical, to specify a range you have to use `skopt.range` (better name?). Could also flip it around depending on what is more common use case.

```python
param_grid = [
  {'C': [1, 10, 100, 1000], # categorical C because len()>2
   'kernel': ['linear']},
  {'C': [1,1000], # continuos C because len()==2
   # categorical gamma, special notation to distinguish 2 element categorical
   'gamma': skopt.categorical(0.001, 0.0001),
   # single element or multi element list of non numerical -> categorical
   'kernel': ['rbf'],
   'something_categorical': ['hello', 'world']},
 ]
```

glouppe:
Hmmm, why not. Maybe take the same as in https://github.com/hyperopt/hyperopt/wiki/FMin#21-parameter-expressions but with some syntax shortcuts? (a list == `hp.choice`, a pair == `hp.uniform`)

MechCoder:
[...]

Issue #68, opened 2016-04-23
Ask-and-tell interface?


AlexanderFabisch:
Hi, I just discovered this project. I wonder whether it is really the goal to provide only a scipy-like interface or whether you think it would be possible to provide an [ask-and-tell](https://www.lri.fr/~hansen/collette2010Chap14.pdf) interface, too. That would be much more convenient for use cases in which the optimization process is controlled actually by the objective function.

glouppe:
Am I correct in understanding that it is like a warm start interface?

AlexanderFabisch:
If warm start meant that the state of the optimization algorithm is preserved when it is not running that would be true. The scipy-like implementation based on ask and tell would look like

```python
def myoptimizer_minimize(func, bounds, ...):
    optimizer = MyOptimizer(bounds, ...)

    while not optimizer.stop():
        x = optimizer.ask()
        y = func(x)
        optimizer.tell(x, y)
    ...
```

You could return the optimizer and do a warm start with it.

AlexanderFabisch:
The use case that I have in mind is this:

```python
opt = MyOptimizer(...)
sim = MyComplexSimulation(...)
func = MyFunction(...)

while not opt.stop():
    x = optimizer.ask()

    # Here we start a complex simulation or an experiment in the real world.
    # This is usually not only a single function call and I would like to have
    # the control in my hand and not in the hand of the optimizer.
    result = sim.do_simulation(x, ...)
    y = func(result)

    optimizer.tell(x, y)
```

MechCoder:
Interesting, if I understand correctly the idea is to decouple the function call from the optimization strategy?

Would you be able to provide a more specific example? What would be an example of `do_simulation` when you want a GP to fit the `do_simulation + func` call?

betatim:
I like the idea. You could also use this to support cases where the "function" you want to optimise is a physical process that involves a human in a lab doing something, days later coming back with the answer.

One way of supporting it would be to support passing in a set of parameters `x`s and corresponding `func(x)`s that you somehow obtained previously and run only one iteration that returns the next point at which to sample.

AlexanderFabisch:
> Interesting, if I understand correctly the idea is to decouple the function call from the optimization strategy?
> 
> Would you be able to provide a more specific example? What would be an example of do_simulation when you want a GP to fit the (x, func(do_simulation(x)) call?

[...]

PR #70, opened 2016-04-25
[WIP] API for specifying discrete and categorical variables

betatim:
This is a proposal for an API to specify and sample variables that are categorical and/or discrete.

It follows up on the discussion in #59

The best place to start to learn about the API is `skopt/tests/test_sample.py`

- [ ] Change transform of `Categorical` to do a OneHotEncoding.
- [ ] Add documentation
- [ ] Address inline comments.


betatim:
Need to look at #62 and coordinate with the work there.

betatim:
I am happy with how this looks now. Feedback from @glouppe and @MechCoder on the API would be good. Then I will add some documentation and maybe start modifying the various `XXX_optimize()` functions to use this.

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+class Categorical:
+    def __init__(self, *args):
+        self.values = args
+
+    def rvs(self, random_state=None):
+        rng = check_random_state(random_state)
+        return self.values[rng.randint(len(self.values))]
+
+    @property
+    def low(self):
```
MechCoder: We can store this as an attribute as well, right?

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+from sklearn.utils.fixes import sp_version
+
+
+class Categorical:
+    def __init__(self, *args):
+        self.values = args
+
+    def rvs(self, random_state=None):
+        rng = check_random_state(random_state)
+        return self.values[rng.randint(len(self.values))]
```
MechCoder: Seems like, you can rely on scipy's randint for this case as well.

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+        return max(self.values)
+
+
+class Discrete:
+    def __init__(self, low, high):
+        self.low = low
+        self.high = high
+
+    def rvs(self, random_state=None):
+        return randint(self.low, self.high).rvs(random_state=random_state)
```
MechCoder: I really think we should store the rvs as a private attribute during class instantiation rather than recreating for every rvs call?

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+class Continuous:
+    def __init__(self, low, high):
+        self.low = low
+        self.high = high
+
+    def rvs(self, random_state=None):
+        return uniform(self.low, self.high).rvs(random_state=random_state)
+
+
+def points(grid, random_state=None):
```
MechCoder: `points` -> `search_points` | `sample_points`?

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+    if isinstance(grid, Mapping):
+        grid = [grid]
+
+    for sub_grid in grid:
```
MechCoder: I do not understand this interface | use-case of supplying multiple sub-grids.

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+            if isinstance(v, (Categorical, Continuous, Discrete)):
+                pass
+
+            elif len(v) > 2 or isinstance(v[0], str):
+                sub_grid[k] = Categorical(*v)
+
+            # important to check for Integral first as int(3) is
+            # also a Real but not the other way around
+            elif isinstance(v[0], Integral):
+                sub_grid[k] = randint(*v)
```
MechCoder: Why not `Discrete` here? Just to support lower versions of SciPy? In that case we should be do the handling inside `Discrete` and not outside..

skopt/sample.py
MechCoder: Btw, what do you think about making this as a class similar to `ParameterSampler` in grid_search. During instantiation we can do the "list to "skopt" conversions. Then we can have a `sample` method that returns the points

skopt/sample.py
MechCoder: Hmm, I seem to get the use-case from the example below, which makes me think should we allow `scipy.stats` objects as well?
I would rather keep the API simple, i.e to generate a random point given a `dict of objects` rather than a `list of dict of objects`

skopt/sample.py
```diff
@@ -0,0 +1,79 @@
...
+            elif isinstance(v[0], Real):
+                sub_grid[k] = uniform(*v)
+
+    rng = check_random_state(random_state)
+
+    while True:
+        sub_grid = rng.choice(grid)
+        items = sorted(sub_grid.items())
```
MechCoder: Is this sorting necessary? `params` is a dict and does not have an order anyway.

MechCoder:
@betatim Thanks for the work, the new interface looks great!

I would be happy to merge this as is if @glouppe is happy with the interface. After doing this we should merge https://github.com/scikit-optimize/scikit-optimize/pull/62 . After that we can refactor `gp_minimize` to make use of the new API.

Now how we plan to support categorical inputs for trees etc, is an entirely different question..

skopt/sample.py
betatim: You can pass a `dict of objects` or a `list of dict of objects`. I copied this behaviour from `ParameterGrid` as I thought it was quite nice compared to other methods.

betatim: Don't see a reason not to allow `scipy.stats` if we can think of a use case. The only difficulty is that I couldn't find a way to reliably extract the bounds from a "naked" scipy distribution. Hence the "wrappers" for `Categorical` and `Continuous` (would be nicer if they simply were aliases).

skopt/sample.py
betatim: Overlooked it. Fixed now.

skopt/sample.py
betatim: This comes straight from scikit-learn's `ParameterSampler`. I think the idea is to always iterate over the keys in the same order instead of in a random order.

skopt/sample.py
betatim: ✅ can't remember why I kept it like this, fixed now.

betatim: Actually, we should briefly discuss if it makes sense to "order" categoricals. What is the order of: banana, apple, orange, chocolate?

What is the use case of extracting the bounds now that we can sample directly? Maybe we can get rid of `low` and `high`. This would make it simple to accept arbitrary `scipy.stats` as well.

skopt/sample.py
betatim: How about `random_points` as name? I didn't go for `sample_points` as it would be `sample.sample_points` which seems a bit silly.

Do we need to make this a class? I started with a class and didn't like having to define several double-underscore functions ... I like it as a generator.

betatim:
Is `sample` the best we can do in terms of naming? How about `search_space` or `space` or ??? I probably dislike the last two more than `sample` but I also don't really like `sample` as this is all about specifying the search space and sampling is just one side thing that comes with that.

glouppe:
I'll review tomorrow morning.

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class Categorical:
+    def __init__(self, *args):
+        self.values = args
+        self.low = min(self.values)
+        self.high = max(self.values)
+        self._rvs = randint(0, len(self.values))
+
+    def rvs(self, random_state=None):
```
MechCoder: Can you move the `random_state` version check from below to here? So that doing `Categorical([1, 2, 3, 4]).rvs()` does not raise an error in older versions of SciPy?

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
+from collections import Mapping
+from numbers import Integral
+from numbers import Real
+
+from scipy.stats._distn_infrastructure import rv_frozen
```
MechCoder: unused import

skopt/sample.py
MechCoder: When would you want to pass a `list of dict of objects?` (Not a rhetoric)

I wrongly thought that the idea was to provide some sort of prior. For example, `test_sub_grids` indicates that you want a higher chance of picking a number from 4-5 rather than from 1-4

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+            # important to check for Integral first as int(3) is
+            # also a Real but not the other way around
+            elif isinstance(v[0], Integral):
+                sub_grid[k] = Discrete(*v)
+            elif isinstance(v[0], Real):
+                sub_grid[k] = Continuous(*v)
+
+    rng = check_random_state(random_state)
+
+    while True:
```
MechCoder: We should provide a `maxiter` or `num_points` parameter. What if someone does list(points())?

skopt/sample.py
MechCoder: I admit, I don't understand why it is present over there as well. I removed the line and all tests pass.

Maybe @glouppe has a better idea?

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+
+from sklearn.utils import check_random_state
+from sklearn.utils.fixes import sp_version
+
+
+class Categorical:
+    def __init__(self, *args):
+        self.values = args
+        self.low = min(self.values)
+        self.high = max(self.values)
```
MechCoder: As discussed in the hidden comment, let us just remove `self.low` and `self.high`. They are not being used anywhere.

skopt/sample.py
MechCoder: Ah, looking at the `ParamterGrid`, it is used to support the following use-case

`grid = [{'kernel': ['linear']}, {'kernel': ['rbf'], 'gamma': [1, 10]}]`

I suppose we can do away with supporting this for now.

MechCoder:
I seemed to have missed an important point. How do you propose to tie the `parameters` dict as obtained from `points` to the underlying function `func`? Two quick suggestions.

1. Make `func` to accept `params` as a dict. So the previous

        def func(x):
            return x[0] + 2 * x[1]**2

becomes

        def func(**params):
            return params['x'] + 2 * params['y']**2


2. Let us not do a `dict of objects`, but a `list of objects` instead. So `{'a': Categorical(1, 2), 'b': Discrete(1, 4)}` becomes just [Categorical(1, 2), Discrete(1, 4)]

The only advantage of the first option that I can think of is to allow stuff like https://github.com/scikit-optimize/scikit-optimize/pull/70/files#r61027824 while being less `scipy-like`. The second seems simpler and easier to maintain.

My vote would be for the 2nd option. Curious to know what your thoughts are.

MechCoder:
The point of returning a dict of params for every iteration in `ParameterSampler` | `ParamterGrid` is because it ties well with the `clone`, `set_params`, `get_params` interface of sklearn. Not sure how relevant it is in the more functional approach of this package.

skopt/sample.py
MechCoder: I opened a PR here https://github.com/scikit-learn/scikit-learn/pull/6716

skopt/sample.py
glouppe: +1, and it does not make sense on a "categorical" (== symbolic, unordered) variable

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+        self._rvs = randint(0, len(self.values))
+
+    def rvs(self, random_state=None):
+        return self.values[self._rvs.rvs(random_state=random_state)]
+
+
+class Discrete:
```
glouppe: I would rather call this `Integer`.

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, low, high):
+        self.low = low
+        self.high = high
+        self._rvs = randint(self.low, self.high)
+
+    def rvs(self, random_state=None):
+        return self._rvs.rvs(random_state=random_state)
+
+
+class Continuous:
```
glouppe: I would rather call this `Uniform`

skopt/sample.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, low, high):
+        self.low = low
+        self.high = high
+        self._rvs = uniform(self.low, self.high)
+
+    def rvs(self, random_state=None):
+        return self._rvs.rvs(random_state=random_state)
+
+
+def points(grid, random_state=None):
```
glouppe: This should be named better. It is not clear what it does from the name only.

skopt/sample.py
glouppe: It is important for getting the same results for the same random state. It you dont iterate in the same order, then the `rvs` methods might be called in different orderings, resulting in distinct samples, even if `random_state` is set to the same value in calls.

This should be kept.

glouppe:
This looks very similar to https://github.com/hyperopt/hyperopt/wiki/FMin#21-parameter-expression.
If we want to go for a function-based API (as suggested in #68), hyperopt's API currently looks functional than this proposal. What do you think?

Also, it is not clear to me how do you propose to use the specification of the input space (or "grid") within the minimize functions?

MechCoder:
> Also, it is not clear to me how do you propose to use the specification of the input space (or "grid") within the minimize functions?

What do you think of the suggestions here?
https://github.com/scikit-optimize/scikit-optimize/pull/70#issuecomment-214615655

glouppe:
> What do you think of the suggestions here?
#70 (comment)

I prefer the 2nd option as well.

skopt/sample.py
MechCoder: Integers can be categorical as well. For instance in `sklearn.preprocessing.OneHotEncoder` :P

skopt/sample.py
betatim: This is exactly the use-case I had in mind. You don't think it is a useful use-case?

skopt/sample.py
betatim: For me `Uniform` doesn't convey the idea that it is about continuous values :-/ You could have a uniform random variable over integers as well.

betatim:
Trying to summarise the inline comments that are still open:

* better name for `sample`. `sample_points`? `random_points`?
* better name for the module. `sample` is a bit nondescript as this is also about specifying the input space
* `Discrete`/`Continuous` vs `Integer`/`Uniform`

----

I think I also prefer the `list of objects`. Naming the parameters would make it easier for output and plotting, but I don't think that is worth it. There is `collections.OrderedDict` but that feels a bit OTT.

I prefer to the nested structure

```
[('case 1', Continuous(0,1)), (('case 2', Continuous(-10,10))]
```
compared to
```
hp.choice('a',
    [
        ('case 1', hp.uniform('c1', 0, 1)),
        ('case 2', hp.uniform('c2', -10, 10))
    ])
```
is this difference what you were referring to wrt hyperopt? After getting rid of low/high you can now use any `scipy.stats` distribution (`Continuous` and `Discrete` are just aliases). `Categorical` is still special. For me `Discrete` is clearer than `randint`. So I'd provide three nice names for the most common cases.

betatim:
Proposal for specifying the type and distribution of each parameter. Any of the following should be used when calling `XXX_minimize` instead of the current `bounds` argument.
```
# continous-uniform, continous-loguniform, categorical/symbolic, discrete-uniform
# explicit
(Uniform(0, 13), LogUniform(1,3), Categorical('a', 'b', 'c'), Integer(1, 5))
# implicit
((0., 13.), ('a', 'b', 'c'), (1, 5))
# extended specification for categories and integers
# to give non-uniform weights to each possible value
(Categorical('a', 'b', 'c', p=[0.2, 0.2, 0.6]), Integer(1, 5, p=[0.1, 0.1, 0.2, 0.6]))
```

Interface:
```python
class Distribution:
  def bounds() # bounds, in warped
  def rvs(n_samples=1): # return a random point(s) in original
  def transform(value): # original -> warped
  def inverse_transform(value): # warped -> original

class Uniform(Distribution):
  def __init__(self, low, high):
    self._low = low
    self._high = high
  def bounds(self):
    return (self._low, self._high)
  def transform(self, x):
    return x
  def inverse_transform(self, x):
    return x
  def rvs(self, n_samples=1, random_state=None):
    return scipy.stats.distributions.uniform(self._low, self._high, random_state=random_state)
```

The following should be possible: `dist.transform(dist.rvs(n_samples=12))`

Bounds only make sense in the warped space, especially for categorical values.

MechCoder:
> Proposal.....

The first two are okay with me as well. I can see why the third is needed but we can support it on a "on-need" basis.

Also, I would keep it simple it for now so that we can easily merge this and then add `transform` and `inverse_transform` maybe later.

MechCoder:
As far as naming goes, I would name the module as `search.py` and the function as `search_points` inspired from sklearn's `GridSearchCV`, `RandomizedSearchCV` etc.

Continuous can imply any distribution but we are doing a uniform sampling. Which is why I seem to prefer the Integer | Uniform interface to Discrete | Continuous but it looks like the majority is already in favour of that.

Please ping back when you rewrite the API to support list of objects interface

MechCoder:
@betatim Any news on this? Or are you too busy with your startup :-)

betatim:
Thanks for reminding me! github-notification-driven-development ;)

> By transform for Categorical, do you mean a wrapper around `LabelEncoder`? We can always refactor that later if required.

Yes exactly.

betatim:
Made a start on this, still missing a lot (docs!).

 Thoughts and opinions on: `OneHotEncoder` vs `LabelEncoder`? Used the latter but now thinking you'd want `OneHotEncoder`.

@glouppe do you remember what our thinking was on how to support input space warping via `transform()`? I think we concluded that eg `LogUniform` would sample uniformly but that `transform()` would take the `log`.

```python
dist = LogUniform(0.1, 100)
dist.rvs(n_samples=3) # -> 0.11, 99, 50
dist.transform(dist.rvs(n_samples=3)) # -> -1.0, 2.0 1.7 (rounded)
```

and that the minimisation algorithms would use the `transform()`ed values.

MechCoder:
I wouldn't have time to give a detailed pass till next week at least. So if you and @glouppe come to some consensus and the work is complete by then, feel free to merge.

glouppe:
`LabelEncoder` is fine.

glouppe:
> @glouppe do you remember what our thinking was on how to support input space warping via transform()? I think we concluded that eg LogUniform would sample uniformly but that transform() would take the log.

The opposite in fact. `rvs` should sample according to the distribution (log uniformly). 

Then `transform` should be use to warp the original space (where here the density is contained in the first decimals) to a warped space (where now the density is evenly distributed in the range), effectively by taking the log.

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+            elif len(dist) > 2 or isinstance(dist[0], str):
+                sub_grid_[i] = Categorical(*dist)
+
+            # important to check for Integral first as int(3) is
+            # also a Real but not the other way around
+            elif isinstance(dist[0], Integral):
+                sub_grid_[i] = Integer(*dist)
+            elif isinstance(dist[0], Real):
+                sub_grid_[i] = Uniform(*dist)
```
glouppe: Maybe be best to factor out the first half of this function into a `check_grid` util.

skopt/tests/test_parameter.py
```diff
@@ -0,0 +1,86 @@
...
+    assert_equal(x.rvs(random_state=1), random_val)
+
+
+def test_distributions():
+    yield (check_distribution, Uniform, (1., 4.), 2.668088018810296)
+    yield (check_distribution, Uniform, (1, 4), 2.668088018810296)
+    yield (check_distribution, Integer, (1, 4), 2)
+    yield (check_distribution, Integer, (1., 4.), 2)
+    yield (check_distribution, Categorical, ('a', 'b', 'c', 'd'), 'b')
+    yield (check_distribution, Categorical, (1., 2., 3., 4.), 2.)
```
glouppe: Might be best to check that Uniform.rvs returns a float, that Integer.rvs returns and that Categorical(["a", "b", ...]) returns a string, all within the given bounds/set. What do you think?

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+    def transform(self, values):
+        """Transform `values` from original into warped space."""
+        return values
+    def inverse_transform(self, values):
+        """Transform `values` from warped into original space."""
+        return values
+
+
+class Uniform(Distribution):
+    def __init__(self, low, high):
```
glouppe: We should make clear whether the upper bound is inclusive or not.

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+        self._rvs = uniform(self._low, self._high)
+
+    def bounds(self):
+        return (self._low, self._high)
+
+    def rvs(self, n_samples=None, random_state=None):
+        return self._rvs.rvs(size=n_samples, random_state=random_state)
+
+
+class Integer(Distribution):
```
glouppe: I would rather be explicit and rename this to `UniformInteger`

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self, low, high):
+        self._low = low
+        self._high = high
+        self._rvs = uniform(self._low, self._high)
+
+    def bounds(self):
+        return (self._low, self._high)
+
+    def rvs(self, n_samples=None, random_state=None):
+        return self._rvs.rvs(size=n_samples, random_state=random_state)
```
glouppe: Shall pass `random_state` through `sklearn.utils.check_random_state`?

skopt/tests/test_parameter.py
```diff
@@ -0,0 +1,86 @@
...
+        assert_equal(p, expected[i])
+
+
+def test_categorical_transform():
+    categories = ['apple', 'orange', 'banana']
+    cat = Categorical(*categories)
+
+    # LabelEncoder sorts classes alphabetically -> banana == 1
+    assert_array_equal(cat.transform(['apple', 'banana']), [0, 1])
+    assert_array_equal(cat.inverse_transform([0, 1]), ['apple', 'banana'])
```
glouppe: Hmm, after all, one-hot-encoding is maybe a better strategy. LabelEncoder forces labels to be ordered arbitrarily, which will not make any sense further ahead during optimization.

skopt/parameter.py
MechCoder: Is that neceesary?

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+        self.categories = categories
+        self._rvs = randint(0, len(self.categories))
+        self._transformer = LabelEncoder()
+        self._transformer.fit(self.categories)
+
+    def rvs(self, n_samples=None, random_state=None):
+        choices = self._rvs.rvs(size=n_samples, random_state=random_state)
+        if n_samples is None:
+            return self.categories[choices]
+        else:
```
MechCoder: Should we convert self.categories to a NumPy array and do fancy indexing?

skopt/tests/test_parameter.py
MechCoder: +1

skopt/tests/test_parameter.py
```diff
@@ -0,0 +1,86 @@
...
+    yield (check_distribution, Categorical, ('a', 'b', 'c', 'd'), 'b')
+    yield (check_distribution, Categorical, (1., 2., 3., 4.), 2.)
+
+
+def test_simple_grid():
+    expected = [(2, 1), (1, 2), (2, 2), (2, 1), (1, 2)]
+
+    for i,p in enumerate(sample_points([(1, 3), (1, 4)],
+                                       len(expected), random_state=1)):
+        assert_equal(p, expected[i])
```
MechCoder: You can use `assert_array_equal` right?

skopt/tests/test_parameter.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def test_simple_categorical():
+    expected = [(2, 1), (1, 2), (2, 2), (2, 1), (1, 2)]
+
+    for i,p in enumerate(sample_points([Categorical(1, 2), (1, 4)],
+                                       len(expected), random_state=1)):
+        assert_equal(p, expected[i])
```
MechCoder: It's odd that the test above and this return the same results :P

MechCoder: You can maybe increase the length of the interval in the above and check that
1. Every value lies within the bounds if the length is equal to 2
2. Every value is always in each tuple, if Categorical is explicitly provided or length of each tuple is greater than 2.

betatim:
After a chat with @glouppe on Tuesday I thought I had a better idea how to handle the transformations but now unsure again. The two things that are getting clear to me now is that there are two things:

1. your prior belief about what values are good ones to sample
1. transforming the space of the parameter to make it easier for the optimiser

The idea on Tuesday was to have an argument `transformer` with which you can transform the space. 

Trying to wrap my head around the case where a parameter has a very large range (0.1 -> 10^6) but all the action is at less than 10 (say). In this case you want to use `log` as transform. You also wouldn't want to sample uniformly between 0.1 and 10^6 though, instead you'd sample according to an exponential(??). Something that has a lot of mass at small values. Does tat seem sensible?

Will slowly work on addressing the inline comments.

MechCoder:
@betatim I might have some time to spend on this right now. Would you be able to address the inline comments?

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Integer(Distribution):
+    def __init__(self, low, high):
+        self._low = low
+        self._high = high
+        self._rvs = randint(self._low, self._high)
+
+    def bounds(self):
+        return (self._low, self._high)
```
MechCoder: not to nitpick: but the brackets are redundant.

MechCoder: not to nitpick: but the brackets are redundant.

skopt/parameter.py
```diff
@@ -0,0 +1,109 @@
...
+    for n in range(n_points):
+        sub_grid = grid_[rng.randint(0, len(grid_))]
+
+        params = []
+        for dist in sub_grid:
+            if sp_version < (0, 16):
+                params.append(dist.rvs())
+            else:
+                params.append(dist.rvs(random_state=rng))
+        yield tuple(params)
```
MechCoder: Is it necessary that this is immutable?

MechCoder:
With respect to the transform, I also believe that onehotencoding is the more reasonable thing to do.

MechCoder:
Note that one hot encoding does not handle string labels. So you should label encode first and then one hot encode again.

+1 for merge except the updated TODO list. (Also let me know if you are a bit busy to address them, I can send a follow-up PR to address them myself)

betatim:
I've been experimenting with this on and off and the more I fiddle the more unhappy I am :( Especially one-hot encoding string categories. I can't come up with a nice way of doing this. So maybe we skip categories for the moment?

I pushed the current state of my branch so we have it here, but I think the best way forward is to abandon this PR and start over fresh (hinthint Tim won't have time hinthint 😄 ).

skopt/parameter.py
```diff
@@ -0,0 +1,258 @@
...
+        return self
+
+    def transform(self, values):
+        return np.log(values)
+
+    def inverse_transform(self, values):
+        return np.exp(values)
+
+
+class CategoryTransform(TransformerMixin):
```
betatim: This is a bit of a mess, so might be most useful as an anti-pattern

MechCoder:
Merging this for now.

Hopefully you will be around to review my follow-up PR.

Issue #60, opened 2016-04-20
Unify GBRT and GP interfaces


betatim:
Based on: https://github.com/scikit-optimize/scikit-optimize/pull/34#discussion_r60404019

We should aim to converge on a unified interface where possible. It is too much work to duplicate all the acquisition functions etc.

MechCoder:
Would you be able to handle this?

betatim:
First we need some ideas on how to do it :)

MechCoder:
My thoughts.

1. For now, we should have a common `gaussian_acquisition_function` since for both the conditionals are estimated to be gaussians. I would refactor the `_expected_improvement` in `gbrt_opt.py` into this.
2. For this functions, the `base_estimator` should have a `return_std` parameter.

wdyt?

glouppe:
Is this still an issue?

MechCoder:
only on github....

PR #71, opened 2016-04-26
Refactor _ei_improvement into gaussian_acquisition

MechCoder:
Allow `return_std` parameter for GBRT.predict
Move `_ei_improvement` and `acquisition` into `gaussian_acquisition` which suggests that the conditional is a gaussian.

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -5,77 +5,168 @@ import numpy as np
 from scipy.stats import norm
 
 
-def gaussian_acquisition(X, model, y_opt=None, method="LCB",
-                         xi=0.01, kappa=1.96):
+def _gaussian_acquisition(X, model, y_opt=None, method="LCB",
+                          xi=0.01, kappa=1.96):
     """
-    Returns the acquisition function computed at values x0 where the
-    conditional is assumed to be a Gaussian with the mean and std
-    provided by the underlying base estimator.
-
-    Parameters
-    ----------
-    X : array-like
-        Values where the acquisition function should be computed.
-
-    model: sklearn estimator that implements predict with ``return_std``
-        The fit sklearn gaussian process estimator that approximates
-        the function. It should have a ``return_std`` parameter
-        that returns the standard deviation.
-
-    y_opt: float, optional
-        The previous best value over which we want to improve.
-        Useful only when `acq` is set to "EI"
-
-    method: string, "LCB" or "EI"
-        If set to "LCB", then the lower confidence bound is taken.
-        If set to "EI", then the expected improvement condition is taken.
-
-    xi: float, default 0.01
-        Controls how much improvement one wants over the previous best
-        values.
-
-    kappa: float, default 1.96
-        Controls how much of the variance in the predicted values should be
-        taken into account. If set to be very high, then we are favouring
-        exploration over exploitation and vice versa.
-        Useless if acq is set to "EI"
-
-    Returns
-    -------
-    values: array-like, length X
-        Acquisition function values computed at X.
+    Wrapper so that the output of this function can be
+    directly passed to a minimizer.
     """
     # Check inputs
     X = np.asarray(X)
     if X.ndim != 2:
         raise ValueError("X should be 2-dimensional.")
 
+    # Evaluate acquisition function
+    if method == "LCB":
+        return gaussian_lcb(X, model, kappa)
+    elif method == "EI":
+        return -gaussian_ei(X, model, y_opt, xi)
+    elif method == "PI":
+        return -gaussian_pi(X, model, y_opt, xi)
+    else:
+        raise ValueError("Acquisition function not implemented.")
+
+    return values
+
+
+def gaussian_lcb(X, model, kappa=1.96):
+    """
+    Use the lower confidence bound to estimate the acquisition
+    values.
+
+    The trade-off between exploitation and exploration is left to
+    be controlled by the user through the parameter ``kappa``.
+
+    Parameters
+    ----------
+    * `X` [array-like, shape=(n_samples, n_features)]:
+        Values where the acquisition function should be computed.
+
+    * `model` [sklearn estimator that implements predict with ``return_std``]:
+        The fit estimator that approximates the function through the
+        method ``predict``.
+        It should have a ``return_std`` parameter that returns the standard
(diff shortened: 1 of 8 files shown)


MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/60

@betatim @glouppe Reviews?

glouppe:
Thanks! But while you are refactoring this, what would you think of splitting `gaussian_acquisition` into several actual acquisition functions? (e.g. `gaussian_expected_improvement`, `gaussian_lower_confidence_bound`, etc)

MechCoder:
Sounds like a good idea.

MechCoder:
@glouppe Done. I hope the new documentation is more helpful.

skopt/learning/gbrt.py
```diff
@@ -51,6 +51,31 @@ def fit(self, X, y):
...
+        predicted_quantiles = np.asarray(
+            [rgr.predict(X) for rgr in self.regressors_])
+        if not return_std:
+            return predicted_quantiles.T
+        else:
+            std_quantiles = [0.16, 0.5, 0.84]
+            is_present_mask = np.in1d(std_quantiles, self.quantiles)
+            if not np.all(is_present_mask):
+                raise ValueError(
+                    "return_std works only if the quan")
```
betatim: unfinished sentence ;)

skopt/learning/gbrt.py
```diff
@@ -51,6 +51,31 @@ def fit(self, X, y):
...
+            and `n_features` is the number of features.
+        """
+        predicted_quantiles = np.asarray(
+            [rgr.predict(X) for rgr in self.regressors_])
+        if not return_std:
+            return predicted_quantiles.T
+        else:
+            std_quantiles = [0.16, 0.5, 0.84]
+            is_present_mask = np.in1d(std_quantiles, self.quantiles)
+            if not np.all(is_present_mask):
```
betatim: This is neat! Hadn't seen this before

skopt/acquisition.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def gaussian_ei(X, model, y_opt=None, xi=0.01):
+    """
+    Use the expected improvement to calculate the acquisition values.
+
+    The predictions form a gaussian with a certain mean and standard deviation
+    as given by the model.
+
+    The PI condition is derived by computing ``E[u(f(x))]``
```
betatim: Is this duplication from L117 on purpose?

MechCoder: The line numbers got messed up because I pushed. What were you referring to?

skopt/acquisition.py
```diff
@@ -0,0 +1,203 @@
+import warnings
+
+import numpy as np
+
+from scipy.stats import norm
+
+
+def gaussian_acquisition(X, model, y_opt=None, method="LCB",
```
glouppe: Do we still need this function? Couldnt this be part of the minimize function instead?

skopt/acquisition.py
```diff
@@ -0,0 +1,203 @@
...
+
+    return mu - kappa * std
+
+
+def gaussian_pi(X, model, y_opt=0, xi=0.01):
+    """
+    Use the probability of improvement to calculate the acquisition values.
+
+    The predictions form a gaussian with a certain mean and standard deviation
+    as given by the model.
```
glouppe: I would rather say something like "The conditional probability p(y=f(x)|x) is assumed to be gaussian distributed, with mean and standard deviation approximated by `model`."

skopt/acquisition.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        mu, std = model.predict(X, return_std=True)
+
+    values = np.zeros_like(mu)
+    mask = std > 0
+    improvement = y_opt - xi - mu[mask]
+    values[mask] = norm.cdf(improvement / std[mask])
+    return -values
```
glouppe: I am wondering if you should negate the value at this place, rather than in `minimize`. It would be cleaner not to do it here in my opinion, so that `gaussian_pi` implements the correct definition of probability of improvement.

skopt/acquisition.py
```diff
@@ -0,0 +1,203 @@
...
+        warnings.simplefilter("ignore")
+        mu, std = model.predict(X, return_std=True)
+
+    values = np.zeros_like(mu)
+    mask = std > 0
+    improvement = y_opt - xi - mu[mask]
+    exploit = improvement * norm.cdf(improvement / std[mask])
+    explore = std[mask] * norm.pdf(improvement / std[mask])
+    values[mask] = exploit + explore
+    return -values
```
glouppe: same comment

skopt/acquisition.py
betatim: There were two bits of docs that contained "The PI condition is derived ..." but it is fixed now.

skopt/acquisition.py
MechCoder: But then we will have to write a `if else` block for every new function added. Are you okay with that?

MechCoder: It makes more sense to make this private, I guess

MechCoder:
@glouppe @betatim Addressed your comments. Please merge if happy so that we can move on to other interesting stuff

skopt/tests/test_acquisition.py
```diff
@@ -0,0 +1,44 @@
...
+
+from skopt.acquisition import gaussian_ei
+from skopt.acquisition import gaussian_lcb
+from skopt.acquisition import gaussian_pi
+
+
+class ConstSurrogate:
+    def predict(self, X, return_std=True):
+        return np.zeros(X.shape[0]), np.ones(X.shape[0])
+
```
betatim: `pep8` 🚓  two blank lines please

glouppe:
+1 for merge on my side

betatim:
This looks fine. Unsure what I think about `gaussian_acquisition` as a public or private method. It would be convenient, especially for an interface that is more DIY style like #68, to provide something like this to users so that not everyone has to make the `if else` block by hand. I think we should wait, gather some more experience and then decide. Right now the API is still evolving so we can easily switch from one to the other (right?).

PR #72, opened 2016-04-28
Applying the flake8 steamroller

betatim:
Down to to one complaint about a long line in `skopt/__init__.py` don't think
there is much we can do about that if we want to have the badge.

For future use I used `flake8 skopt/*`.


glouppe:
+1

MechCoder:
Thanks!

betatim:
🍰

PR #75, opened 2016-06-03
[MRG] Cleanup skopt/parameter.py with docs and some minor changes

MechCoder:
Added docs and some minor cosmetics.
Removed prior for now so that we can compare our results with a randomized search where we have some prior knowledge about the candidates. (I understand it might be useful but YAGNI)


MechCoder:
@AlexanderFabisch @fabianp Since you have used the library to an extent, we would like to hear your thoughts on this. The idea is to support function calls like this at some point of time

    gp_optimize(func, [Real(2, 4), ["red", "green"]])

`bounds` will be either a list of Distribution objects or lists. If it is a list of lists, then the type of parameter will be inferred.

(I am just cleaning up the work of @betatim done here https://github.com/scikit-optimize/scikit-optimize/pull/70)

AlexanderFabisch:
That's a good idea. Here are some suggestions:

* Maybe the argument name `bounds` is a little bit misleading for `[Real(2, 4), ["red", "green"]]`. You could introduce an additional argument `param_distributions`.
* `Real` should actually be `UniformReal`. There could be `NormalReal` or `LogReal` as well like (similar to [hyperopt](https://github.com/hyperopt/hyperopt/wiki/FMin#21-parameter-expressions)).

However, my main focus at the moment is the optimization of real vectors so I actually don't really care about this use case.

AlexanderFabisch:
By the way, is it possible to set a seed in scipy.stats distributions? It would be much better if you could simply pass a seed for the random number generator in the function call like in sklearn.

betatim:
I would vote to keep it `Real` (hahaha...) and support non uniform priors via `Real(..., prior=dist.norm(..))` or `Real(..., prior=dist.exp(..))` etc. At least that was my thinking about having a `prior` parameter. What do you think?

To me it seems worth having the prior and the transform separate. The prior represents where you think likely good values are and the transform takes care of turning the values into something that is "ncie" for the optimiser to handle (onehot encoding, similar ranges)

fabianp:
Hi, thanks for the heads up.

I'm general I'm skeptical for the need of such framework (but I could be wrong). In the usercases I'm familiar with you never need more than to set bounds and/or optionally transform to logspace, so this might be overkill. I know RandomizedSearchCV takes a distribution argument, but I've never used it for anything else besides uniform sampling (sometimes in log space).

MechCoder:
Thanks a lot for your feedback.

@fabianp By skeptical, are you skeptical about the use of priors or the support for categorical parameters? A common use case would be the type of scaling used or the kernel in SVM (I remember the paper on randomized search had something on those lines). 

I also think it would be okay to support priors via arguments later.

@AlexanderFabisch Yes, we would pass a random seed to the function call which will then be passed internally to the rvs method and then sampled (as being done currently at https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/gp_opt.py#L121). Were you meaning this or something else?

MechCoder:
@betatim Please review and if you are happy, do merge. In follow-up PR's I can address the remaining.

AlexanderFabisch:
> @AlexanderFabisch Yes, we would pass a random seed to the function call which will then be passed internally to the rvs method and then sampled (as being done currently at https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/gp_opt.py#L121). Were you meaning this or something else?

That is what I meant.

glouppe:
+1 for decoupling the domain/type of a variable (i.e., real, integer, categorical) from how it should be converted before being fed to the optimizer. 

Overall, things should be very simple by default, e.g ``Real(2, 4))`` (no transform, uniform distribution), but extendable in a plug and play manner with optional arguments, e.g. ``Real(2, 4, transformer="log", prior="gaussian")``.

glouppe:
Also I agree that we should start by having something very simple to begin with, i.e. supporting real integer and categorical types, all assumed to be uniformly distributed, with an optional log transform for the real values.

glouppe:
Is this ready for review? I am confused as this PR is adding `parameter.py` while `parameters.py` (with an `s`) is already there.

glouppe:
Could you instead base your PR on master? It is otherwise difficult to compare what you are actually changing. Not sure what was wrong with Tim's proposal.

skopt/parameter.py
```diff
@@ -151,6 +178,8 @@ def __init__(self, low, high, prior='uniform', transformer='identity'):
 
         if transformer == 'identity':
             self.transformer = Identity()
+        elif transformer == 'log':
+            self.transformer = Log()
```
glouppe: Does it make sense to allow log transform for integers? Not sure we should support it, or at least the inverse transform should be cast back to integers.

MechCoder: Maybe not or at least I can't think of a quick use case.

MechCoder:
There are still tests to be added.

MechCoder:
Checkout the bug in the last commit :P

betatim:
Woah! Thanks for catching this.

skopt/parameter.py
```diff
@@ -50,59 +51,77 @@ def inverse_transform(self, values):
...
+        return self.transformer.transform(values)
+
+    def inverse_transform(self, values):
+        """Transform `values` from warped into original space."""
+        return self.transformer.inverse_transform(values)
+
+    @abc.abstractmethod
+    def rvs(self, n_samples=None, random_state=None):
+        """
+        Sample points randomly.
```
betatim: Change to "Randomly sample points from the original space" to explain that the samples are not in the warped space.

skopt/parameter.py
```diff
@@ -112,15 +131,23 @@ def __init__(self, low, high, prior='uniform', transformer='identity'):
...
+        elif transformer == "log10":
+            self.transformer = Log10()
         elif isinstance(transformer, TransformerMixin):
             self.transformer = transformer
         else:
             raise RuntimeError('%s is not a valid transformer.'%transformer)
 
         if prior == 'uniform':
-            self._rvs = uniform(self._low, self._high)
+            self._rvs = uniform(self._low, self._high - self._low)
```
betatim: Thanks for defusing this 💣

skopt/parameter.py
```diff
@@ -174,39 +205,36 @@ def __init__(self, *categories, prior=None, transformer='labels'):
...
+        * `prior` [array-like, shape=(categories,), default None]:
             Prior probabilities for each category. By default all categories
             are equally likely.
 
-        transformer: string or fitted TransformerMixin, default 'labels'
+        * `transformer` [string or fitted TransformerMixin, default 'onehot']:
             Transformer to convert between original and warped search space.
             Parameter values are always transformed before being handed to the
-            optimizer. Defaults to `LabelEncoder`.
+            optimizer. Defaults to `OneHotEncoder`.
```
betatim: `OneHotEncoder` -> `CategoryTransform`?

skopt/parameter.py
```diff
@@ -50,59 +51,77 @@ def inverse_transform(self, values):
...
-        high : float
+        * `high` [float]:
             Upper bound of the parameter.
 
-        prior : string or rv_frozen, default 'uniform'
+        * `prior` [string or rv_frozen, default='uniform']:
             Distribution to use when sampling random points for this parameter.
 
-        transformer: string or fitted TransformerMixin, default 'identity'
+        * `transformer` [instance of TransformerMixin, "identity", "log", "log10", default='identity']:
```
betatim: I prefer the comment from L171: 

```
* `transformer` [string or fitted TransformerMixin, default=....
```

that makes it clear the transformer needs to be fitted (if it needs fitting).

skopt/tests/test_parameter.py
```diff
@@ -16,57 +18,74 @@ def check_distribution(Dist, vals, random_val):
 
 
 def test_distributions():
-    yield (check_distribution, Real, (1., 4.), 2.668088018810296)
-    yield (check_distribution, Real, (1, 4), 2.668088018810296)
+    a = Real(1, 25)
```
betatim: Dead code?

skopt/tests/test_parameter.py
```diff
@@ -16,57 +18,74 @@ def check_distribution(Dist, vals, random_val):
...
     yield (check_distribution, Integer, (1., 4.), 2)
     yield (check_distribution, Categorical, ('a', 'b', 'c', 'd'), 'b')
     yield (check_distribution, Categorical, (1., 2., 3., 4.), 2.)
 
 
-def test_simple_grid():
-    expected = [(2, 1), (1, 2), (2, 2), (2, 1), (1, 2)]
+def check_distribution_limits(value, lower_bound, upper_bound):
+    assert_less_equal(lower_bound, value)
+    assert_greater_equal(upper_bound, value)
```
betatim: `assert_greater` as the upper bound is not included. At least `np.alltrue(uniform(2, 12.-2).rvs(1000000)  < 12)` -> `True`

skopt/tests/test_parameter.py
```diff
@@ -16,57 +18,74 @@ def check_distribution(Dist, vals, random_val):
...
-def test_sub_grids():
-    expected = [('a', 1), ('a', 2), ('b', 5), ('b', 5), ('a', 1)]
+#     assert_array_equal(dist.transform(values), values)
+#     assert_array_equal(dist.inverse_transform(values), values)
 
-    for i,p in enumerate(sample_points([(['a'], (1, 4)), (['b'], (4, 6))],
-                                       len(expected), random_state=3)):
-        assert_equal(p, expected[i])
 
+# def test_simple_grid():
```
betatim: Is the rest a place holder for more work? Otherwise we should delete it and add it back later.

skopt/tests/test_parameter.py
```diff
@@ -16,57 +18,74 @@ def check_distribution(Dist, vals, random_val):
...
+#     yield (check_simple_grid, (1., 4.), [2.668, 3.881], Real)
+#     yield (check_simple_grid, (1, 2, 3), [2, 3], Categorical)
+
+
+# def test_sub_grids():
+#     expected = [('a', 1), ('a', 2), ('b', 5), ('b', 5), ('a', 1)]
+
+#     for i,p in enumerate(sample_points([(['a'], (1, 4)), (['b'], (4, 6))],
+#                                        len(expected), random_state=3)):
+#         assert_equal(p, expected[i])
```
glouppe: Why removing these tests? Is it simply because it is still a work in progress? We need tests for `check_grid`, as it will further be the entry point to validate and check user parameters.

MechCoder: Oh sorry, it is a WIP. I'll finish it up tonight

skopt/parameter.py
```diff
@@ -50,59 +51,77 @@ def inverse_transform(self, values):
...
-        low : float
+        * `low` [float]:
             Lower bound of the parameter.
 
-        high : float
+        * `high` [float]:
             Upper bound of the parameter.
 
-        prior : string or rv_frozen, default 'uniform'
+        * `prior` [string or rv_frozen, default='uniform']:
```
MechCoder: Should we remove this argument for now? If the user provides say a Gaussian or Poisson prior, the low and high arguments become meaningless.

MechCoder:
@glouppe @betatim 

I'm slightly confused on the API of using  `Real` with say a `log` transformer vs using `Categorical` with a `one-hot` transformer. I would like to know your thoughts.

When using `Categorical` with a one-hot transformer, it makes sense to sample in the original space and then transform.

But when using `Real` it makes more sense to do this.

Original space: Bounds = (10**-5, 10**5)
Sample uniformly from (log10(10**-5), log10(10**5)) = (-5, 5)
Raise to the power 10, when providing it to the optimizer.

That is sampling in the warped space, rather than the original space.

MechCoder:
Does it make sense to do away with the `transform` method, and assume that `rvs` provides whatever needed, which can be directly fed to the optimizer? I believe it save us a lot of confusion. These are my proposed changes.

```python
a = Real(0, 10)
a.rvs() # Sample uniformly from 0 to 10
a = Real(10**-5, 10**5, transformer="log10")
a.rvs() # log(Sample uniformly from -5 to 5)

a = Categorical(["a", "b", "c"], transformer="one-hot")
a.rvs() # One-hot-encode (randomly selected from ["a", "b", "c"])
a = LabelEncoder(["a", "b", "c"], transformer="label")
a.rvs() # LabelEncode (label encode randomly selected from ["a", "b", "c"]
```

MechCoder:
Commenting again since the previous comment was hidden from the diff view.

Also does it make sense to remove the prior argument? Since giving a say `gaussian` or `poisson` (or anything other than uniform) prior makes the `low` and `high` values meaningless.

glouppe:
Two things:

- The optimizer should operate in the warped space and be agnostic of that transformation. It is given bounds and points and proceed with that. No additional assumption are required. 
- We need decide on the definition of the prior, i.e. whether it concerns the original or the warped space. I am fine with i) deciding it concerns the warped space, ii) assuming a uniform prior by default.

MechCoder:
Just to make sure that we are on the same page, do you agree that the current behaivour of using `transformer="log"` with `prior="uniform"` is incorrect?

MechCoder:
The terminology is still not crystal clear to me.

Let us say `alpha=(10**-5, 10**5)`, the warped space will be from `(-5, 5)`.

It seems logical to sample from (-5, 5), raise it to the power of 10 and then provide it to the optimizer, no?

In that case the optimizer is not "operating in the warped space" right?

glouppe:
> It seems logical to sample from (-5, 5), raise it to the power of 10 and then provide it to the optimizer, no?

No. I would feed values from (-5, 5) to the optimizer, and then use `inverse_transform` to convert back the found solution. Otherwise, you would still have these scaling issues at optimization.

glouppe:
> Just to make sure that we are on the same page, do you agree that the current behaivour (output) whentransformer="log" is used with prior="uniform" for Real is incorrect?

If the prior concerns the warped space, as I suggest above, then it would just amount to uniformly space in log-space.

MechCoder:
Great, I'll change it then.

betatim:
My thinking behind splitting the prior and transformer was that there are two problems that need solving: 

1. how to transform values that humans like (eg strings to label categories) to values that optimisers like (eg one hot encoded categories)
2. express how I believe the most likely/useful values of a parameter are distributed (eg gaussian around 5123 with std of 42)

Personally I find it easier to think about the prior in the original space and then let the computer transform it to the warped space for the optimiser: `gauss(5123, std=42)` plus `StandardScaler()` to get it to a sensible range for the optimiser. Maybe this is overkill/just my weird brain :)

MechCoder:
Yes, which is why it makes sense to keep sampling in the warped space just for the uniform prior.

glouppe:
Proposal that could simplify things:

Let us hide the transform mechanism from the external API + Let define both "uniform" and "log-uniform" priors for `Real`.

glouppe:
Proposal:

1) We hide the transformation mechanism. This is something internal and users should not deal with it.

2) We define two hardcoded priors for now, `"uniform"` and `"log-uniform"`. In the later case, everything is done internally so that the optimizer deal with values in log-space, and then converted back to the original space.

MechCoder:
@glouppe @betatim I've done as you have suggested. The tests should be the best way to understand the new API.

skopt/parameter.py
```diff
@@ -1,20 +1,19 @@
...
-from scipy.stats.distributions import rv_frozen
 from scipy.stats.distributions import uniform
 
-from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.preprocessing import LabelEncoder, OneHotEncoder
 from sklearn.utils import check_random_state
 from sklearn.utils.fixes import sp_version
 
 
-class Identity(TransformerMixin):
```
MechCoder: Removed because `TransformerMixin` assumes input to transform to be 2-D

skopt/parameter.py
```diff
@@ -38,178 +37,156 @@ def inverse_transform(self, values):
...
         return self
 
     def transform(self, values):
-        vals = np.asarray(values)
-        vals = self._label.transform(vals).reshape(-1, 1)
-        return self._onehot.transform(vals).toarray()
+        """
+        Transform an array of categories to a one-hot encoded representation.
 
-    def inverse_transform(self, values):
```
MechCoder: Removed because `OneHotEncoder` does not have an inverse transform

MechCoder:
I have tried to keep the API as simple as possible.

fabianp:
I like where this is going :-)

skopt/parameter.py
```diff
@@ -38,178 +37,156 @@ def inverse_transform(self, values):
...
 
-class Log(TransformerMixin):
-    """Natural logarithm transform."""
-    def fit(self, values):
-        return self
+class Distribution:
 
-    def transform(self, values):
-        return np.log(values)
+    @abc.abstractmethod
```
glouppe: Do we really need `abc` and stuff? A simple `raise NotImplementedError` wouldnt be enough?

(This is an open question, I have no strong opinion against one or the other.)

glouppe:
Thanks Manoj, this is looking good!

I think we should add a last test showcasing how non-trivial grids can be defined. E.g., by calling `sample_points` for grids defined in different but equivalent formats (e.g. `Real(0, 1)` and `(0., 1.)`) and checking for the same random_states, they yield the same points.

glouppe:
For ease of use, we should also accept triples for `Real`. That is:

```
(0., 1.) -> Real(0., 1., prior="uniform")
(0., 1., "log-uniform") -> Real(0., 1., prior="log-uniform")
```

skopt/parameter.py
betatim: I am -0 on `abc`

skopt/parameter.py
betatim: Do you know why it doesn't have `inverse_transform`? I can't think of a reason other than "hasn't been implemented yet".

How do we present the result of the optimisation to the user if there is no `inverse_transform`? I like @glouppe's idea of hiding the transform from the user, but in that case we need to `inverse_transform` :-/

skopt/parameter.py
```diff
@@ -38,178 +37,156 @@ def inverse_transform(self, values):
...
-            self._rvs = prior
+            raise ValueError(
+                "prior should be either 'uniform' or 'log-uniform' "
+                "Got %s" % self._rvs)
 
     def rvs(self, n_samples=None, random_state=None):
-        return self._rvs.rvs(size=n_samples, random_state=random_state)
+        random_vals = self._rvs.rvs(size=n_samples, random_state=random_state)
+        if self.prior == "log-uniform":
+            return np.exp(random_vals)
```
betatim: `np.exp(x)` -> `10**x` to match the `np.log10` done earlier.

skopt/parameter.py
```diff
@@ -38,178 +37,156 @@ def inverse_transform(self, values):
...
+            prior = np.tile(1. / len(self.categories), len(self.categories))
         self._rvs = rv_discrete(values=(range(len(self.categories)), prior))
 
     def rvs(self, n_samples=None, random_state=None):
         choices = self._rvs.rvs(size=n_samples, random_state=random_state)
-        if n_samples is None:
-            return self.categories[choices]
-        else:
-            return [self.categories[choice] for choice in choices]
+        return self.categories[choices]
```
glouppe: @betatim The inverse transform is actually done here. Might be better indeed to move this step (from integer to category) within an inverse_transform of LabelEncoder. In any case, this will eventually be required when presenting the solution to the user.

skopt/parameter.py
MechCoder: The difference is that making it an `abc` will prevent the instance being created if it does not implement those methods, while raising a `NotImplementedError` will allow the instance being created.

(I am not a purist and I don't mind removing them)

MechCoder:
All right, fixed both the inverse_transform and the tests. We can introduce the tuple shorthand in another PR. Please merge if happy

glouppe:
This is great, thank you! Waiting for Travis green light, and then I'll merge.

glouppe:
I'll address the shorthand.

MechCoder:
Yay! We should also refactor the existing minimize code to make use of the new API.

glouppe:
I just made #81 with a few thrown ideas.

betatim:
Whoop!

PR #80, opened 2016-06-10
[MRG] Shorthand for prior of Real variables

glouppe:
This PR adds the shorthand `(lower, upper, prior)` for defining `Real` parameters. Also includes some cosmit changes.

### DIFF
--- a/skopt/parameter.py
+++ b/skopt/parameter.py
@@ -81,12 +81,23 @@ class CategoricalEncoder(object):
 
 
 class Distribution:
-    def rvs(self, n_samples=None, random_state=None):
+    def rvs(self, n_samples=1, random_state=None):
         """
         Randomly sample points from the original space
+
+        Parameters
+        ----------
+        * `n_samples` [int or None]:
+            The number of samples to be drawn.
+
+        * `random_state` [int, RandomState instance, or None (default)]:
+            Set random state to something other than None for reproducible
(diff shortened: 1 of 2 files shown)


skopt/parameter.py
```diff
@@ -100,8 +80,29 @@ def inverse_transform(self, values):
         return self._lb.inverse_transform(values)
 
 
+class Distribution:
+    def rvs(self, n_samples=None, random_state=None):
+        """
+        Randomly sample points from the original space
+        """
+        random_vals = self._rvs.rvs(size=n_samples, random_state=random_state)
+        return self.transformer.inverse_transform(random_vals)
```
MechCoder: I don't know why but does `self.inverse_transform(random_vals)` look better?

skopt/parameter.py
```diff
@@ -100,8 +80,29 @@ def inverse_transform(self, values):
         return self._lb.inverse_transform(values)
 
 
+class Distribution:
+    def rvs(self, n_samples=None, random_state=None):
+        """
+        Randomly sample points from the original space
```
MechCoder: Should we document `n_samples` and `random_state` here?

MechCoder:
Just some minor comments. Otherwise LGTM

glouppe:
Thanks! Comments addressed. Merging.

Issue #81, opened 2016-06-10
Refactor minimize functions to make use of sampling API


glouppe:
Now that #75 has been merged, we should refactor all `*_minimize` functions in order to make use of the new API.

We may need to make a few internal changes since `sample_points` return values in the original space, while we will need to feed the transformed values instead to the optimizer. 

I would expect something along the following lines:

1. Make `_check_grid` a public util returning the corresponding list of Distribution objects.
2. `sample_grid(grid, n_samples)`
3. `warp(grid, samples)`: from original to warped space
4. `unwarp(grid, samples)`: from warped to original sapce

AlexanderFabisch:
The names `warp` and `unwarp` seem to be very generic. I think I would easily forget whether I want to use `warp` or `unwarp`. Maybe we can make it more explicit? Something like `optimizer_space_to_parameter_space`? Maybe we should come up with something shorter. :)

MechCoder:
Maybe just `opt_to_param` and `param_to_opt` ?

MechCoder:
What do we do when we don't want to transform some hyperparameters and don't want to transform others? (For instance, when the model can handle categorical parameters out of the box, for instance in https://github.com/scikit-optimize/scikit-optimize/pull/62). Provide a mask?

glouppe:
Also, not sure we should care about nested grids for now. What do you think?

MechCoder:
Sure. KISS ;)

betatim:
How about `transform` and `inverse_transform` as alternative to `warp`/`unwarp`? Personally I like `warp` because it is short and there doesn't seem to be a good one word name for the parameter-space-used-by-the-optimiser (aka the warped one).

Could we leave it to the user to configure the parameter to use the `Identity` transform in the case where the optimiser natively supports it? An alternative could be that each optimiser provides its own `warp`/`unwarp` which knows to skip certain types?

MechCoder:
+1 for `transform` and `inverse_transform`. I would be in favour of an explicit parameter (a mask) that tells us which parameters to warp and which to leave as is.

Any volunteers? :-)

glouppe:
I'll try to have a go at the refactoring sometime this week.

MechCoder:
Thanks!

glouppe:
Done as part of #83

PR #83, opened 2016-06-14
[MRG] Refactoring to use search space API

glouppe:
This is a work in progress towards #81.

### DIFF
--- a/skopt/dummy_opt.py
+++ b/skopt/dummy_opt.py
@@ -3,10 +3,10 @@ import numpy as np
 from scipy.optimize import OptimizeResult
 from sklearn.utils import check_random_state
 
-from .utils import extract_bounds
+from .space import Space
 
 
-def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
+def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
     """Random search by uniform sampling within the given bounds.
 
     Parameters
@@ -15,9 +15,17 @@ def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
         Function to minimize. Should take a array of parameters and
         return the function values.
 
-    * `bounds` [array-like, shape=(n_parameters, 2)]:
-        - ``bounds[i][0]`` should give the lower bound of each parameter and
-        - ``bounds[i][1]`` should give the upper bound of each parameter.
+    * `dimensions` [list, shape=(n_dims,)]:
+        List of search space dimensions.
+        Each search dimension can be defined either as
+
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - an instance of a `Dimension` object (for `Real`, `Integer` or
+          `Categorical` dimensions), or
+        - as a list of categories (for `Categorical` dimensions).
 
     * `maxiter` [int, default=1000]:
         Number of iterations to find the minimum. In other words, the
@@ -43,16 +51,14 @@ def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
         http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
     """
     rng = check_random_state(random_state)
+    space = Space(dimensions)
+    X = space.rvs(n_samples=maxiter, random_state=rng)
 
-    n_params = len(bounds)
-    lb, ub = extract_bounds(bounds)
-
-    X = lb + (ub - lb) * rng.rand(maxiter, n_params)
     init_y = func(X[0])
     if not np.isscalar(init_y):
         raise ValueError(
             "The function to be optimized should return a scalar")
-    y = np.asarray([init_y] + [func(X[i]) for i in range(maxiter - 1)])
+    y = np.asarray([init_y] + [func(X[1 + i]) for i in range(maxiter - 1)])
 
     res = OptimizeResult()
     best = np.argmin(y)
--- a/skopt/dummy_opt.py
+++ b/skopt/dummy_opt.py
@@ -23,9 +23,9 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
           dimensions),
         - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
           dimensions),
-        - an instance of a `Dimension` object (for `Real`, `Integer` or
-          `Categorical` dimensions), or
-        - as a list of categories (for `Categorical` dimensions).
+        - as a list of categories (for `Categorical` dimensions), or
+        - an instance of a `Dimension` object (`Real`, `Integer` or
+          `Categorical`).
 
     * `maxiter` [int, default=1000]:
         Number of iterations to find the minimum. In other words, the
@@ -58,7 +58,7 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
     if not np.isscalar(init_y):
         raise ValueError(
             "The function to be optimized should return a scalar")
-    y = np.asarray([init_y] + [func(X[1 + i]) for i in range(maxiter - 1)])
+    y = np.asarray([init_y] + [func(X[i]) for i in range(1, maxiter)])
 
     res = OptimizeResult()
     best = np.argmin(y)
--- a/skopt/gbrt_opt.py
+++ b/skopt/gbrt_opt.py
@@ -36,9 +36,9 @@ def gbrt_minimize(func, dimensions, base_estimator=None, maxiter=100,
           dimensions),
         - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
           dimensions),
-        - an instance of a `Dimension` object (for `Real`, `Integer` or
-          `Categorical` dimensions), or
-        - as a list of categories (for `Categorical` dimensions).
+        - as a list of categories (for `Categorical` dimensions), or
+        - an instance of a `Dimension` object (`Real`, `Integer` or
+          `Categorical`).
 
     * `base_estimator` [`GradientBoostingQuantileRegressor`]:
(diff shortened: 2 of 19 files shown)


glouppe:
@MechCoder @betatim I have made the necessary changes to have a full API representing the original and transformed space. Next step is now to change all `*_minimize` functions. I'll start that tonight or tomorrow.

MechCoder:
Thanks! The new interface looks great. I'll give a detailed review tonight.

Also, do you think we should merge https://github.com/scikit-optimize/scikit-optimize/pull/62 and https://github.com/scikit-learn/scikit-learn/issues/6868 (at least port this issue) before refactoring `gp_minimize`?

glouppe:
Too late, I ported `gp_minimize` already, in 246339ce01898c92ce9e143895c5def9c46dcb78.

glouppe:
Bim, job done! Everything should have been updated. Ready for reviews :)

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+        rows = []
+
+        for i in range(n_samples):
+            r = []
+            for j in range(self.n_dims):
+                r.append(columns[j][i])
+
+            rows.append(r)
+
+        return np.asarray(rows)
```
MechCoder: Doing this creates object dtypes. What is your opinion about just returning a list of lists?

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+            The transformed samples.
+        """
+        # Pack by dimension
+        columns = []
+        for dim in self.dimensions:
+            columns.append([])
+
+        for i in range(len(X)):
+            for j in range(self.n_dims):
+                columns[j].append(X[i][j])
```
MechCoder: You can do `columns[j].append(self.dimensions[j].transform(X[i][j]))` here...

MechCoder: Do you think it is worthy to have a `dtype` proprty for each `Distribution` object? If we do that we can simply do:

```python
transformed_columns = []
for i, dim in enumerate(self.dimensions):
    curr_dtype = dim.dtype
    transformed_columns.append(dim.transform(dim.columns[:, i].astype(curr_dtype)).reshape(len(X), -1))

return np.hstack(transformed_columns)
```

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+        rows = []
+
+        for i in range(len(Xt)):
+            r = []
+            for j in range(self.n_dims):
+                r.append(columns[j][i])
+
+            rows.append(r)
+
+        return np.asarray(rows)
```
MechCoder: `np.transpose(columns)` does not work? (Neglecting the fact that it will return an object dtype)

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+
+    @property
+    def n_dims(self):
+        """The dimensionality of the original space."""
+        return sum([dim.size for dim in self.dimensions])
```
MechCoder: This is just `len(self.dimensions)` right?

glouppe: Assuming that dimensions are all of size 1, yes. (I think it is a safe assumption, I'll change that.)

skopt/space.py
glouppe: It should create object dtypes only when there is Categorical variables in the mix right? I like returning a Numpy array, as it often makes it easier then to write the function to optimize. I.e., we would have to rewrite the benchmarks functions so that they accept lists. Maybe we could do that in another PR?

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+
+    @property
+    def transformed_bounds(self):
+        if self.prior == "uniform":
+            return (self._low, self._high)
+
+        elif self.prior == "log-uniform":
+            return (np.log10(self._low), np.log10(self._high))
+
+        else:
```
MechCoder: nit: this will never happen right?

skopt/space.py
```diff
@@ -0,0 +1,458 @@
...
+    def transformed_size(self):
+        return len(self.categories)
+
+    @property
+    def bounds(self):
+        return self.categories
+
+    @property
+    def transformed_bounds(self):
+        return [(0.0, 1.0) for i in range(len(self.categories))]
```
MechCoder: np.tile?

skopt/space.py
MechCoder: Yes, I agree. I looked at this file first and hence assumed that the `_optimize` functions had been refactored to accept categorical parameters.

skopt/space.py
MechCoder: `np.transpose(columns)` should work here as well..

MechCoder:
Just some minor comments, I'll give another pass tomorrow and that should be it.

Sleepy now :yawn:

betatim:
Do we still need [this hack](https://github.com/scikit-optimize/scikit-optimize/pull/83/files#diff-89203a9ab976fbdea06be9034e00772bL30)?

skopt/dummy_opt.py
```diff
@@ -15,9 +15,17 @@ def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
...
-        - ``bounds[i][1]`` should give the upper bound of each parameter.
+    * `dimensions` [list, shape=(n_dims,)]:
+        List of search space dimensions.
+        Each search dimension can be defined either as
+
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - an instance of a `Dimension` object (for `Real`, `Integer` or
```
betatim: I'd swap lines 26 and 28 and change to "an instance of a `Dimension` object (`Real`, `Integer` or `Categorical`)" (remove the for)

skopt/dummy_opt.py
```diff
@@ -43,16 +51,14 @@ def dummy_minimize(func, bounds, maxiter=1000, random_state=None):
...
-    n_params = len(bounds)
-    lb, ub = extract_bounds(bounds)
-
-    X = lb + (ub - lb) * rng.rand(maxiter, n_params)
     init_y = func(X[0])
     if not np.isscalar(init_y):
         raise ValueError(
             "The function to be optimized should return a scalar")
-    y = np.asarray([init_y] + [func(X[i]) for i in range(maxiter - 1)])
+    y = np.asarray([init_y] + [func(X[1 + i]) for i in range(maxiter - 1)])
```
betatim: How about `[func(X[i]) for i in range(1, maxiter)]` to get rid of the +1 -1 business?

skopt/gbrt_opt.py
```diff
@@ -37,9 +28,17 @@ def gbrt_minimize(func, bounds, base_estimator=None, maxiter=100,
...
-        - ``bounds[i][1]`` should give the upper bound of each parameter.
+    * `dimensions` [list, shape=(n_dims,)]:
+        List of search space dimensions.
+        Each search dimension can be defined either as
+
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - an instance of a `Dimension` object (for `Real`, `Integer` or
```
betatim: same as above

skopt/gp_opt.py
```diff
@@ -47,9 +47,17 @@ def gp_minimize(func, bounds, base_estimator=None, acq="LCB", xi=0.01,
...
-        - ``bounds[i][1]`` should give the upper bound of each parameter.
+    * `dimensions` [list, shape=(n_dims,)]:
+        List of search space dimensions.
+        Each search dimension can be defined either as
+
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - an instance of a `Dimension` object (for `Real`, `Integer` or
```
betatim: same as above

skopt/space.py
```diff
@@ -0,0 +1,432 @@
...
+        self._low = low
+        self._high = high
+        self._rvs = randint(self._low, self._high + 1)
+        self.transformer = _Identity()
+
+    def inverse_transform(self, Xt):
+        """Inverse transform samples from the warped space back into the
+           orignal space.
+        """
+        return super(Integer, self).inverse_transform(Xt).astype(np.int)
```
betatim: I can't work out why we need this, can you add a comment that explains it to future us?

skopt/space.py
```diff
@@ -0,0 +1,432 @@
...
+            are equally likely.
+        """
+        self.categories = np.asarray(categories)
+        self.transformer = _CategoricalEncoder()
+        self.transformer.fit(self.categories)
+
+        if prior is None:
+            prior = np.tile(1. / len(self.categories), len(self.categories))
+
+        self._rvs = rv_discrete(values=(range(len(self.categories)), prior))
```
betatim: We should check that `sum(prior)==1`, `rv_discrete` doesn't seem to do that

skopt/space.py
```diff
@@ -0,0 +1,432 @@
...
+            else:
+                columns.append(dim.rvs(n_samples=n_samples, random_state=rng))
+
+        return np.transpose(columns)
+
+    def transform(self, X):
+        """Transform samples from the original space into a warped space.
+
+        Note: this transformation is expected to be used to project samples
+              in a space where numerical optimization is suited.
```
betatim: -> "into a suitable space for numerical optimization."

betatim:
No additional comments.

glouppe:
Thanks both of you for comments! I think I addressed most of them. I'll merge if Travis is happy, so that we can all keep iterating over this.

MechCoder:
@glouppe Thanks! I"ve addressed a minor optimization here (https://github.com/scikit-optimize/scikit-optimize/pull/85)

PR #85, opened 2016-06-17
Optimized transform

MechCoder:
Follow up to https://github.com/scikit-optimize/scikit-optimize/pull/83

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -351,22 +351,12 @@ class Space:
             The transformed samples.
         """
         # Pack by dimension
-        columns = []
-        for dim in self.dimensions:
-            columns.append([])
+        X = np.asarray(X)
+        X_transform = [
+            self.dimensions[j].transform(X[:, j]).reshape(len(X), -1)
+            for j in range(self.n_dims)]
 
-        for i in range(len(X)):
-            for j in range(self.n_dims):
-                columns[j].append(X[i][j])
-
-        # Transform
-        for j in range(self.n_dims):
-            columns[j] = self.dimensions[j].transform(columns[j])
-
-        # Repack as an array
-        Xt = np.hstack([np.asarray(c).reshape((len(X), -1)) for c in columns])
-
-        return Xt
+        return np.hstack(X_transform)
 
     def inverse_transform(self, Xt):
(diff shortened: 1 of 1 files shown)


skopt/space.py
```diff
@@ -351,22 +351,12 @@ def transform(self, X):
             The transformed samples.
         """
         # Pack by dimension
-        columns = []
-        for dim in self.dimensions:
-            columns.append([])
+        X = np.asarray(X)
```
glouppe: Is this always safe? My concern is that this transforms `X` into an array, which comes with casting. Shall we maybe set the dtype to `np.object` to avoid some downcasting effects?

MechCoder:
@glouppe addressed..

skopt/space.py
```diff
@@ -17,10 +17,10 @@ def fit(self, X):
         return self
 
     def transform(self, X):
-        return X
+        return np.asarray(X, dtype=np.float)
 
     def inverse_transform(self, Xt):
-        return Xt
+        return np.asarray(Xt, dtype=np.float)
```
glouppe: Not sure we should inverse transform everything to float by default. Shouldn't we instead do that on a case by case basis, in the `Dimension` sub-classes?

MechCoder: Right, I assumed this was used just in `Real`

MechCoder:
@glouppe ok now?

skopt/space.py
```diff
@@ -351,22 +357,12 @@ def transform(self, X):
...
-
-        # Repack as an array
-        Xt = np.hstack([np.asarray(c).reshape((len(X), -1)) for c in columns])
+        X = np.asarray(X, dtype=object)
+        X_transform = [
+            self.dimensions[j].transform(X[:, j]).reshape(len(X), -1)
+            for j in range(self.n_dims)]
 
-        return Xt
+        return np.hstack(X_transform)
```
glouppe: Then to be sure that everything is under control, shall we cast here? i.e. return `np.hstack(X_transform).astype(np.float)`?

glouppe:
Sorry for the nitpicks... One last comment and feel free to merge.

Issue #86, opened 2016-06-17
Return type of `Space.rvs(x)` and `Space.inverse_transform(X)`


MechCoder:
Both `Space.rvs(X)` and `Space.inverse_transform(X)` return arrays of object dtypes. Are we okay with that?

glouppe:
Indeed, this is not very nice, except when we have dimensions of mixed types. If all are integers or all are reals, then maybe should return an array of ints or an array of floats instead.

betatim:
Started looking into this. It seems the dtype of the return value is determined by the type of the first `Dimension`:

```python
Space([('a', 'b', 'c'), (1, 3)]).rvs(n_samples=3, random_state=0)
# ->
[['b' '2']
 ['c' '2']
 ['b' '3']]
```

betatim:
Actually has nothing to do with the type of the first dimension, but something else because

```
Space([(1, 3), ('a', 'b', 'c')]).rvs(n_samples=3, random_state=0)
# ->
[['1' 'b']
 ['2' 'b']
 ['1' 'b']]
```

glouppe:
This only happens for mixed types. When all dimensions share the same type, it should be consistent. Isnt it?

betatim:
Yes, when everything is of the same type it works as expected. Trying to work out how `np.transpose` chooses the type of the resulting array. I think this is where the "upset" is happening. My guess is it chooses the "most compatible" type somehow

betatim:
For a mix of `int` and `float` you get `float` which also makes sense. The only annoying one (IMHO) is that for anything containing a string you end up with strings everywhere.

glouppe:
Hmm, as part of #112, I minimize over int and real variables. With the current implementation however, `inverse_transform` returns an array of float in that case. This is problematic for hyper-parameter optimisation, since the semantic of the parameters may change depending on whether the values are integer or real (e.g. for `max_features` in trees). 

My opinion is that `inverse_transform` should really return a list of lists of values. This is I think the easiest way to guarantee that values are returned in their correct type.

glouppe:
I'll start working on that, I think the solution should be close to my initial implementation, before we introduce the usage of `np.transpose` as a shortcut.

glouppe:
Closing :beers:

PR #88, opened 2016-07-07
[WIP] Test return type of Space.rvs()

betatim:
Investigating #86 

So far it seems that `np.transpose` is the bit that causes the return value to acquire the type it does. Trying to find docs on how it makes its decision and if we can influence it. However, `np.array`s need to have one shared dtype for all columns :-/

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -338,7 +338,10 @@ class Space:
             else:
                 columns.append(dim.rvs(n_samples=n_samples, random_state=rng))
 
-        return np.transpose(columns)
+        # Build a mixed type array, we have to name fields but can use
+        # an empty string to have names generated for us
+        dtypes = [('', col.dtype) for col in columns]
+        return np.array(list(zip(*columns)), dtype=dtypes)
 
     def transform(self, X):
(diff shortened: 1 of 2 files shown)


betatim:
The original report in #86 suggests that you get a return value of type `object` but I can't reproduce that.

glouppe:
I think  the rule is simple: be as loss-less as possible, while remaining closest from the data. (I.e. int => int, int+float=>float, int+str => object)

Maybe the question is rather, do we want to return an array? (could be instead a list of arrays, each properly typed, or a pandas dataframe, where each series is properly typed, etc)

betatim:
It would be nice to have mixed arrays like this but it seems like it breaks lots of `numpy` stuff elsewhere. 😢

glouppe:
Well, such arrays are flat (but with structured values inside), which breaks the assumption we have that we are dealing with 2d arrays of shape (n_samples, n_dims), hence the failing tests.

MechCoder:
+1 for list of arrays, or even list of lists.

betatim:
See #86 for solution

PR #89, opened 2016-07-09
[MRG] Return variance of decision tree predictions

MechCoder:
Minor modifications to sklearn's tree predict method to return predictions.


MechCoder:
@glouppe @betatim

skopt/learning/tree.py
```diff
@@ -0,0 +1,14 @@
+from sklearn.utils import check_array
+from sklearn.tree import DecisionTreeRegressor as DTRegressor
+
+
+class DecisionTreeRegressor(DTRegressor):
+    """
+    DecisionTreeRegressor that can return the variance of predictions.
+    """
+    def predict(self, X, return_variance=False):
```
betatim: should we `return_std` instead to be consistent with GBRT and GP?

betatim:
Nice! I didn't realise that the impurity in a leaf would be the variance. It is such a simple solution: totally rad bro!

Some thoughts:

* [ ] should we use a different name instead of overloading `DecisionTreeRegressor`? I am +0 on a new name but don't have a good suggestion right now 😦 
* [ ] could you add a test like (or similar, if we make a better one we should backport it to GBRT)
```
rgr = DecisionTreeRegressor(max_depth=3)
rng = np.random.RandomState(1)
N = 10000
X = np.ones((N, 1))
y = rng.normal(size=N)
rgr.fit(X, y)
estimates = rgr.predict(X, return_variance=True)
assert_almost_equal([0, 1], np.mean(estimates, axis=1))
```
* [ ] do we need to check that the impurity criterion is set to MSE?

---

School time: why is the impurity the variance? It must be connected to the splitting criterion being MSE right? If there is an easy way to see this I'd love a short explanation. Or is it as simple as this?

skopt/learning/tree.py
betatim: Moar `__doc__`?

glouppe:
> School time: why is the impurity the variance? It must be connected to the splitting criterion being MSE right? If there is an easy way to see this I'd love a short explanation. Or is it as simple as this?

It matches with the variance because the prediction at a leaf is the mean node value. See eq. 3.22 in https://arxiv.org/pdf/1407.7502.pdf

skopt/learning/tree.py
```diff
@@ -0,0 +1,14 @@
+from sklearn.utils import check_array
+from sklearn.tree import DecisionTreeRegressor as DTRegressor
```
glouppe: `DTRegressor` -> `sklearn_DecisionTreeRegressor`? (to emphasize we are just defining a thin layer over it)

skopt/learning/tree.py
glouppe: +1 for `return_std` instead

glouppe:
> should we use a different name instead of overloading DecisionTreeRegressor? I am +0 on a new name but don't have a good suggestion right now

Not sure this is necessary. I am fine with reusing the same name, as we are anyway overloading sklearn's tree with a drop-in replacement that should be compatible wherever sklearn's is.

glouppe:
> do we need to check that the impurity criterion is set to MSE? (MSE is the default and only choice right now)

+1, better be safe

glouppe:
Thanks Manoj! Besides the small comments, +1 for this.

betatim:
👍 from me too

MechCoder:
@betatim It is good to note that the mean square is the same as the variance in predictions since the predicted value is the mean. Thanks for that! (Not that it matters but I did not see that the split criterion is set to "mse" and not "impurity").

MechCoder:
I'm going to merge when Travis passes. Feel free to address any nitpicks in follow-up PR's. Thanks!

betatim:
Nice work 🎆

PR #90, opened 2016-07-12
[MRG] Added `return_std` support for forest regressors

glouppe:
This adds `return_std` support for `RandomForestRegressor` and `ExtraTreesRegressor`, such that `std[i] ~= std(y | X[i])`.

This implements `std_v3` as shown in https://github.com/glouppe/notebooks/blob/master/Variance%20from%20tree-based%20models.ipynb


glouppe:
Travis is happy, but are you? :) @betatim @MechCoder

skopt/__init__.py
```diff
@@ -1,6 +1,6 @@
 """
 Scikit-Optimize, or `skopt`, is meant to be a simple and efficient library
```
betatim: "is mean to be a simple" -> "is a simple"

glouppe: Fixed

betatim:
Pondering if we should merge the common tests or add the GBRT to the tests in `tests/test_forest.py`?

Why turn of bootstrapping? 😕

glouppe:
> Why turn of bootstrapping? :confused:

Because this may yield to trees that are different to what is expected and hardcoded in those tests :)

betatim:
LGTM

glouppe:
> Pondering if we should merge the common tests or add the GBRT to the tests in tests/test_forest.py?

Yes in general, the test suite could be given some love :)

betatim:
I'll do some test suite loving.

MechCoder:
That was quick. I'll have a look tonight.

betatim:
@glouppe waits for no one 🚄

MechCoder:
Thanks for this!

By the way, for the default settings for the RandomForest, i.e when each tree is grown to full depth, the variance in prediction of each sample becomes zero and thus the equation reduces to (\mu_b)**2 / B - (\mu)**2 . What are your thoughts on this?

skopt/learning/forest.py
```diff
@@ -0,0 +1,105 @@
...
+                raise ValueError(
+                    "Expected impurity to be 'mse', got %s instead"
+                    % self.criterion)
+
+            # This derives std(y | x) as described in 4.3.2 of arXiv:1211.0906
+            std = np.zeros(len(X))
+
+            for tree in self.estimators_:
+                var_tree = tree.tree_.impurity[tree.apply(X)]
+                mean_tree = tree.predict(X)
```
MechCoder: How I wish we could `return_std=True` directly here.

glouppe:
> By the way, for the default settings for the RandomForest, i.e when each tree is grown to full depth, the variance in prediction of each sample in each tree becomes zero and thus the equation reduces to (\mu_b)^2 / B - (\mu)^2 . What are your thoughts on this?

Yes, this is correct. The assumption is that leaves have enough density to derive meaningful statistics. By default, we should therefore combine forests with something like `min_samples_leaf=10`. (In SMAC they do something similar and construct the tree such that impurity>0.01 for all leaves.)

MechCoder:
So, a RF based approach would make sense only when there are a huge number of hyperparameters to train. For all other cases, a GP based approach would suffice.

MechCoder:
Oh sorry, I typed too quickly, here the feature space is the hyperparameter space!! It should be interesting to benchmark after https://github.com/scikit-optimize/scikit-optimize/pull/91/ is merged

PR #91, opened 2016-07-12
Optimisation with tree based models

betatim:
Generalised `gbrt_minimize` to work with all the tree based models added in #89 

Need to tune the limits/hyperparameters a bit to make `test_tree_based_minimize` converge.

Is there a better way to have each of the models have its own `XXX_minimize()` and not duplicate the doc string?


glouppe:
One thing I realize: `n_points` should be set to a much larger value since querying the tree-model should be super fast (it is not the case for `gp_optimize` though). I would set a default value around `n_points=1000` instead.

glouppe:
+1 for merging all these methods into one. Say `forest_minimize`? Then add a `base_estimator` parameter that can either be a string (`"gbrt"`, `"rf"`, `"et"`, "dt"` which initializes the corresponding estimator with good defaults) or an estimator supporting `return_std`.

glouppe:
Maybe `gbrt_minimize` could still live on its own though, since it is a bit special and based on boosted quantile regression.

glouppe:
Maybe fuse rf+et+dt into a single `forest_minimize` function? with an optional `base_estimator` argument, taking either a string or an estimator object.

betatim:
With some tuning (one iteration 😃) you can get most to work except for the RF one.

Also increased `n_points` but I am not sure how much that helps. My feeling is that if things get stuck it isn't the optimisation of the acquisition function but the model itself. (We need some better plotting/debug tools/diagnosis plots.)

+1 on merging ET, DT and RF to one.

MechCoder:
If I understand the paper correctly, they combine the bootstrap effects of RandomForest along with the uniform split criterion feature of ERT's

skopt/tree_opt.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    rng = check_random_state(random_state)
+
+    if isinstance(base_estimator, str):
+        if base_estimator not in ("rf", "et", "dt"):
+            raise ValueError("Valid values for base_estimator parameter"
+                             " are: 'rf', 'et' or 'dt', not '%s'" % base_estimator)
+
+        if base_estimator == "rf":
+            base_estimator = RandomForestRegressor(min_samples_leaf=10,
```
MechCoder: Would we want to set `min_samples_leaf=5` and `max_features=10` just because the paper says so? (or at least for the initial benchmarks?)

glouppe:
> If I understand the paper correctly, they combine the bootstrap effects of RandomForest along with the uniform split criterion feature of ERT's

They are not exactly using `ExtraTreesRegressor(bootstrap=True)`. Rather they build standard decision trees as in random forests, but instead of using the mid-point between consecutive feature values as the cut-off, they draw it uniformly at random within that range. In ERT, cut-offs are not evaluated at each pair of consecutive feature values, but rather directly drawn at uniformly between the min and max feature values.

skopt/tree_opt.py
betatim: I set this to ten because it makes the test pass and to have a number bigger than one. No particular opinion on what to set it to beyond setting as few parameters by hand as possible.

betatim:
All forest related methods now have `min_samples_leaf=10` and `gbrt_minimize` uses less `n_estimators`. Main motivation for that is it speeds up the test.

I also changed `check_minimize` to be "best of N" to determine if it passes. After testing different hyper-parameter settings and seeds the impression I get is that we are basically being lucky and having a random sample hit the minimum. Or maybe the problems are not hard enough.

Either way I think as the minimisation process contains some randomness we should run it a "few" times and check that "on average" it converges. A single instance means nothing, you are jsut being lucky 🎲

skopt/tree_opt.py
```diff
@@ -0,0 +1,255 @@
...
+        - `func_vals` [array]: function value for each iteration.
+
+        For more details related to the OptimizeResult object, refer
+        http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
+    """
+    rng = check_random_state(random_state)
+
+    # Default estimator
+    if base_estimator is None:
+        gbrt = GradientBoostingRegressor(n_estimators=20, loss='quantile')
```
betatim: Should this be set here or moved to `GradientBoostingQuantileRegressor`? I can't make up my mind, please help :)

skopt/tree_opt.py
```diff
@@ -0,0 +1,255 @@
...
+        elif base_estimator == "et":
+            base_estimator = ExtraTreesRegressor(min_samples_leaf=10,
+                                                 random_state=rng)
+
+        elif base_estimator == "dt":
+            base_estimator = DecisionTreeRegressor(min_samples_leaf=10,
+                                                   random_state=rng)
+
+    else:
+        if not isinstance(base_estimator, RegressorMixin):
```
betatim: Should we get all technical and introspect the `base_estimator`'s `predict()` to see if it has an argument `return_std` or overkill? I'm thinking overkill.

MechCoder:
@glouppe Thanks for the clarification! I was under the wrong impression that feature values themselves are chosen as cut-offs in decision trees.

MechCoder:
@glouppe Thanks for the clarification! I was under the wrong impression that feature values themselves are chosen as cut-offs in decision trees.

skopt/tree_opt.py
glouppe: Nope it is good here. (The rationale being that the best default values for model-based optimisation are not necessarily the same as the best default values for all supervised learning problems.)

skopt/tree_opt.py
glouppe: Overkill. Let users be grown-ups and document that the base estimator is expected to support that :)

glouppe: Also, the proper way of checking that the object is a regressor is to verify that the attribute `base_estimator._estimator_type` exists and that its value is `"regressor"`.

betatim:
Travis is finally happy 🎉, merge?

glouppe:
Bim! :beers:

betatim:
🍰 !

MechCoder:
Awesome! Too bad I can't have a look before the PR gets merged. (Time zones :/)

betatim:
Maybe we should leave big PRs like this hanging in "to be merged" state a bit longer so everyone can get a chance to look? Or we do "post publication review" -> comment even after merge and we address it in a new PR?

MechCoder:
Waiting for a day or a couple of days would be ideal from now onwards, since we have got the basic framework out.

skopt/tests/test_tree_opt.py
```diff
@@ -0,0 +1,102 @@
...
+                         forest_minimize, lambda x: 0., [],
+                         base_estimator=42)
+
+    assert_raise_message(ValueError,
+                         "The base_estimator parameter has to either",
+                         forest_minimize, lambda x: 0., [],
+                         base_estimator=DecisionTreeClassifier())
+
+
+def check_minimize(minimizer, func, y_opt, dimensions, margin, maxiter):
```
MechCoder: It would be helpful to add a docstring to the future reader so that he can understand why this is being done N number of times instead of just 1.

skopt/tests/test_tree_opt.py
```diff
@@ -0,0 +1,102 @@
...
+    success = 0
+    N = 3
```
MechCoder: Should we increase N to a higher number? I understand that the drawback is that this increases the test time, but then it is better since it can still reach the minimum 2 out of 3 time by chance.

skopt/tests/test_tree_opt.py
```diff
@@ -0,0 +1,102 @@
...
+
+def test_no_iterations():
+    yield (check_no_iterations, gbrt_minimize)
+    yield (check_no_iterations, partial(forest_minimize, base_estimator='et'))
+    yield (check_no_iterations, partial(forest_minimize, base_estimator='dt'))
+    yield (check_no_iterations, forest_minimize)
+
+
+def test_one_iteration():
+    result = gbrt_minimize(branin, [(-5.0, 10.0), (0.0, 15.0)],
```
MechCoder: Should we not raise an error asking the user to either set `maxiter >= n_start`?

IMO, we should not implicitly change the default value of `n_start=10` or in other words the user should know what is happening.

skopt/tests/test_tree_opt.py
```diff
@@ -0,0 +1,102 @@
...
+                         random_state=1)
+
+    assert_raise_message(ValueError, "at least one starting point",
+                         minimizer,
+                         branin, [(-5.0, 10.0), (0.0, 15.0)], n_start=0,
+                         maxiter=2, random_state=1)
+
+
+def test_no_iterations():
+    yield (check_no_iterations, gbrt_minimize)
```
MechCoder: You can define a list of estimators as done in `test_tree_based_minimize` on the top of the file and avoid code duplication here.

skopt/tree_opt.py
```diff
@@ -0,0 +1,256 @@
...
+        elif base_estimator == "et":
+            base_estimator = ExtraTreesRegressor(min_samples_leaf=10,
+                                                 random_state=rng)
+
+        elif base_estimator == "dt":
+            base_estimator = DecisionTreeRegressor(min_samples_leaf=10,
+                                                   random_state=rng)
+
+    else:
+        if not (hasattr(base_estimator, '_estimator_type') and
```
MechCoder: Just curious to know, why this is better than just `isinstance(RegressorMixin)`. Both are ducktyping right?

skopt/tests/test_tree_opt.py
```diff
@@ -0,0 +1,102 @@
...
+
+    assert_equal(len(result.models), 0)
+    assert_array_equal(result.x_iters.shape, (1, 2))
+    assert_array_equal(result.func_vals.shape, (1,))
+    assert_array_equal(result.x, result.x_iters[np.argmin(result.func_vals)])
+    assert_almost_equal(result.fun, branin(result.x))
+
+
+def test_seven_iterations():
+    result = gbrt_minimize(branin, [(-5.0, 10.0), (0.0, 15.0)],
```
MechCoder: This should be tested for all optimizers.

skopt/tree_opt.py
glouppe: `isinstance(RegressorMixin)` does not allow ducktyping, since it requires the object to inherit from the mixin. The official way, as we describe it in scikit-learn, is to check for that attribute.

MechCoder: OK. using `sklearn.base.is_regressor` might be cleaner (which does the same thing).

MechCoder:
@betatim Just some minor comments related to testing. Thanks!

betatim:
Addressed all your comments in #98 

Outstanding is whether or not to increase the number of iterations for the convergence test. Currently it takes about 60s to run that test on my laptop, so I would not increase it above 5. Otherwise I predict people will start disabling that test when they run locally 🙉

Issue #23, opened 2016-03-28
Definition of terms related to uncertainty


betatim:
Some thoughts on "uncertainty". This issue was inspired by @MechCoder's comment in #9. The first part of this issue tries to correctly define various terms that often get used interchangeably and are easy to confuse (I confidently predict that I will make at least one error in this post). Once we have defined the terms, we can decide which of them we need in order to evaluate various acquisition functions.

[Standard deviation](http://mathworld.wolfram.com/StandardDeviation.html) (\sigma): this is the square root of the variance. Can be calculated for any sample no matter what distribution the samples come from.

[Standard error](http://mathworld.wolfram.com/StandardError.html) (of the mean): `\sigma / \sqrt(N)` a measure of the uncertainty associated with the estimated value of the mean.

[Confidence interval](http://mathworld.wolfram.com/ConfidenceInterval.html) (CI): The N% confidence interval will contain the measured value N% of the time. Alice wants to estimate the value of a parameter `t`, so she constructs an estimator `that` as well as a CI. The 68% CI (around `that`) will contain the true value `t` in 68% of experiments (that is we clone Alice and repeat what she did many times).

N% quantile: The N% quantile starts at negative infinity and goes until a point `x`, think of it as the integral of the p.d.f. between `-inf` and `x` which equals N%.

If `that` is distributed according to a normal distribution then the 68% CI is [`that` - `sigma`, `that` + `sigma`].

For a normal distribution `mu`-`sigma` = the 16% quantile.

For our purposes we have a surrogate model (a GP or what have you) for the true, expensive function `f`. At a given point `x` our best estimate of the true value of `f` is the mean `mu(x)` of our surrogate model.

-------------

Now my understanding runs out -> need help.

What is the band we get from a GP and then feed into EI and friends? Is it the "standard error on the mean" or "68% confidence interval" or "68% [credible interval](https://en.wikipedia.org/wiki/Credible_interval)" or something else?

glouppe:
> What is the band we get from a GP and then feed into EI and friends?

A GP defines a probability over functions that could have generated the observed data (x_i, y_i). At a fixed x, you therefore also get a conditional distribution for y|x (by integrating over all possible functions and evaluating their value at x). This conditional distribution is a model of the true distribution of y|x, if you could observe infinitely many values y for the same x (e.g. because the relation between x and y involves some random fluctuations). The band you get from a GP corresponds to the standard deviation of that  conditional distribution, under the model assumed by the GP. 

[...]

PR #93, opened 2016-07-13
[MRG] Refactor standard deviation computation into _return_std helper

MechCoder:


MechCoder:
@betatim @glouppe

skopt/learning/forest.py
```diff
@@ -3,6 +3,44 @@
 from sklearn.ensemble import ExtraTreesRegressor as sk_ExtraTreesRegressor
 
 
+def _return_std(X, trees, predictions):
+    """
+    Return standard deviation in predictions given a list of trees.
```
glouppe: Looks good, except for this docstring. We are computing `std(y|x)`, not `std(\hat(y)|x)`.

MechCoder: By given a list of trees, I had meant `given a list of trees as a parameter in the programming sense` and not `std given a tree in the probabilistic sense` . My bad, that it implied that :P

MechCoder:
will merge when travis passes

skopt/learning/forest.py
glouppe: I was rather saying that we are computing the standard deviation of the target `y`, rather than the standard deviation of the prediction `\hat(y)` of the target. For me "standard deviation in predictions"  meant the later.

skopt/learning/forest.py
```diff
@@ -3,6 +3,47 @@
 from sklearn.ensemble import ExtraTreesRegressor as sk_ExtraTreesRegressor
 
 
+def _return_std(X, trees, predictions):
+    """
+    Returns ``std(Y | X)``.
```
betatim: only one ` needed

MechCoder:
@glouppe I admit, I don't completely understand the argument.

The method computes the variance in predictions of each random forest, while giving an estimate of the variance in the target. In other words, we are assuming `std(Y_hat | x)` as an estimate of `std(Y | x)`. The method still computes `std(Y_hat | x) right?`

skopt/learning/forest.py
MechCoder: Great, I'll push this directly.

glouppe:
> In other words, we are assuming std(Y_hat | x) as an estimate of std(Y | x). The method still computes std(Y_hat | x) right?

No, we are not. Y_hat is itself a random variable, with a corresponding mean and variance, and is distinct from Y. For this precise reason, this is why we dont use `np.std([tree.predict(X) for tree in forest], axis=0)` as an estimate of `std(Y|X)`. This would indeed rather compute the standard deviation `std(Y_hat|X)` of the prediction Y_hat,  over the distribution of random trees.

MechCoder:
Just to make sure we are on the same page, in your explanation `Y_hat` is a random variable that denotes the prediction in each tree?

glouppe:
Yes, Y_hat(_tree) is the prediction of my model (here an individual tree). Note that Y_hat_forest would itself be another random variable, which is distinct from Y_hat(_tree).

MechCoder:
Yes, by "variance in predictions", I was meaning the "variance in predictions of the random forest model" not `Var(Y_hat | X)`. More precisely, the one described as `Var(L)` described in page 17 of the paper. Does it make sense then?

glouppe:
Well, for the same reasons, I understand "variance in predictions of the random forest model"  as `Var(Y_hat_forest | X)`, which for me is distinct from `Var(Y | X)`.

MechCoder:
Ok, it is just a bit confusing to me to say that it "computes" `Var(Y | X)` when it estimates `Var(Y | X)`. The same way that you cannot compute the true mean given data sampled from a distribution (unless you know the distribution) but you can just estimate the mean.

glouppe:
Yes sure, we only approximates it, but still we need to approximate the right quantity.

betatim:
Worth adding some of these definitions/clarifications/shared vocabulary to #23 (and eventually build that issue into a section of the docs)? I can see this issue coming up again and again.

MechCoder:
It is due to my misunderstanding.

For me, by "variance in predictions", I was referring to the variance in the leaf node for each tree but the correct terminology is as described by @glouppe over here (https://github.com/scikit-optimize/scikit-optimize/pull/93#issuecomment-232270233). That is why random forests are used, to reduce the "variance in predictions" :-)

MechCoder:
And I do think, they should be added (preferably by me)

PR #94, opened 2016-07-13
Round off variance in each tree to 0.01 in accordance with SMAC

MechCoder:
Described in section 4.3.2 of http://arxiv.org/pdf/1211.0906v2.pdf (`\sigma^2_min` = 0.01)

"To avoid making deterministic predictions for leaves with few data points, we round the stored variance up to `sigma^2_min`=0.01

### DIFF
--- a/skopt/learning/forest.py
+++ b/skopt/learning/forest.py
@@ -34,6 +34,13 @@ def _return_std(X, trees, predictions):
 
     for tree in trees:
         var_tree = tree.tree_.impurity[tree.apply(X)]
+
+        # This rounding off is done in accordance with the
+        # adjustment done in section 4.3.3
+        # of http://arxiv.org/pdf/1211.0906v2.pdf to account
+        # for cases such as leaves with 1 sample in which there
+        # is zero variance.
+        var_tree[var_tree < 0.01] = 0.01
         mean_tree = tree.predict(X)
         std += var_tree + mean_tree ** 2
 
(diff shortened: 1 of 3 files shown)


MechCoder:
@betatim @glouppe

MechCoder:
I was under the impression that they stopped training each tree when impurity<0.01 but the term "round off the variance" seems ambiguous to me.

skopt/learning/forest.py
```diff
@@ -34,6 +34,13 @@ def _return_std(X, trees, predictions):
 
     for tree in trees:
         var_tree = tree.tree_.impurity[tree.apply(X)]
+
+        # This rounding off is done in accordance with the
+        # adjustment done in section 4.3.3
+        # of http://arxiv.org/pdf/1211.0906v2.pdf to account
+        # for cases such as leaves with 1 sample in which there
+        # is zero variance.
+        var_tree[var_tree < 0.01] = 0.01
```
glouppe: Can we instead define this threshold as an optional argument of `_return_std`, e.g. `min_variance=0.01`.

glouppe:
Can we instead define this threshold as an optional argument with `0.0` as default value, e.g `min_variance=0.0`, and then change it as appropriately in the forest-based minimizer?  I am quite curious about the necessity of this hack in practice.

betatim:
+1 for making this optional. We should understand each of the tweaks/hacks/seemingly arbitrary choices made in SMAC instead of implementing them just because they are in SMAC. Looked at it a different way: we know enough about this to attempt to build a solution that is _better_ than SMAC 😉

MechCoder:
@betatim @glouppe I agree as well. I made it a constructor parameters, so that it can be tuned etc. and added a minor test. Please merge. if happy.

glouppe:
Merging, thanks!

MechCoder:
I pushed a stronger test to master directly after it passed locally (https://github.com/MechCoder/scikit-optimize/commit/7795bb8c78a10e2a52ca5ccc222d3f34ecb7bfed), hopefully that is not an issue.

PR #95, opened 2016-07-13
Squashing warnings

betatim:
Squashing some warnings. I added

```python
import traceback
import warnings
import sys

def warn_with_traceback(message, category, filename, lineno, file=None, line=None):
    traceback.print_stack()
    log = file if hasattr(file,'write') else sys.stderr
    log.write(warnings.formatwarning(message, category, filename, lineno, line))
    raise RuntimeError

warnings.showwarning = warn_with_traceback
```

to `skopt/__init__.py` and then ran `nosetest --pdb`. Is there a way to
get the same effect via env variable or commandline argument for nose?


glouppe:
+1 for the changes, not sure about your question though

PR #98, opened 2016-07-14
Improved tests for forest_minimize

betatim:
This is a follow up to #91.

----

What would have happened if I had pushed these changes to the same branch as #91 used?


MechCoder:
lgtm pending travis

MechCoder:
I think pushing to the same branch and creating a new pull request comparing that branch and master might work. (but not completely sure)

glouppe:
Looks good to me too, merging.

PR #99, opened 2016-07-14
[MRG] fix support for categorical variables

### LINKED ISSUE #97: Support for categorical parameters seems to be broken
```python
def bench1(x):
    return np.asscalar(np.asarray(x))

def bench2(x):
    return np.asscalar(np.asarray(x, dtype=np.int))

bench1([1])
1
bench2(["1"])
1

# Works
forest_minimize(bench1, ((1.0, 4.0),))
# Fails
forest_minimize(bench2, (("1", "2", "3", "4"),))

# Works
gp_minimize(bench1, ((1.0, 4.0),), maxiter=5)
# Fails
gp_minimize(bench2, (("1", "2", "3", "4"),))
```

glouppe:
This fixes #97, modulo the fact that the function to minimize actually returns a float.

The main issue was that `Xi = np.zeros((maxiter, space.n_dims))` was forcing entries to be cast to floats, no matter the type of the dimensions. Samples are now stored in a list instead, as was done in `gp_minimize`.

### DIFF
--- a/skopt/tree_opt.py
+++ b/skopt/tree_opt.py
@@ -53,7 +53,7 @@ def _tree_minimize(func, dimensions, base_estimator, maxiter,
         # for the moment.
         X = space.transform(space.rvs(n_samples=n_points,
                                       random_state=rng))
-        values = gaussian_ei(X, rgr, np.min(yi))
+        values = -gaussian_ei(X, rgr, np.min(yi))
         next_x = X[np.argmin(values)]
 
         next_x = space.inverse_transform(next_x.reshape((1, -1)))[0]
--- a/skopt/tree_opt.py
+++ b/skopt/tree_opt.py
@@ -6,7 +6,6 @@ from scipy.optimize import OptimizeResult
 
 from sklearn.base import clone
 from sklearn.base import is_regressor
-from sklearn.base import RegressorMixin
 from sklearn.ensemble import GradientBoostingRegressor
 from sklearn.utils import check_random_state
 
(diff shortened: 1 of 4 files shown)


betatim:
Merge commits, woah! Can I be pedantic: rebase please 🙏

glouppe:
F*, i messed up the history now...

glouppe:
> Merge commits, woah! Can I be pedantic: rebase please :pray:

Done

glouppe:
Hmm some tests are failing, but this has nothing to do with what I changed... ?!

betatim:
Ergh, for some of the surrogate models some of the convergence tests fail. Have you tried this branch before and after applying 57025cc (or does it not happen at all locally?)

Is the right conclusion from this that somewhere we are getting non reproducible random numbers? As in we changed some code and now get a different sequence?

skopt/tree_opt.py
```diff
@@ -39,38 +33,39 @@ def _tree_minimize(func, dimensions, base_estimator, maxiter,
         raise ValueError("Total number of iterations set by maxiter has to"
                          " be larger or equal to n_start.")
 
-    Xi[:n_start] = space.rvs(n_samples=n_start, random_state=rng)
-    yi[:n_start] = [func(xi) for xi in Xi[:n_start]]
-    i = np.argmin(yi[:n_start])
-    best_y = yi[i]
-    best_x = Xi[i]
+    Xi = space.rvs(n_samples=n_start, random_state=rng)
+    yi = [func(x) for x in Xi]
```
MechCoder: nitpick: We can still preallocate `y`

MechCoder:
> some tests are failing, but this has nothing to do with what I changed

It has "almost" nothing to do with what you changed :-D . You need to revert `next_x = X[np.argmin(values)]` to `next_x = X[np.argmax(values)]` as before. We would like to maximize the expected improvement, which minimizes the objective.

MechCoder:
Thanks for addressing this, +1 pending comments (and hopefully Travis)

PR #100, opened 2016-07-14
Deploy notebooks with the static documentation

glouppe:
I am merging this right now, as I need to check whether Travis deploys correctly. We will improve things afterwards.


betatim:
💯

glouppe:
Travis seems to have internal issues, jobs are queuing...

glouppe:
https://scikit-optimize.github.io/notebooks/Bayesian%20optimisation.html !

The architecture is properly setup, now the next thing is to write a nice notebook describing Bayesian optimisation and our API :)

glouppe:
Note that the notebook I shipped was only for test purposes. It is mostly copy/pasted from one of our examples. It should really be improved with narrative documentation, better figures, etc.

betatim:
Is there a way to specify/automagic a nicer URL? Or we just agree to use "nice" file names for the notebooks so the URLs are nice too.

This is super cool. /me dreams of interactive docs ... one day :)

glouppe:
A better solution could certainly be hacked than what I do for the moment. E.g., by renaming the notebook files once they have been rendered (hence using the original nice name for the link text, and the renamed version for the url).

MechCoder:
This looks great!

glouppe:
Just pushed a1fbee21fcd1d4b1eeb390e1afe4657ff878a70f

URLs should now be better (once Travis has finished redeploying the doc)

glouppe:
https://scikit-optimize.github.io/notebooks/bayesian-optimisation.html  does this url look better? :)

MechCoder:
great, thanks!

betatim:
Oh yeah!

PR #109, opened 2016-07-16
[WIP] Add smarter sampling to optimize the acquisition function

MechCoder:
Add smarter sampling based on the previous "best" points to `tree_opt`

The idea is given a point of dimension `n_params`, the next candidates are got by a one-exchange neighborhood.

1. They differ in exactly one parameter.
2. The neighbours of the n_th param are got by a univariate gaussian with the point[n_param] as mean and a configurable std.


MechCoder:
I added a test in `test_paramils.py` that seems to work. I still have to figure out why the tests in `test_tree_opt.py` are failing. Also there are numerous optimizations to be made, hence I would prefer feedback on correctness rather than speed (for now).

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,92 @@
...
+    actual_x_min = _transform_back(
+        transformed_x_min, lower_bounds, upper_bounds)
+
+    fun_min = func(actual_x_min.reshape(1, -1))[0]
+    n_params = len(x0)
+
+    while True:
+        # TODO: Cache and modify in-place.
+        transformed_points = np.tile(
+            transformed_x_min, (20 * n_params, 1))
```
MechCoder: SMAC says 4 neighbors but I added 20 here. We should configure this more properly, since prediction is typically free (except for GP's)

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,92 @@
...
+        if curr_min < fun_min:
+            transformed_x_min = transformed_points[min_func_ind]
+            actual_x_min = _transform_back(
+                transformed_x_min, lower_bounds, upper_bounds)
+            fun_min = curr_min
+        else:
+            break
+    return actual_x_min
+
+def paramils_smac(func, x0, bounds, n_neighbors=4, std=0.6,
```
MechCoder: SMAC says `std=0.2` but I made it 0.6 for the tests to pass. I am open to suggestions.

skopt/tree_opt.py
```diff
@@ -51,11 +54,27 @@ def _tree_minimize(func, dimensions, base_estimator, maxiter,
...
-
+        curr_min = np.min(yi)
+        if acq_min == "sampling":
+            X = space.transform(space.rvs(n_samples=n_points,
+                                          random_state=rng))
+            values = -gaussian_ei(X, rgr, curr_min)
+            next_x = X[np.argmin(values)]
+        elif acq_min == "paramils":
+            for dimension in space.dimensions:
+                if isinstance(dimension, Categorical):
```
MechCoder: It is not straightforward to support the categorical case, without changing the space API to optionally not transform some of the dimensions. I would prefer handling this later.

MechCoder:
I believe the idea of using the univariate gaussian ties with the exploration / exploitation idea. Smartly sample points closer to the previous minimum while not ignoring the points far away.

MechCoder:
The tests in `local_solve/test_paramils.py` convince me. Do they convince you?
@betatim @glouppe :-)

MechCoder:
Compare the number of iterations in `test_slow_minimize` and `test_tree_based_minimize`. Looks like we are getting somewhere ;)

MechCoder:
Also see: https://github.com/scikit-optimize/scikit-optimize/pull/110

glouppe:
I'll have a deeper look later today. However I was wondering, in SMAC, the local search is combined with random sampling of 10000 points. Do you also want to implement this? 

Or more generally, do we want that `forest_minimize` to exactly implement SMAC? (I would prefer *not*, as I am not convinced everything is optimal there, but we could instead of a `smac_minimize` or something)

betatim:
> Or more generally, do we want forest_minimize to exactly implement SMAC? (I would prefer not,
> as I am not convinced everything is optimal there, but we could instead have a smac_minimize 
> implementation or something)

+1 It seems like there are a lot of tradeoffs that one could make and explore. -> let's try and make `tree_minimize` "general" and have specialisations like `smac_minimize` that follow SMAC as closely as possible.

I would like to look at this PR in depth and comment but won't have time to do so till Tuesday. If that is too slow then go ahead without me and I will comment post-merge (and we iterate).

skopt/local_solve/__init__.py
```diff
@@ -0,0 +1 @@
+from .paramils import paramils_smac
```
glouppe: `local_solve` -> `solvers`? I guess this will not be the only one we write.

skopt/tree_opt.py
glouppe: Not sure the API needs to be changed for that, but okay for dealing with that in another PR.

skopt/tree_opt.py
```diff
@@ -10,15 +10,18 @@
...
 from .learning import GradientBoostingQuantileRegressor
 from .learning import RandomForestRegressor
+from .local_solve import paramils_smac
 from .space import Space
 
 
 def _tree_minimize(func, dimensions, base_estimator, maxiter,
-                   n_points, n_start, random_state=None):
+                   n_points, n_start, random_state=None,
+                   acq_min="sampling"):
```
glouppe: Let's be explicit and name this `acq_minimizer` instead.

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+        curr_min = func_values[min_func_ind]
+        if curr_min < fun_min:
+            transformed_x_min = transformed_points[min_func_ind]
+            actual_x_min = actual_points[min_func_ind]
+            fun_min = curr_min
+        else:
+            break
+    return actual_x_min
+
+def paramils_smac(func, x0, bounds, n_neighbors=20, std=0.6,
```
glouppe: the doc says that `n_neighbors=4`?

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+    return x0 * (upper_bounds - lower_bounds) + lower_bounds
+
+def _local_search(func, x0, bounds, std, n_neighbors,
+                  random_state=None, maxiter=10):
+    rng = check_random_state(random_state)
+    upper_bounds = bounds[:, 1]
+    lower_bounds = bounds[:, 0]
+
+    transformed_x_min = np.asarray(x0)
+    actual_x_min = _transform_back(
```
glouppe: Why not transforming back in `paramils_smac` instead?

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+    fun_min = func(actual_x_min.reshape(1, -1))[0]
+    n_params = len(x0)
+
+    for i in range(maxiter):
+        # TODO: Cache and modify in-place.
+        transformed_points = np.tile(
+            transformed_x_min, (n_neighbors * n_params, 1))
+
+        for n_param in range(n_params):
+            curr_mean = transformed_x_min[n_param]
```
glouppe: `curr_mean` is not a good name. The mean of what are you referring to?

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+
+    fun_min = func(actual_x_min.reshape(1, -1))[0]
+    n_params = len(x0)
+
+    for i in range(maxiter):
+        # TODO: Cache and modify in-place.
+        transformed_points = np.tile(
+            transformed_x_min, (n_neighbors * n_params, 1))
+
+        for n_param in range(n_params):
```
glouppe: `n_param` is not a good name either. I would assume it corresponds to the number of parameters, but it is not, it is the index of a parameter. 

Also `parameter` is confusing. Are you referring to a sample (== a configuration of the "parameters") or to a dimension/feature (associated to a parameter)?

glouppe: We should really try to be consistent across the whole package about how we name things...

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+            start_ptr = n_param * n_neighbors
+            end_ptr = start_ptr + n_neighbors
+            transformed_points[start_ptr: end_ptr, n_param] = random_points
+
+        actual_points = _transform_back(
+            transformed_points, lower_bounds, upper_bounds)
+        func_values = func(actual_points)
+        min_func_ind = np.argmin(func_values)
+        curr_min = func_values[min_func_ind]
+        if curr_min < fun_min:
```
glouppe: Sorry for being pedantic, but please try to chose better variable snames. E.g., here we `min_func_ind` which is associated to `curr_min` and not to `fun_min`. This is is really confusing.

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+    # Also prevents useless re-checks when this function is called
+    # in the *_minimize functions.
+    bounds = np.asarray(bounds)
+    upper_bounds = bounds[:, 1]
+    lower_bounds = bounds[:, 0]
+    x0 = (x0 - lower_bounds) / (upper_bounds - lower_bounds)
+
+    # TODO: Use n_jobs
+    # Faster to iterate over lists over numpy arrays.
+    x0_list = x0.tolist()
```
glouppe: Not sure these micro-optimizations are necessary.

skopt/local_solve/paramils.py
```diff
@@ -0,0 +1,109 @@
...
+            random_points = np.clip(
+                curr_mean + std * rng.randn(n_neighbors), 0, 1)
+
+            start_ptr = n_param * n_neighbors
+            end_ptr = start_ptr + n_neighbors
+            transformed_points[start_ptr: end_ptr, n_param] = random_points
+
+        actual_points = _transform_back(
+            transformed_points, lower_bounds, upper_bounds)
+        func_values = func(actual_points)
```
glouppe: Could we embed the `transform_back` machinery inside of `func`? In this way, we would not have to deal with it in `_local_search`, but only at the level of `paramils_smac`.

skopt/tree_opt.py
```diff
@@ -153,7 +173,8 @@ def gbrt_minimize(func, dimensions, base_estimator=None, maxiter=100,
 
 
 def forest_minimize(func, dimensions, base_estimator='rf', maxiter=100,
-                    n_points=100, n_start=10, random_state=None):
+                    n_points=100, n_start=10, random_state=None,
+                    acq_min="sampling"):
```
glouppe: Same here -> `acq_min` -> `search`

MechCoder:
Please do not make any more passes yet.

skopt/solvers/paramils.py
```diff
@@ -0,0 +1,113 @@
...
+            # and provided standard deviation.
+            random_points = np.clip(
+                transformed_x_min[dim_ind] + std * rng.randn(n_neighbors),
+                0, 1)
+
+            start_ptr = dim_ind * n_neighbors
+            end_ptr = start_ptr + n_neighbors
+            transformed_points[start_ptr: end_ptr, dim_ind] = random_points
+
+        actual_points = _transform_back(
```
MechCoder: @glouppe I think I lost your comment here. Did you mean handling the transform back into the original space inside the acquisition function?

In that case, we will not be able to use this standalone solver right?
Or did you mean something else?

glouppe: I meant in `paramils_smac`.

MechCoder: For every inner iteration, I do a function call here (https://github.com/MechCoder/scikit-optimize/blob/e57bb969414a41e5f93fc51754402f98b4734739/skopt/solvers/paramils.py#L42) which operates in the transformed space. I need to do this call to set `transformed_x_min` here. So it might not be possible to move that to the `paramils_smac` level.

glouppe: I think it should still work if you do something like this in `paramils_smac`:

```
func_ = lambda x: func((x - lower_bounds) / (upper_bounds - lower_bounds))
...
x = _local_search(func_, ...)
x = x * (upper_bounds - lower_bounds) + lower_bounds
```

MechCoder: I had the same idea in mind, but I decided against the approach because I felt lambdas in general make the code less readable.
2. I would have to do the transformation anyway back in the `paramils_smac`

glouppe: Yeah, okay. That's not really important. I just think the readability could be improved if removed the logics regarding data transformation.

skopt/tree_opt.py
MechCoder: Sorry for not being clear about what I mean about the API has to be changed. Here are my thoughts:

Assuming some-day we do have estimators that support categorical parameters natively (hopefully GP's soon) and estimators that do not, we would have to make sure that if `dimensions` contains `Categorical`, `space.transform` can return the categorical variables for the estimators that can have categorical support and one-hot encoded space for those that don't. That is what I meant by optionally transform.

Now why, I bring this here is that for categorical variables, in paramils_smac the `n_neighbors` for a given category are the other categories apart from the current given category. Because of this I would like the original categorical space and not the transformed one-hot encoded space.

(I can of course hack up a solution but I would prefer the Space API to do it directly for me)

MechCoder: I'll move this discussion to a separate issue.

MechCoder:
Here is recreating the graph from @betatim 's notebook.

It performs at par with the `ExtraTreesRegressor`. No huge improvement. :-/

index

glouppe:
Could you check on a problem with more dimensions? (where random search is expected to perform worse and worse)

MechCoder:
With Hart6:

index

PR #110, opened 2016-07-17
[WIP] Change default from min_samples_leaf=10 to min_samples_split=10

MechCoder:
Motivation:

For the current defaults, there is a possibility that no split is produced till the number of iterations till the number of iterations are 20 (worst case), which means that the predicted mean and the variance are the same till `n_iter` is 20.

This PR changes this to `min_samples_split` forcing a split when `n_iter` is greater than 10.

Let's see if I can back this up with benchmarks.
(This also ties well with SMAC's benchmarks of n_min described in page 9 of http://www.cs.ubc.ca/~hutter/papers/10-TR-SMAC.pdf)

### DIFF
--- a/skopt/tree_opt.py
+++ b/skopt/tree_opt.py
@@ -236,15 +236,15 @@ def forest_minimize(func, dimensions, base_estimator='rf', maxiter=100,
                              " are: 'rf', 'et' or 'dt', not '%s'" % base_estimator)
 
         if base_estimator == "rf":
-            base_estimator = RandomForestRegressor(min_samples_leaf=10,
+            base_estimator = RandomForestRegressor(min_samples_split=10,
                                                    random_state=rng)
 
         elif base_estimator == "et":
-            base_estimator = ExtraTreesRegressor(min_samples_leaf=10,
+            base_estimator = ExtraTreesRegressor(min_samples_split=10,
                                                  random_state=rng)
 
         elif base_estimator == "dt":
-            base_estimator = DecisionTreeRegressor(min_samples_leaf=10,
+            base_estimator = DecisionTreeRegressor(min_samples_split=10,
                                                    random_state=rng)
 
     else:


glouppe:
I agree that setting `min_samples_leaf` too high implies that we need at least `2*min_samples_leaf` iterations for the first split. On the other hand, setting it too low, e.g. `min_samples_leaf=1`, will result to poor variance estimates since the later are well calibrated only if the leaves have enough density. 
For this reason, I would change `min_samples_leaf` only, since it is the only parameter that guarantees this. Maybe we can change the default value to `min_samples_leaf=5`  instead, what do you think?

glouppe:
For example, this is what you get on toy data with extra trees and `min_samples_leaf=1,5,10`:

et-min_samples_leaf 1
et-min_samples_leaf 5
et-min_samples_leaf 10

glouppe:
Also, while we are at it, we should change the default value of `n_estimators`. It currently uses scikit-learn's `n_estimators=10` which is really much too low to obtain good variance estimates. Since the training data on which trees are built should remain small (the objective is expensive to evaluate), we can afford building many more trees, e.g. something like `n_estimators=100` (or even more). This should really help to obtain better estimates, at almost no additional cost.

betatim:
Do we gain anything in terms of statistical "power" by having more trees for a fixed, small number of evaluations of `func`? There should be some kind of scaling of `n_estimators` with the number of samples no? Might be that in practice this doesn't matter for choices of `n_estimators ~ 100`

I prefer `min_samples_leaf` as I find it easier to reason about in terms of getting a useful estimate of the variance. The value of ten was chosen arbitrarily so setting it using some information is 👍

glouppe:
> Do we gain anything in terms of statistical "power" by having more trees for a fixed, small number of evaluations of func? There should be some kind of scaling of n_estimators with the number of samples no? Might be that in practice this doesn't matter for choices of n_estimators ~ 100

Yes. You reduce the variance in the predictions which is due to randomness (and has nothing to do with the problem you are trying model), which in practice means that you are smoothing the decision surface.

Issue #102, opened 2016-07-14
Notebook: hyper-parameter optimisation of a scikit-learn estimator


glouppe:

betatim:
Related: #78 this notebook could be the foundation for #78 maybe

Issue #103, opened 2016-07-14
Notebook: comparison between GP-based, dummy and tree-based optimisation


glouppe:

betatim:
Should use a problem where it is clear whether or not dummy opt is being outperformed or not. The current benchmarks might be "too easy". If we increase the number of dimensions of the problem we should (exponentially?) reduce the chances of hitting the minimum with a random sample. #16 might contain something useful wrt a harder benchmark.

betatim:
I started work on this.

PR #112, opened 2016-07-18
[MRG] Notebook on hyperparameter optimisation

glouppe:
My attempt at #102

### DIFF
--- a/examples/hyperparameter-optimisation.ipynb
+++ b/examples/hyperparameter-optimisation.ipynb
@@ -1,0 +1,380 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tuning a scikit-learn estimator with `skopt`\n",
+    "\n",
+    "Gilles Louppe, July 2016."
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


glouppe:
#113 should be merged first, so that I can remove these ugly casts in the `objective` function.

glouppe:
Rebased and cleaned up, since #113 is merged.

MechCoder:
Awesome!

MechCoder:
Whoops, should we set `n_start=1` and the compare the performance? The first `n_start` iterations really don't tell us anything right?

MechCoder:
I get this when I do that.


index

betatim:
I'm a bit confused what we are discussing, my guess is that this is about setting the starting point to be the same for all of them?

I think we should do something similar to #103 by running the minimisation many times, and highlighting the average performance.

If we stick with `n_start=10` we should try and find an example where you are unlikely to reach the "optimum" in around ten steps. Or we reduce `n_start` but that makes life harder for the models as they have nothing to fit to -> they are returning more or less random samples.

(New issue or should we keep discussing here?)

glouppe:
+1 for showing something to what you did in the other notebook. Otherwise, methods may still be better than others only by chance.

PR #116, opened 2016-07-19
[MRG] Comparing performance of different minmizers

### LINKED ISSUE #103: Notebook: comparison between GP-based, dummy and tree-based optimisation

betatim:
Fixes #103

New example comparing the performance of different minmizers using `branin`. We could tune the hyper-parameters for the models a bit if someone has an idea.


glouppe:
Really really nice notebook! I love it! (I like its conclusion :D)

Some minor remarks:
- Comparing different models -> Comparing surrogate models?
- `evaluate` could be simplified using a list comprehension (`return [minimizer(random_state=n).func_vals for n in range(n_iter)]`)
- not sure it is necessary to set a default value to `n_iter`, since you change it below to `max_iterations`
- Maybe better to use explicit labels in the last figure ("GP" -> "Gaussian process", etc)
- Is there a reason for setting `min_samples_leaf` to different values wrt RF and ETs? (would be best to set the same value for both I guess, to not give the impression that RFs is misconfigured)

betatim:
All done. If someone has an even better idea for the title that would be good, otherwise we go with this for now.

glouppe:
LGTM!

betatim:
🍨 🍦 🍉 !

Issue #114, opened 2016-07-19
Specify starting point `x0`


glouppe:
As we have already noticed, the starting points of the optimization process may be critical in some situations. For this reason, I think it might be nice to be able to provide one or several starting points, similarly as with the `x0` parameter in `scipy.optimize.minimize`. This would be an alternative to random initialization, but one that would allow to bootstrap the process with a priori knowledge.

What do you think?

betatim:
👍 

We should have a think how/if we could combine this with the ability to "resume" the optimisation process. Similar to `warm_start`. Run `*_minimize(maxiter=25)`, inspect the output, decide that you should have done `maxiter=30`, now you want to resume from where you left off or provide the values of `func` at the first 25 iterations.

Related #68

glouppe:
We should probably put this a bit higher on our list of priorities, see eg https://twitter.com/balazskegl/status/755375668582547456 by @kegl

MechCoder:
Fixing this right now.

glouppe:
Fixed by #119

Issue #6, opened 2016-03-23
Stopping criteria


MechCoder:
What is a good stopping criteria for blackbox optimization?

MechCoder:
Right now, I just let it run it upto `max_iter`

glouppe:
We might add something like an improvement threshold between consecutive iterations, but this may be tricky to get correctly, since BO trades off exploitation for exploitation, and therefore it is far from guaranteed that the best optimum found so far will improve from one iteration to another.

AlexanderFabisch:
Most implementations of CMA-ES (another black-box optimizer) use the following stopping criteria (e.g. [this one](https://github.com/CMA-ES/c-cmaes/blob/master/cmaes_initials.par#L27)):

* max_iter / maximum number of function evaluations
* minimum fitness value (f(x) < threshold)
* *differences* of function values (does not mean that we have an improvement, exploration is small near convergence, |f(xt) - f(xt-1)| < threshold)
* difference of N best function values
* difference of evaluated parameter vectors (||xt - xt-1|| < threshold)

betatim:
Came up in #324. When working on this we should also investigate why there are so many warnings about a point already being evaluated in the unittests. My guess would be that with a convergence criteria like proposed here we would stop instead of continuing to evaluate the objective.

betatim:
I will investigate this a bit and open a PR.

betatim:
#336 created the basics and implemented "difference of evaluated parameter vectors (||xt - xt-1|| < threshold)" from https://github.com/scikit-optimize/scikit-optimize/issues/6#issuecomment-220965349

PR #119, opened 2016-07-19
[MRG] Allow providing initial points (parameters and their evaluations) to the minimize functions.

mehdidc:
Here is a proposition of a way of bootstrapping the surrogate model with user provided list of input points and their evaluations. two new parameters are provided to gp_minimize, forest_minimize, gbrt_minimize and dummy_minimize : x0 and y0 

x0 is a list of input parameters and y0 a list function evaluations of the input parameters at 
each element of x0. 

Solution for #114


betatim:
@MechCoder posted in #114 that he is also working on this. Can we make a collaborative effort between you two?

skopt/dummy_opt.py
```diff
@@ -53,13 +65,19 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
...
-    X = space.rvs(n_samples=maxiter, random_state=rng)
-
-    init_y = func(X[0])
+    if x0 is None:
+        x0 = []
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+    if len(x0) != len(y0):
+        raise ValueError('x0 and y0 should have the same length')
+    X = x0 + space.rvs(n_samples=maxiter, random_state=rng)
```
MechCoder: This will not work if `x0` is not a list. You should have a condition that converts x0 and y0 to lists before this line.

skopt/tree_opt.py
```diff
@@ -33,8 +33,15 @@ def _tree_minimize(func, dimensions, base_estimator, maxiter,
         raise ValueError("Total number of iterations set by maxiter has to"
                          " be larger or equal to n_start.")
 
-    Xi = space.rvs(n_samples=n_start, random_state=rng)
-    yi = [func(x) for x in Xi]
+    if x0 is None:
+        x0 = space.rvs(n_samples=n_start, random_state=rng)
```
MechCoder: This assumes that if `x0` is provided, `n_start` is not useful, right? I would go with your logic in `dummy_opt.py` instead. WDYT?

skopt/tests/test_opt.py
```diff
@@ -0,0 +1,32 @@
...
+            branin, [(-5.0, 10.0), (0.0, 15.0)], x0=x0, y0=y0,
+            random_state=0, maxiter=100)
+        assert_array_equal(res.x_iters[0:len(x0)], x0)
+        assert_array_equal(res.func_vals[0:len(y0)], y0)
+
+        res = optimizer(
+            branin, [(-5.0, 10.0), (0.0, 15.0)], x0=x0,
+            random_state=0, maxiter=100)
+        assert_array_equal(res.x_iters[0:len(x0)], x0)
+        assert_array_equal(res.func_vals[0:len(y0)], y0)
```
MechCoder: Can you check length of `res.x_iters` and `res.func_vals` also, they should be equal to `maxiter+ n_init`. (Also you might want to reduce the number of iterations to lesser since we are not testing the correctness of the algorithm here)

skopt/tree_opt.py
```diff
@@ -207,6 +230,19 @@ def forest_minimize(func, dimensions, base_estimator='rf', maxiter=100,
     * `n_points` [int, default=1000]:
         Number of points to sample when minimizing the acquisition function.
 
+    * `x0` [list, shape=(n_initial,)]:
+        List of initial input parameters used to bootstrap the surrogate
+        model. Note that if `x0` is provided, no random initialization
+        points are used, so in that case `n_start` is ignored.
```
MechCoder: Oh, should we ignore `n_start`? Sometimes the user might want to use bootstrap along with random initialization. For this use case, we should make the user explicitly supply `n_start=0` instead of implicitly ignoring a parameter.

skopt/tests/test_opt.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_init_vals():
+    optimizers = [
+        dummy_minimize,
+        gp_minimize,
+        forest_minimize,
+        gbrt_minimize
+    ]
+    x0 = [[1, 2], [3, 4], [5, 6]]
+    y0 = map(branin, x0)
```
MechCoder: This will be a map object in Python3, hence you need to explicitly convert it to a list internally.

mehdidc:
@betatim Sure, sorry @MechCoder I have just seen that you were working on this as well.

MechCoder:
Thanks for the addition @mehdidc ! Let us know when you have addressed the comments.

skopt/dummy_opt.py
mehdidc: yes makes sense, I will add this.

skopt/tree_opt.py
mehdidc: Yes, it ignores n_start if x0 is provided. Yes, I agree with what you said in the comment below, we could just provide n_start=0 to have this behavior.

skopt/dummy_opt.py
```diff
@@ -31,6 +32,17 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
         Number of iterations to find the minimum. In other words, the
         number of function evaluations.
 
+    * `x0` [list, shape=(n_initial,)]:
+        List of initial input parameters.
```
glouppe: parameters -> points

skopt/dummy_opt.py
```diff
@@ -31,6 +32,17 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
         Number of iterations to find the minimum. In other words, the
         number of function evaluations.
 
+    * `x0` [list, shape=(n_initial,)]:
```
glouppe: `list` or `list of lists`

skopt/dummy_opt.py
```diff
@@ -53,13 +65,19 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
...
-
-    init_y = func(X[0])
+    if x0 is None:
+        x0 = []
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+    if len(x0) != len(y0):
+        raise ValueError('x0 and y0 should have the same length')
+    X = x0 + space.rvs(n_samples=maxiter, random_state=rng)
+    init_y = func(X[0]) if len(y0) == 0 else y0[0]
```
glouppe: if `len(y0)==0`, does this call `func` once more than permitted? (it is called twice for `func(X[0])` in that case)

skopt/gp_opt.py
```diff
@@ -106,6 +107,19 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
     * `n_restarts_optimizer` [int, default=10]:
         The number of restarts of the optimizer when `search` is `"lbfgs"`.
 
+    * `x0` [list, shape=(n_initial,)]:
+        List of initial input parameters used to bootstrap the surrogate
+        model. Note that if `x0` is provided, no random initialization
+        points are used, so in that case `n_start` is ignored.
```
glouppe: same comments for the docstring

glouppe:
Thanks! Before going further, I think we should agree on the semantics of `x0`, `n_start` and `maxiter`, and how they interplay with each others. This is not really clear at the moment. E.g. if `n_start > 0` and `x0 is not None`, what is the expected behaviour?

MechCoder:
Either one of these two:

1. If both `n_start` and `x0` provided, we assume that the user wants to do a random start in addition to `x0`. If he would like to disable this, then he should explicitly set `n_start=0`

2. Allow `n_start='auto'`. If `x0` is provided, this will be set to zero, otherwise it will be set to the default value of 10.

WDYT?

glouppe:
And then what about `maxiter`, does this correspond to the number of calls to func?  (i.e. does this behaves differently when `y0 is None` and vice-versa?)

MechCoder:
Are you +1 for the first or the second behaviour of `n_start`? Thinking about it, the second one seems a bit better to me.

I believe the behavior of `maxiter` should be the number of calls to func as you say and the number of surrogate model fits should be `maxiter - n_start`

MechCoder:
Oh sorry, I misread. That is a tough question :P

MechCoder:
I tend to look at `maxiter - n_start` as the number of surrogate model fits.

1. If `x0`, `y0` and `n_start` is provided, we assume that the user wants to random start along with the given initial parameters. In that case the behaviour of `maxiter` is obvious.

2. If only `x0` is provided, we assume that the user is lazy to do the function call outside the `*_minimize` functions and hence do not consider these function calls either in `n_start` or `maxiter`. We still do `n_start` random initializations and `maxiter - n_start` surrogate model fits.

How does that proposal sound?

glouppe:
In case of 1. does this mean we build `maxiter - n_start - len(x0)` models? If yes, then +1 for this.

MechCoder:
Can we change `maxiter` to `model_iter` (or some other better name?). That will resolve confusion and make things clearer?

So irrespective of what `x0` and `y0` are, we do `n_start` random calls to func and build `model_iter` number of models.

glouppe:
Let's rename `n_start` with `n_random_start` to be self-explanatory and `maxiter` with `n_model_iter` with the behaviour you suggest?

MechCoder:
https://github.com/scikit-optimize/scikit-optimize/pull/123

mehdidc:
Thanks for your comments @MechCoder  and @glouppe, I updated the code. Let me know what you think about it. Also, I guess I should update the code after #123 is merged.

betatim:
Thanks a lot @mehdidc for the contribution and also for having patience with us trying to figure stuff out on the fly!

---

> In case of 1. does this mean we build `maxiter - n_start - len(x0)` models? If yes, then +1 for this.

I would have expected that `maxiter - n_start` models would be fit, as `maxiter` sets the budget for the minimiser so it should be independent of `len(x0)`. As a user if I provide `x0` and `maxiter=10` I am thinking: please do another 10 iterations and here are some points I previously evaluated. Instead of having to set `maxiter = len(x0) + 10` to achieve that.

glouppe:
> I would have expected that maxiter - n_start models would be fit, as maxiter sets the budget for the minimiser so it should be independent of len(x0). As a user if I provide x0 and maxiter=10 I am thinking: please do another 10 iterations and here are some points I previously evaluated. Instead of having to set maxiter = len(x0) + 10 to achieve that.

Yes, but only if `y0 is not None`. If `y0 is None`, then it requires calls on `func`.

glouppe:
Also i think `maxiter` is a terrible name. We should be more specific, and rename to that something like `n_calls`.

betatim:
1. rename `maxiter` to `n_calls`. This sets the upper limit of how often a minimiser will call `func` (see #123)
1. if `x0 is not None and y0 is None` use `len(x0)` calls to evaluate `func`,  the number of models fitted: `n_calls - len(x0) - n_random_start`
1. if `x0` and `y0` are provided, perform `n_random_start` random samples, the number of models fitted: `n_calls - n_random_start`

Reasoning for 2.: if `func` is expensive you won't be lazy and not provide `y0` if you have it -> if `y0` is not provided it means we need to call `func` which means we are using up the budget set by `n_calls`. If you are resuming/warm starting then you have `x0` and `y0` so you should provide them.

Show your agreement with emoji reactions, if you disagree write a comment 📝 .

MechCoder:
@mehdidc Would you be able to rebase over master? Sorry for the troubles.

mehdidc:
@MechCoder Sure, I will do that.

MechCoder:
I think you might have to rebase again. @glouppe changed the default n_calls of dummy_minimize in master. (Sorry :/)

skopt/dummy_opt.py
```diff
@@ -52,20 +67,37 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
     """
     rng = check_random_state(random_state)
     space = Space(dimensions)
-    X = space.rvs(n_samples=n_calls, random_state=rng)
-
-    init_y = func(X[0])
+    if x0 is None:
+        x0 = []
+    if type(x0) is not list:
```
MechCoder: You can do `x0 = list(x0)` directly here.

skopt/dummy_opt.py
```diff
@@ -52,20 +67,37 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
...
+        x0 = list(x0)
+    if len(x0) > 0 and type(x0[0]) is not list:
+        x0 = [x0]
+    n_random_starts = n_calls
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+        n_random_starts -= len(y0)
+    if type(y0) is not list:
```
MechCoder: I think we can support y0 to be an iterable (http://stackoverflow.com/a/1952655/1170730). I'm saying this because you are doing `y = np.array(y)` below.

skopt/dummy_opt.py
```diff
@@ -27,9 +28,23 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
         - an instance of a `Dimension` object (`Real`, `Integer` or
           `Categorical`).
 
-    * `n_calls` [int, default=100]:
+    * `n_calls` [int, default=1000]:
         Number of calls to `func` to find the minimum.
 
+    * `x0` [list or list of lists]:
```
glouppe: It can also be `None`, in which case the behaviour should be described.

skopt/dummy_opt.py
```diff
@@ -27,9 +28,23 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
...
+        List of initial input points (if it is a list of lists)
+        or an initial input point (if it is a list).
+
+    * `y0` [list or scalar]
```
glouppe: same

skopt/dummy_opt.py
```diff
@@ -52,20 +67,37 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
...
+    n_random_starts = n_calls
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+        n_random_starts -= len(y0)
+    if type(y0) is not list:
+        y0 = [y0]
+    if len(x0) != len(y0):
+        raise ValueError("x0 and y0 should have the same length")
+    X = x0 + space.rvs(n_samples=n_random_starts, random_state=rng)
+    init_provided = False if len(y0) == 0 else True
```
glouppe: yuck. 

`init_provided = len(y0) != 0`

skopt/gp_opt.py
```diff
@@ -95,7 +96,9 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
         Number of calls to `func`.
         If `n_random_starts` > 0, `n_calls - n_random_starts`
         additional evaluations of `func` are made that are guided
-        by the `base_estimator`.
+        by the `base_estimator`. If `x0` is provided but not `y0`
+        then `n_calls - len(x0) - n_random_starts` evaluations
+        are made instead of `n_calls - n_random_starts` .
```
glouppe: In total `n_calls` evaluations are always made. 

We should rather say something like: "If x0 is provided but not y0, then these points are first evaluated, followed by `n_random_starts` random initialization points,  and finally `n_calls - len(x0) - n_random_starts` subsequent evaluations as guided by the surrogate model." (Text can be improved, but you get the idea)

skopt/dummy_opt.py
mehdidc: Ok, it should also be possible for x0 to be an iterable (and not only a list) right ?

skopt/gp_opt.py
mehdidc: Ok, what do you think about this  : 


      * `n_calls` [int, default=100]:
          Number of calls to `func`.
          If `x0` is provided but not `y0`, then the elements of `x0` are
          first evaluated, followed by `n_random_starts` evaluations.
          Finally, `n_calls - len(x0) - n_random_starts` evaluations are
          made guided by the surrogate model. If `x0` and `y0` are both
          provided then `n_random_starts` evaluations are first made then
          `n_calls - n_random_starts` subsequent evaluations are made
          guided by the surrogate model.

skopt/dummy_opt.py
betatim: Why not `init_provided = x0 is not None`?

skopt/gp_opt.py
betatim: Can we move this text further up in the doc string? It relates to both `x0`, `n_calls`, `n_random_starts`, etc and is about the whole function not just the `n_calls` parameter.

For `n_calls` I propose: Maximum number of calls to `func`.

betatim:
There are several occurrences of `if type(x) is int` which should be `if isinstance(x, list)` (same for other types).

To check if a list contains an item (`len(x) > 0`) in python you simply test the truth value of the list: `if x:` instead of computing the length.

Take a look at [PEP8](https://www.python.org/dev/peps/pep-0008/) which we try and follow when possible. You can also install [`pep8` the command line program](https://pypi.python.org/pypi/pep8) to give you advice.

skopt/gp_opt.py
mehdidc: Just after dimension or right at the top, before func ? as the ordering of the parameters on the signature should correspond to the one on the doc, after dimesions I can switch ordering in the signature as well, but it would not be possible before func because they are not optional.

skopt/dummy_opt.py
MechCoder: We had decided that the `rvs` from Space objects should return a list of lists over here (https://github.com/scikit-optimize/scikit-optimize/issues/86) . TLDR; it used to return a numpy array of dtype object when the elements of the lists are of different types (float | int)

So for this PR to make things easier, I would suggest to just support list of lists.

For y, I  have not strong opinion and I am ok both ways (whether to support lists, iterables either in this PR or outside)

skopt/gp_opt.py
betatim: Move the general discussion about the interplay of n_calls and n_random_starts to above the section `Parameters`. Keep the documentation of the parameter `n_calls` where it is :) This would be my proposal.

skopt/tree_opt.py
```diff
@@ -209,6 +263,21 @@ def forest_minimize(func, dimensions, base_estimator='et', n_calls=100,
     * `n_points` [int, default=1000]:
         Number of points to sample when minimizing the acquisition function.
 
+    * `x0` [list or list of lists or None]:
+        List of initial input points (if it is a list of lists)
```
MechCoder: Can you organize these into points?

Similar to what is being done here (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/space.py#L150)

skopt/dummy_opt.py
```diff
@@ -6,7 +7,8 @@
 from .space import Space
 
 
-def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
+def dummy_minimize(func, dimensions, n_calls=1000,
+                   x0=None, y0=None, random_state=None):
```
glouppe: Can you rebase on master? `n_calls` is now `100` by default

skopt/dummy_opt.py
```diff
@@ -52,20 +69,38 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
...
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+        n_random_starts -= len(y0)
+    if isinstance(y0, Iterable):
+        y0 = list(y0)
+    else:
+        y0 = [y0]
+    if len(x0) != len(y0):
+        raise ValueError("x0 and y0 should have the same length")
+    X = x0 + space.rvs(n_samples=n_random_starts, random_state=rng)
```
glouppe: How could this possibly work? `n_random_starts` does not exist in `dummy_minimize`

MechCoder: I haven't had a detailed look yet, but I think it is renamed above.

glouppe: Oh yeah right

MechCoder: speaking of which; it should be passed directly it gets a bit confusing to read.

skopt/dummy_opt.py
```diff
@@ -52,20 +69,38 @@ def dummy_minimize(func, dimensions, n_calls=1000, random_state=None):
...
+    y = np.array(y)
     res = OptimizeResult()
     best = np.argmin(y)
     res.x = X[best]
     res.fun = y[best]
     res.func_vals = y
     res.x_iters = X
     res.space = space
-
     return res
```
glouppe: can you leave this space please?

skopt/dummy_opt.py
```diff
@@ -28,7 +30,23 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
           `Categorical`).
 
     * `n_calls` [int, default=100]:
-        Number of calls to `func` to find the minimum.
+        Maximum number of calls to `func` to find the minimum.
```
MechCoder: For now, I think we should just retain the previous better definition of better `n_calls`, till we have better stopping criteria. See https://github.com/scikit-optimize/scikit-optimize/issues/6

skopt/dummy_opt.py
```diff
@@ -52,14 +70,32 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
     rng = check_random_state(random_state)
     space = Space(dimensions)
-    X = space.rvs(n_samples=n_calls, random_state=rng)
-
-    init_y = func(X[0])
+    if x0 is None:
+        x0 = []
+    x0 = list(x0)
+    if x0 and not isinstance(x0[0], list):
+        x0 = [x0]
```
MechCoder: I find the code a bit easier to read, if it is restructured like this:

```python
if x0 is None:
    x0 = []
elif not isinstance(x0[0], list):
    x0 = [x0]

if not isinstance(x0, list):
    raise ValueError("Expected x0 to be a list, but got %s" % type(x0))

```

WDYT: Note we can decide about support for `x0` to be a numpy array in a follow-up PR

skopt/dummy_opt.py
```diff
@@ -52,14 +70,32 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+    if y0 is None:
+        y0 = [func(x) for x in x0]
+        n_calls -= len(y0)
+    if isinstance(y0, Iterable):
+        y0 = list(y0)
+    else:
+        y0 = [y0]
```
MechCoder: Can we restructure like this?

```python
if y0 is None and x0:
    y0 = [func(x) for x in x0]
    n_func_calls = len(y0)
elif x0:
    if instance(y0, Iterable):
        y0 = list(y0)
    elif isinstance(y0, numbers.Number):
        y0 = [y0]
     else:
        raise ValueError
     if len(x0) != len(y0):
         raise ValueError
```

And do explicitly do `n_samples = n_calls - len(y0)` below, so that it a bit easier to read.

skopt/dummy_opt.py
```diff
@@ -52,14 +70,32 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+    if len(x0) != len(y0):
+        raise ValueError("x0 and y0 should have the same length")
+    X = x0 + space.rvs(n_samples=n_calls, random_state=rng)
+    init_provided = len(y0) != 0
```
MechCoder: Can we just do it like this:

```python
X_new = space.rvs(n_samples=n_calls - n_func_calls, random_state=rng)
init_y = func(X_new[0])
if not np.isscalar(init_y):
    raise ValueError..
y_new = [init_y] + [func(X_new[i]) for i in range(1, n_calls - n_func_calls)]
```

And then;
```python
y  = y0 + y_new
```

skopt/gp_opt.py
```diff
@@ -142,25 +166,47 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
...
 
-    if n_random_starts <= 0:
+    # Initialize with provided points (x0 and y0) and/or random points
+    if x0 is None:
+        x0 = []
+    x0 = list(x0)
+    if x0 and not isinstance(x0[0], list):
+        x0 = [x0]
+
+    n_init = len(x0) if y0 is None else 0
```
MechCoder: Can we rename this to `n_init_func_calls`?

skopt/tests/test_opt.py
```diff
@@ -0,0 +1,65 @@
+from sklearn.utils.testing import assert_array_equal, assert_equal
```
MechCoder: Can we move this to `tests/test_common.py`?

skopt/tests/test_opt.py
```diff
@@ -0,0 +1,65 @@
...
+
+def test_init_vals():
+    n_random_starts = 5
+    optimizers = [
+        dummy_minimize,
+        partial(gp_minimize, n_random_starts=n_random_starts),
+        partial(forest_minimize, n_random_starts=n_random_starts),
+        partial(gbrt_minimize, n_random_starts=n_random_starts)
+    ]
+    x0 = [[1, 2], [3, 4], [5, 6]]
```
MechCoder: Can you also check where `x0 is a list of categories?` using `bench4`

MechCoder:
@mehdidc Almost there! Thanks!

skopt/dummy_opt.py
mehdidc: it looks better like this. But one difference with the previous code is that when y0 is provided
it will be used to check the init and raise the exception without evaluation, whereas here we will wait for `n_calls - n_func_calls` evaluations before raising the exception.

MechCoder: I'm not sure what you mean, if `y0` is provided, you are making the checks here right (https://github.com/scikit-optimize/scikit-optimize/pull/119/files/98df84c12a8d13b5f62393b16357d7ee54d864c0#r71977002)?

MechCoder: If you are worried about 1 function evaluation, we should raise the Exceptions for that case in the block above then (where y0 is initialized)

mehdidc: Sorry I meant 1 evaluation not `n_calls - n_func_calls`. I will then add an Exception also when y0 is initialized.

mehdidc: For finding other names to `X_new` and `y_new`, maybe `X_rest`/`y_rest` or `X_remaining`/`y_remaining` or `X_left`/`y_left` ?

MechCoder: I'm ok with `X_left`

skopt/dummy_opt.py
```diff
@@ -30,6 +33,22 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
     * `n_calls` [int, default=100]:
         Number of calls to `func` to find the minimum.
 
+    * `x0` [list or list of lists or None]:
+        Initial input points.
+        - If it is a list of lists, use it as a list of input points.
+        - if it is a list, use it as an initial input point.
```
glouppe: as a _single_ initial input point

glouppe: Also, `If` with an upper `I`

skopt/dummy_opt.py
```diff
@@ -30,6 +33,22 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+        - If it is `None`, no initial input points are used.
+
+    * `y0` [list or scalar or None]
+        Evaluation of initial input points.
+        - if it is a list, then it corresponds to evaluations of the function
```
glouppe: if -> If

skopt/dummy_opt.py
```diff
@@ -30,6 +33,22 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+          at each element of `x0` : the i-th element of `y0` corresponds
+          to the function evaluated at the i-th element of `x0`.
+        - If it is a scalar, then it corresponds to the evaluation of the
+          function at `x0`.
+        - if it is None and `x0` is provided, then the function is evaluated
```
glouppe: if -> If

skopt/dummy_opt.py
```diff
@@ -52,14 +71,45 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+        if not all(map(np.isscalar, y0)):
+            raise ValueError(
+                "y0 elements should be scalars")
+        n_func_calls = 0
+    else:
+        y0 = []
+        n_func_calls = 0
+
+    X_left = space.rvs(n_samples=n_calls - n_func_calls, random_state=rng)
+    init_y = func(X_left[0])
```
glouppe: That's a border case, but if `len(x0) == n_calls` and `y0 is None`, then you evaluate `func` once more than allowed.

skopt/dummy_opt.py
```diff
@@ -52,14 +71,45 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+    elif not isinstance(x0[0], list):
+        x0 = [x0]
+
+    if not isinstance(x0, list):
+        raise ValueError("Expected x0 to be a list, but got %s" % type(x0))
 
-    init_y = func(X[0])
+    if y0 is None and x0:
+        y0 = [func(x) for x in x0]
+        n_func_calls = len(y0)
```
glouppe: I would simply remove this variable and do `n_calls -= len(y0)`

skopt/dummy_opt.py
```diff
@@ -52,14 +71,45 @@ def dummy_minimize(func, dimensions, n_calls=100, random_state=None):
...
+        n_func_calls = 0
+    else:
+        y0 = []
+        n_func_calls = 0
+
+    X_left = space.rvs(n_samples=n_calls - n_func_calls, random_state=rng)
+    init_y = func(X_left[0])
     if not np.isscalar(init_y):
         raise ValueError(
             "The function to be optimized should return a scalar")
```
glouppe: Would be best to trigger this error on the first call to `func`. Here we may already have consumed `len(x0)` calls.

skopt/gp_opt.py
```diff
@@ -41,6 +44,16 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
...
+    The total number of evaluations, `n_calls`, are performed like the
+    following. If `x0` is provided but not `y0`, then the elements of `x0`
+    are first evaluated, followed by `n_random_starts` evaluations.
+    Finally, `n_calls - len(x0) - n_random_starts` evaluations are
+    made guided by the surrogate model. If `x0` and `y0` are both
+    provided then `n_random_starts` evaluations are first made then
+    `n_calls - n_random_starts` subsequent evaluations are made
+    guided by the surrogate model.
+
+
```
glouppe: can you remove one of those spaces?

skopt/gp_opt.py
```diff
@@ -142,25 +167,56 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
...
+    Xi = x0 + space.rvs(n_samples=n_random_starts, random_state=rng)
+    yi = y0 + [func(x) for x in Xi[len(x0):]]
 
-    n_model_iter = n_calls - n_random_starts
-    # First points
-    Xi = space.rvs(n_samples=n_random_starts, random_state=rng)
-    yi = [func(x) for x in Xi]
     if np.ndim(yi) != 1:
         raise ValueError(
             "The function to be optimized should return a scalar")
```
glouppe: same here

skopt/tests/test_common.py
```diff
@@ -8,7 +8,7 @@
 from sklearn.utils.testing import assert_equal
 from sklearn.utils.testing import assert_raises
 
-from skopt.benchmarks import branin
+from skopt.benchmarks import branin, bench4
```
glouppe: one import per line

glouppe:
Thanks @mehdidc for contribution and patience. Just a few more minor changes and it should be good to go :)

glouppe:
Thanks for addressing those last comments Mehdi! Merging this.

betatim:
🍨

PR #120, opened 2016-07-19
[MRG] better defaults for `forest_minimize`

glouppe:
Supersedes #110


MechCoder:
Could you please be a bit patient?

I was working on that well!!!

glouppe:
Yeah, but what @kegl noticed is a bit embarrasing :/ 

See https://twitter.com/glouppe/status/755312372789772288

betatim:
Did we get a copy of @kegl's script? If possible we should add it/version of it as a test. TBH I would need a bit of explaining to understand exactly what this is showing/how it happens:
cnvuratxgaa84z0

For choosing the defaults we should try and be guided by some benchmarks or something so that we avoid having to change them often. Right now they are set to "random values" which don't work very well. As this is still `pre0.1` we can be Swiss about this: think about the problem, then think about it again, solve it 😃

MechCoder:
Yes that is worrying.

But in any case, no activity for a few hours does not necessarily mean that the person will not be working on it again. If that is true, then all of scikit-learn's pull requests should be superseded.

glouppe:
Sorry about scooping you, just dont want to miss the momentum we have at the moment. I'll try to be more patient...

glouppe:
@kegl Could you retry with those new defaults or give us the script you used? That would be super helpful :)

kegl:
It's better. It still exploits aggressively which is not necessarily a bad thing given that all the others are like random search (on this very flat problem). It found the best hypers so far.

For the hypers, I don't think we should fix them. But these are the questions that only experimentation can decide, that's why we think we should go for a framework rather than a new fixed library.

plot

plot

glouppe:
> It's better. It still exploits aggressively which is not necessarily a bad thing given that all the others are like random search (on this very flat problem). It found the best hypers so far.

Not sure what you think, but I am not convinced this specific benchmark is very representative: the function should be mostly flat when `n_estimators` is a enough and similarly for `max_depth`. We should therefore be cautious before over-interpreting the results. A better benchmark would be instead to tune for `max_features`, `min_samples_split` or `min_samples_leaf`. Alternatively, tuning a boosting model is also more interesting I think (this is what I used for the notebook example).

glouppe:
For the exploration, it might be due to the value of `kappa` is set too low. Also, `forest_minimize` currently only supports the Expected Improvement as acquisition function -- not sure what is used in SMAC & co.

glouppe:
> For the hypers, I don't think we should fix them.

Just for the sake of documenting this in case others are reading: `forest_minimize` takes an optional `base_estimator` object, through which all these parameters can be user-defined.

kegl:
I think most of the packages hyperopt the hypers of the base_estimator. But these are the things we could try if we had the framework -> [https://github.com/scikit-optimize/scikit-optimize/issues/128](url)

MechCoder:
> We surely run max marginal likelihood for setting the GP hypers after adding each training point.

Sorry for not reading through the rest of the discussion but the hyperparameters of the GP kernel (such as the length scale of the kernel etc) are tuned for every iteration as well by maximizing the log-likelihood if that is what you mean. It is just that the hyperparameters of the GaussianProcessRegressor are fixed for every iteration. (that is the parameters in the constructor of http://scikit-learn.org/dev/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor)

kegl:
OK, what I meant was that the skopt.RF hypers should also be autotuned. I wouldn't expect a user using skopt wanting to tune the hypers of the skopt, it sort of defeats the purpose.

glouppe:
Do you know of any paper or implementation of a smac-like algorithm in which the hyper-parameters are not fixed and rather autotuned like you suggest? (Note that the same goes for GPs to some extent, the choice of the kernel is really critical to get the best results)

kegl:
I suppose pysmac does it, but it's not a detail that papers usually discuss :). But again, think about the broader issue. There are tons of little details like this that are not yet clear and hard to settle without a testbench. I have plenty of ideas that I never wrote down because they are too small for publication, which I could test if we had a framework.

glouppe:
> I suppose pysmac does it

I wouldnt bet on that. From looking e.g. at https://github.com/sfalkner/pySMAC/blob/master/pysmac/optimizer.py#L108 a lot of hyper-parameters are passed, and I doubt they are later tuned (e.g. the number of trees, whether or not to use bootstrap, `max_features`, etc) -- this is something very difficult to do efficiently with a forest.

kegl:
OK, you know this better than I :). I would definitely tune it, but then it's recursive :). GP solved it by using bfgs in the second level (since it is arguably fast, small number of points we can just take any stochastic optimizer to stop the recursion). We could just run a small grid search.

glouppe:
> There are tons of little details like this that are not yet clear and hard to settle without a testbench. I have plenty of ideas that I never wrote down because they are too small for publication, which I could test if we had a framework.

Yes, there are also many little things which I find arbitrary/puzzling in SMAC. This is one of the reasons we started this package :)

MechCoder:
Sorry but I don't quite follow. Tuning the meta-hyperparameters of the meta-forest would give us the best fit for the hyperparameters. But that does not solve the inherent inability of the Random Forest to not generalize to unseen data no? Or in other words, from @glouppe 's graph, it is clear that the GP will grab unseen points but the RF has a flat variance in regions outside the unseen points. So the forest will just grab the points that are closest to the previous best.

The reason it might help to tune the hyperparameters of the RF is to get to this previous best quickly in the first few iterations, where it will get stuck at points closest to the previous best. (But I presume, this is not what we want either, we also want to explore unseen regions)

MechCoder:
Btw @kegl https://github.com/scikit-optimize/scikit-optimize/pull/119 in combination with https://github.com/scikit-optimize/scikit-optimize/issues/148 might be the framework for testing that you are looking for :-)

kegl:
The other problem IMHO is that the way we use RFs for Bayesian regression is heuristic: it's not clear what the std of tree predictions mean. In my experience they underestimate the std of the process (compared to GPs).

glouppe:
> it's not clear what the std of tree predictions mean.

Yes this definition is plain wrong and does  not correspond to the correct statistical quantity, but this is not the one we use, nor what is used in SMAC (despite the very very very confusing  description that can be found in the corresponding papers -- the original paper used it, but then they switched to the correct quantity in a subsequent paper). 

In our case, we effectively approximate `Var(Y|X)` by using the empirical  variance of the (weighted) training samples reaching the same leaves (accross all trees) as `X`.

PR #122, opened 2016-07-19
Fixing bug in inputs to bench3 in example scripts

cmmalone:
I was having trouble getting `examples/plot_gp_minimize_1d.py` and `examples/plot_gbrt_minimize_2d.py` to run; it was throwing an IndexError when bench3 was called:

`python plot_gp_minimize_1d.py 

==================================================================
Plot lower confidence bound as a function of number of iterations.
==================================================================

  File "plot_gp_minimize_1d.py", line 16, in <module>
    func_values = [bench3(xi) for xi in x]
  File "plot_gp_minimize_1d.py", line 16, in <listcomp>
    func_values = [bench3(xi) for xi in x]
  File "/Users/civisemployee/src/scikit-optimize/skopt/benchmarks.py", line 38, in bench3
    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2))
IndexError: invalid index to scalar variable.`

bench3 expects a list, but it was getting floats, so in the example scripts I changed the input type going into bench3 so it's a single-item list instead of a float.


glouppe:
Thanks! I am merging these changes.

Note that these scripts are currently being replaced with notebooks. They are still there in the repo because we have not ported everything yet.

betatim:
We send 🍨 as a thank you!

Like @glouppe mentioned, we are moving our examples to notebooks, so if you like one of the examples and want to switch it over, go ahead.

PR #123, opened 2016-07-19
[API] Remove max_iter and n_start and add n_model_iter and n_random_start

MechCoder:
Remove `max_iter` and add support for `n_random_start` and `n_model_iter`.

`n_random_start` should be the number of random point sampled initially.
`n_model_iter` should be the number of surrogate model fits.


MechCoder:
I would wait for the opinions of @betatim as well.

betatim:
What is the thinking behind this compared to using `n_start` and `maxiter`? I like having one parameter that is easy to understand and corresponds to "my expensive function will never ever be called more than `maxiter` times". As a user I know approximately how often I can afford to eval `func`, I don't care sooo much about how the algorithm decides to use those evaluations.

Renaming `n_start` to the more descriptive `n_random_eval` or `n_random_start` could be good.  Then calculate `n_model_iter = maxiter - n_random_start` internally.

betatim:
Just read the discussion in #119. Put my previous comment on hold, will have to think a bit more.

glouppe:
Not sure what is best. It is true that `n_model_iter` is forcing the user to understand what is going on under the hood, while something like a `n_calls` is straightforward to understand, and then it is the business of the algorithm to consume these calls the way it sees fit.

betatim:
I think internally it is a good move to convert to `n_model_iter` because it makes the code clearer. For the user experience I would keep a simple parameter that sets the upper limit on the number of calls to `func` (ignoring convenience calls if `x0 is not None and y0 is None`).

betatim:
Tried to [summarise things](https://github.com/scikit-optimize/scikit-optimize/pull/119#issuecomment-233846575)

If we agree on that plan, will you take care of `n_calls` and `n_random_start` (or `n_random_startS` I can never work out the policy on plural S??) in this PR @MechCoder?

MechCoder:
I agree with your summary, the parameter names make it clear.

MechCoder:
I did the renames. I did not modify the hyperparameter optimization notebook since it has to be changed again anyway to provide the initial points.

skopt/gp_opt.py
```diff
@@ -24,8 +24,8 @@ def _acquisition(X, model, y_opt=None, method="LCB", xi=0.01, kappa=1.96):
 
 
 def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
-                kappa=1.96, search="sampling", maxiter=1000, n_points=500,
```
MechCoder: 1000 is a pretty expensive default to have...

glouppe: indeed :-)

skopt/gp_opt.py
```diff
@@ -91,16 +91,14 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
         lbfgs is run for 10 iterations optimizing the acquisition function
         over the Gaussian prior.
 
-    * `maxiter` [int, default=1000]:
-        Number of iterations to find the minimum. Note that `n_start`
-        iterations are effectively discounted, such that total number of
-        function evaluations is at most `maxiter`.
+    *  `n_calls` [int, default=100]:
+        Number of times the expensive function is called.
```
glouppe: I think we should add the long version of the explanation, detailing the interplay with `n_random_starts`.

examples/hyperparameter-optimisation.ipynb
```diff
@@ -121,7 +121,7 @@
    "outputs": [],
    "source": [
     "from skopt import gp_minimize\n",
-    "res_gp = gp_minimize(objective, space, maxiter=50, random_state=0)"
+    "res_gp = gp_minimize(objective, space, n_random_start=1, n_model_iter=50, random_state=0)"
```
glouppe: Could you still update this notebook? Note sure when #119 will be ready, but your PR clearly is. In the meantime, we should not break things.

glouppe:
+1 besides my comments.

@mehdidc can you wait for this to be merge before proceeding with your PR? (sorry for the back and forth, as you can see we are still figuring out things as we go along...)

MechCoder:
OK I addressed your comments and also renamed `maxiter` to `n_calls` in `dummy_minimize`

examples/hyperparameter-optimisation.ipynb
```diff
@@ -330,20 +330,27 @@
...
      },
      "metadata": {},
      "output_type": "display_data"
     }
    ],
    "source": [
-    "from skopt.plots import plot_convergence\n",
-    "plot_convergence((\"gp_optimize\", res_gp),\n",
-    "                 (\"forest_optimize\", res_forest),\n",
-    "                 (\"dummy_optimize\", res_dummy))"
```
glouppe: Have you rebased on master? This should stay.

glouppe: Similarly, I am wondering what are the changes for the other notebooks?

glouppe: (Github is really not ready for notebooks...)

skopt/dummy_opt.py
```diff
@@ -27,9 +27,11 @@ def dummy_minimize(func, dimensions, maxiter=1000, random_state=None):
...
           `Categorical`).
 
-    * `maxiter` [int, default=1000]:
-        Number of iterations to find the minimum. In other words, the
-        number of function evaluations.
+    * `n_calls` [int, default=100]:
+        Number of calls to `func`.
+        If `n_random_starts` > 0, `n_calls - n_random_starts`
+        additional evaluations of `func` are made that are guided
+        by the `base_estimator`.
```
glouppe: I guess you copied and pasted too quickly ;) this does not apply for `dummy_minimize`.

skopt/gp_opt.py
```diff
@@ -91,16 +91,17 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="LCB", xi=0.01,
...
+        additional evaluations of `func` are made that are guided
+        by the `base_estimator`.
 
     * `n_points` [int, default=500]:
         Number of points to sample to determine the next "best" point.
         Useless if search is set to `"lbfgs"`.
 
-    * `n_start` [int, default=10]:
+    * `n_random_starts` [int, default=10]:
         Number of random initialization points.
```
glouppe: Maybe we should update this as well for something like "Number of evaluations of `func` with random initialization points."

MechCoder:
Should look ok now?

glouppe:
Yup, all good now, thanks! Feel free to merge once Travis is happy

Issue #121, opened 2016-07-19
Doc: optimisation -> optimization


glouppe:
As suggested by @rasbt

https://twitter.com/rasbt/status/755411036136615937

betatim:
[seven starting points](https://github.com/scikit-optimize/scikit-optimize/search?utf8=%E2%9C%93&q=optimisation).

---

I have to ask: we are sure we want 🇺🇸 and not 🇬🇧 ? :)

rasbt:
Haha I probably don't have a say in this but I would definitely prefer the former ... Also, you don't want to change names of functions and modules, right? :)

betatim:
If recent 🇬🇧 behaviour is anything to go by we would be changing all the things 😃.

PR #124, opened 2016-07-20
[MRG] optimisation -> optimization

glouppe:
Fix for #121 

Note: I would wait maybe until tomorrow or the day after for merging this, as it is also renaming the notebooks, which means links that have been shared on twitter lately will break.


betatim:
🐝 zzzZzzZZ

+1 on waiting to merge, also easier to fix conflicts in this PR than in #123 or #119

betatim:
LGTM

PR #130, opened 2016-07-20
[MRG] add plot_convergence

glouppe:
This PR adds a convenience function for drawing one or several convergence traces. Notebooks have been updated (and simplified) accordingly.


betatim:
download 8
download 7
download 6

The three relevant plots.

skopt/__init__.py
```diff
@@ -1,6 +1,6 @@
 """
 Scikit-Optimize, or `skopt`, is a simple and efficient library
-for sequential model-based optimization, accessible to everybody
```
betatim: (any idea why we keep having whitespace at the end of the lines?)

skopt/plots.py
```diff
@@ -0,0 +1,67 @@
...
+    * `args[i]` [`OptimizeResult`, list of `OptimizeResult`, or tuple]:
+        The result(s) for which to plot the convergence trace.
+
+        - if `OptimizeResult`, then draw the corresponding single trace;
+        - if list of `OptimizeResult`, then draw the corresponding convergence
+          traces in light, along with the average convergence trace;
+        - if tuple, then `args[i][0]` should be a string label and `args[i][1]`
+          an `OptimizeResult` or a list of `OptimizeResult`.
+    """
+    colors = cm.viridis(np.linspace(0.25, 1.0, len(args)))
```
betatim: Nice! Didn't know this

betatim:
What do other libraries that create plots do in terms  of using `plt.plot` vs `ax.plot`? My intuition would be that we should create a new figure, plot on it, and return it. Will check `seaborn` and `pandas` to see what they do

glouppe:
> What do other libraries that create plots do in terms of using plt.plot vs ax.plot? My intuition would be that we should create a new figure, plot on it, and return it. Will check seaborn and pandas to see what they do

I could call `fig = plt.figure()` at the beginning and then return it without calling `plt.show()`? (but then it would be a bit less convenient to use)

betatim:
Seems like a lot of `seaborn` and `pandas` plotting stuff does:

```python
def plot_me(..., ax=None):
  if ax is None:
    ax = plt.gca()
  ...
  return ax
```

I'd just copy what they are doing in the absence of knowledge/experience with this subject.

betatim:
Legacy python 🐍 says hello via travis: `def f(*args, kw1=1, kw2=2)` is verboten :-/

skopt/__init__.py
glouppe: Nope :s

glouppe:
Thanks for the review, comments addressed!

betatim:
🍦🍦🍦

MechCoder:
Thanks! I will be using this today to compare the performance of the acquisition minimization techniques

PR #134, opened 2016-07-21
[MRG] Two fixes to Categorical dimension of space

nfcampos:
- support 2 classes
  - LabelBinarizer outputs only 1 column for the 2 classes case (see https://github.com/scikit-learn/scikit-learn/blob/51a765a/sklearn/preprocessing/label.py#L423)
- support `None` as a category
  - useful for eg. class_weight of RandomForestClassifier
  - requires transforming None into/from a string value before/after passing/receiving it from LabelBinarizer

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -43,8 +43,8 @@ class _CategoricalEncoder:
         """Convert labeled categories into one-hot encoded features."""
         self._lb = LabelBinarizer()
         NONE_ = '__None__'
-        self._Xin = np.vectorize(lambda a: a if a is not None else NONE_)
-        self._Xout = np.vectorize(lambda a: a if a is not NONE_ else None)
+        self.encode_none = np.vectorize(lambda a: a if a is not None else NONE_)
+        self.decode_none = np.vectorize(lambda a: a if a != NONE_ else None)
 
     def fit(self, X):
         """Fit a list or array of categories.
@@ -54,7 +54,7 @@ class _CategoricalEncoder:
         * `X` [array-like, shape=(n_categories,)]:
             List of categories.
         """
-        self._lb.fit(self._Xin(X))
+        self._lb.fit(self.encode_none(X))
         self.n_classes = len(self._lb.classes_)
 
         return self
@@ -72,7 +72,7 @@ class _CategoricalEncoder:
         * `Xt` [array-like, shape=(n_samples, n_categories)]:
             The one-hot encoded categories.
         """
-        return self._lb.transform(self._Xin(X))
+        return self._lb.transform(self.encode_none(X))
 
     def inverse_transform(self, Xt):
         """Inverse transform one-hot encoded categories back to their original
@@ -89,7 +89,7 @@ class _CategoricalEncoder:
             The original categories.
         """
         Xt = np.asarray(Xt)
-        return self._Xout(self._lb.inverse_transform(Xt))
+        return self.decode_none(self._lb.inverse_transform(Xt))
 
 
 class Dimension(object):
@@ -258,8 +258,8 @@ class Categorical(Dimension):
 
     @property
     def transformed_size(self):
-        l = len(self.categories)
-        return l if l != 2 else 1
+        size = len(self.categories)
+        return size if size != 2 else 1
 
     @property
     def bounds(self):
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -42,6 +42,9 @@ class _CategoricalEncoder:
     def __init__(self):
         """Convert labeled categories into one-hot encoded features."""
         self._lb = LabelBinarizer()
+        NONE_ = '__None__'
+        self._Xin = np.vectorize(lambda a: a if a is not None else NONE_)
+        self._Xout = np.vectorize(lambda a: a if a is not NONE_ else None)
 
     def fit(self, X):
         """Fit a list or array of categories.
@@ -51,7 +54,7 @@ class _CategoricalEncoder:
         * `X` [array-like, shape=(n_categories,)]:
             List of categories.
         """
-        self._lb.fit(X)
+        self._lb.fit(self._Xin(X))
         self.n_classes = len(self._lb.classes_)
 
         return self
@@ -69,7 +72,7 @@ class _CategoricalEncoder:
         * `Xt` [array-like, shape=(n_samples, n_categories)]:
             The one-hot encoded categories.
         """
-        return self._lb.transform(X)
+        return self._lb.transform(self._Xin(X))
 
     def inverse_transform(self, Xt):
         """Inverse transform one-hot encoded categories back to their original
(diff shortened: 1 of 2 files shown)


betatim:
Thanks for the PR!

Could you add a test that would have caught the two class counting bug? So it doesn't come back 🐛

nfcampos:
sure, do you want a test for `None` as a category as well?

skopt/space.py
```diff
@@ -255,7 +258,8 @@ def rvs(self, n_samples=None, random_state=None):
 
     @property
     def transformed_size(self):
-        return len(self.categories)
+        l = len(self.categories)
+        return l if l != 2 else 1
```
betatim: `l` -> `size` maybe? Right now it looks a bit like `return 1 if 1 != 2 else 1` :)

betatim:
Are there cases where you want to use `None` and numbers? It is much less clear in that case how to encode `None` but if someone has a bright idea we could do this in this PR. Not mandatory though.

skopt/space.py
nfcampos: yep ill change it to size

skopt/space.py
```diff
@@ -42,6 +42,9 @@ class _CategoricalEncoder:
     def __init__(self):
         """Convert labeled categories into one-hot encoded features."""
         self._lb = LabelBinarizer()
+        NONE_ = '__None__'
+        self._Xin = np.vectorize(lambda a: a if a is not None else NONE_)
+        self._Xout = np.vectorize(lambda a: a if a is not NONE_ else None)
```
betatim: rename it as `{en,de}code_none`? Something that is a bit more descriptive would be good.

betatim:
> sure, do you want a test for `None` as a category as well?

All the tests all the time (if you are offering) :)

nfcampos:
For optimizing hyper parameters of scikit classifiers actually what would be useful is a `Mixed` space that supports both numbers between `min` and `max` (like `Integer` or `Real`) as well as discrete categorical values (of which `None` could be one of them). but this would be in another PR

skopt/space.py
MechCoder: Doesn't storing lambda as attributes have issues while pickling?

nfcampos: yep, that's better i'll rename

nfcampos: @MechCoder what do you suggest? is making it a regular function enough?

MechCoder:
I believe you can do that with the Space API. no?
https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/tests/test_space.py#L129

nfcampos:
in that line the catgeories and the numbers are for different parameters right?

nfcampos:
I've addressed all feedback other than the `lambda` question, what do you guys suggest there?

betatim:
The problem is that you [can't pickle a lambda](https://gist.github.com/betatim/ae42c092827d2f8e9d6e4f7219a8cadd). If you make it a method instead it will pickle. Or should it be a helper outside the class @MechCoder?

nfcampos:
got it.
how about this?
```py
class CategoricalEncoder:
  @np.vectorize
  def encode_none(a):
    return a if a is not None else NONE_
```

skopt/space.py
```diff
@@ -42,6 +42,9 @@ class _CategoricalEncoder:
     def __init__(self):
         """Convert labeled categories into one-hot encoded features."""
         self._lb = LabelBinarizer()
+        NONE_ = '__None__'
```
glouppe: Do we really need that variable? could be directly inside the lambda

skopt/space.py
```diff
@@ -42,6 +42,9 @@ class _CategoricalEncoder:
...
+        self.encode_none = np.vectorize(lambda a: a if a is not None else NONE_)
+        self.decode_none = np.vectorize(lambda a: a if a != NONE_ else None)
```
glouppe: also there is not really a point in having those as attributes. Just define encode/decode  as regular functions, outside of the class

nfcampos: ok, I'll define them as functions outside

skopt/space.py
```diff
@@ -255,7 +258,8 @@ def rvs(self, n_samples=None, random_state=None):
 
     @property
     def transformed_size(self):
-        return len(self.categories)
+        size = len(self.categories)
+        return size if size != 2 else 1
```
glouppe: I dont understand. Surely you are adding a new `None` category right, which be counted for.

nfcampos: this is a completely unrelated fix. see https://github.com/scikit-learn/scikit-learn/blob/51a765a/sklearn/preprocessing/label.py#L423 scikit's `LabelBinarizer` outputs arrays of `size == len(categories` for all cases other than passing in exactly two categories to `fit` in which case it outputs a single binary vector

glouppe: If my categories are `["a", "b"]` why wouldnt the size be 2?

nfcampos: no, it'll be 1, see this line https://github.com/scikit-learn/scikit-learn/blob/51a765a/sklearn/preprocessing/label.py#L441

nfcampos: and compare to the example just above with 4

glouppe: Okay thanks. Please add an inline comment then, this is really not obvious without knowing.

nfcampos: yeah good call, I'll add it

skopt/tests/test_space.py
```diff
@@ -65,22 +65,40 @@ def test_integer():
...
 
+def test_categorical_transform_binary():
+    categories = ["apple", "orange"]
+    cat = Categorical(categories)
+
+    apple = [0.]
+    orange = [1.]
+
+    assert_equal(cat.transform(["apple"]).size, cat.transformed_size)
+    assert_array_equal(cat.transform(categories), [apple, orange])
```
MechCoder: We should change `Categorical` to handle this special case, such that this outputs `[[1.0, 0.0], [0.0, 1.0]]`. The reason `LabelBinarizer` handles it this way is because you can infer what the other column is given the output of `transform`.

nfcampos: why would we purposefully take up more memory when this is completely transparent to the user, is anyone supposed to directly use the `warped` space?

MechCoder:
OK. So there are two unrelated changes:

1. The first one is the behaviour of Categorical in the case of the two-category case, for example ["a", b"]. It is correct to return either `[[1.0, 0.0], [0.0, 1.0]]` or `[[1.0], [0.0]]` (current). But I propose to change the output to `[[1.0, 0.0], [0.0, 1.0]]` such as over here (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/logistic.py#L648) to avoid further confusion as to why the number of dimensions are different,

2. The second one is the usage of `None` as a category which might require a bit more discussion. You can easily encode None to be a category and de-encode it back to None inside your function. 

Would you be able to split the changes, so that it is easier to review? Thanks!

nfcampos:
sure

nfcampos:
I've opened the two PRs with these 2 fixes, you can close this one

PR #139, opened 2016-07-21
[WIP] Check return types/API conformity

betatim:
Start checking that we stick to the correct return types. Where have we got defined APIs?

`Space().rvs()` -> `list of lists`. The `*_minimize()` functions return a result object.

### DIFF
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -1,32 +1,71 @@
+from functools import partial
+
+import numpy as np
+
 from sklearn.utils.testing import assert_almost_equal
 from sklearn.utils.testing import assert_array_less
 from sklearn.utils.testing import assert_array_equal
 from sklearn.utils.testing import assert_equal
+from sklearn.utils.testing import assert_raises
 
 from skopt.benchmarks import branin
-from skopt.tests.test_tree_opt import MINIMIZERS as TREE_MINIMIZERS
+from skopt.dummy_opt import dummy_minimize
+from skopt.gp_opt import gp_minimize
+from skopt.tree_opt import forest_minimize
+from skopt.tree_opt import gbrt_minimize
 
-MINIMIZERS = []
-MINIMIZERS += TREE_MINIMIZERS
+
+# dummy_minimize does not support same parameters so
+# treated separately
+MINIMIZERS = (gp_minimize,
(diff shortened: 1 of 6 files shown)


MechCoder:
LGTM. Can you add a test for `inverse_transform` as well?

MechCoder:
Would you like to test the return type of the result objects in this PR as well? In other words before or after https://github.com/scikit-optimize/scikit-optimize/pull/119/ ?

MechCoder:
Here are the list of stuff, obtained from `gp_minmize`

`x` [float]: location of the minimum.
`fun` [float]: function value at the minimum.
 `models`: surrogate models used for each iteration.
 `x_iters` [list of lists]
`func_vals` [list]: function value for each iteration.
`[Space]`: the optimization space.

Minor: Should we relax the condition on `func_vals` ti be an ndarray instead of list?

betatim:
After #119

MechCoder:
ok, then I'll merge after the inverse_transform check. Thanks!

betatim:
make a common test out of the separate tests for trees and GPs (see #146)

skopt/tests/test_common.py
```diff
@@ -0,0 +1,61 @@
...
+from sklearn.utils.testing import assert_almost_equal
+from sklearn.utils.testing import assert_array_less
+from sklearn.utils.testing import assert_array_equal
+from sklearn.utils.testing import assert_equal
+from sklearn.utils.testing import assert_raises
+
+from skopt.benchmarks import branin
+from skopt.dummy_opt import dummy_minimize
+from skopt.gp_opt import gp_minimize
+from skopt.tests.test_tree_opt import MINIMIZERS as TREE_MINIMIZERS
```
betatim: How can I make this work in legacy python? Alternatively should there be `skopt.utils` with a `all_minimizers()`? Trying to find a nice way to not repeat `MINIMIZERS` from the tree tests.

glouppe: I would just repeat it for now. Then maybe we could factor this out in `skopt.testing` util.

betatim:
Anymore for anymore?

MechCoder:
thanks!

betatim:
Bim! 🏁

Happy weekend y'all!

PR #142, opened 2016-07-21
[MRG+1] Fix Categorical dimension not supporting exactly 2 categories

nfcampos:


MechCoder:
Thinking about it again, I'm fine with this behaviour.

skopt/tests/test_space.py
```diff
@@ -84,6 +84,22 @@ def test_categorical_transform():
...
 
 
+def test_categorical_transform_binary():
+    categories = ["apple", "orange"]
+    cat = Categorical(categories)
+
+    apple = [0.]
+    orange = [1.]
+
+    assert_equal(cat.transform(["apple"]).size, cat.transformed_size)
```
MechCoder: Can you just check this size explicilty, that is 

    `cat.transformed_size == 1`?

MechCoder:
looks ok to me. I'll wait for @glouppe or @betatim to merge.

skopt/tests/test_space.py
nfcampos: why? it is exactly when `transformed_size` differs from `transform(...).size` that you get bugs when this is eg. inside `Space`, which is how we hit this bug...

MechCoder: It is possible that `transformed_size` is the same as `transform(..).size` but both are wrong.

MechCoder: I should worded it as "Can you also just check this size explicitly"

nfcampos: ah, yes that makes sense. I've added it. went ahead and added the same size check for the non-binary case

glouppe:
Thanks! Merging this

betatim:
(For the historians: this supersedes #134)

PR #143, opened 2016-07-21
[MRG+1] Fix Categorical dimension not supporting None as a category

nfcampos:


nfcampos:
i guess accepting or not this PR comes down to whether you see the `Categorical` dimension as being for:
1) `space` entries of type `str`
2) or `space` entries that can take *discrete* values of multiple types.
Many optimisable functions take parameters of the second kind (eg. `sklearn.ensemble.RandomForestClassifier`, or just about any of the classifiers or regressors in `sklearn`) so it is definitely useful to a have a `dimension` like the one described in (2).

glouppe:
I think case 2 is an interesting extension, since it would allow to define a categorical dimension like `Categorical([None, 0, 1, 2, 3, "a", "b", "c", 10.0])` if someone really wants to do such thing. `sklearn.preprocessing.LabelEncoder` already supports that use case, with the exception of the `None` value. What do others think? 

(Let us agree first before reviewing the code)

betatim:
(For the historians: this supersedes #134)

skopt/space.py
```diff
@@ -51,7 +61,7 @@ def fit(self, X):
         * `X` [array-like, shape=(n_categories,)]:
             List of categories.
         """
-        self._lb.fit(X)
+        self._lb.fit(encode_none(X))
```
glouppe: I think the use of `np.vectorize` and these extra functions is a bit overkill. Also, I suspect `np.vectorize` may introduce subtle differences due to type coercion (e.g with `otypes="O"`). 

In legacy Python we indeed have:

```
In [1]: import numpy as np

In [2]: s = ["a", "b", None]

In [3]: f = np.vectorize(lambda v: v if v is not None else "__None__", otypes="U")

In [4]: g = np.vectorize(lambda v: v if v != "__None__" else None, otypes="O")

In [5]: f(s)
Out[5]: 
array([u'a', u'b', u'__None__'], 
      dtype='<U8')

In [6]: g(f(s))
Out[6]: array([u'a', u'b', None], dtype=object)

In [7]: g(f(s))[0]
Out[7]: u'a'

In [8]: type(g(f(s))[0])
Out[8]: unicode

In [9]: type(s[0])
Out[9]: str
```

Instead, you could directly iterate over the values of X:

```self._lb.fit([v if v is not None else "__None__" for v in X])```

File: skopt/space.py [hunk 2]
```diff
@@ -86,7 +96,7 @@ def inverse_transform(self, Xt):
             The original categories.
         """
         Xt = np.asarray(Xt)
-        return self._lb.inverse_transform(Xt)
+        return decode_none(self._lb.inverse_transform(Xt))
```
glouppe: and similarly:

```return [v if v != "__None__" else None for v in self._lb.inverse_transform(Xt)]```

File: skopt/space.py (hunk 1 above)
Comment (nfcampos) [inline skopt/space.py]: good point, should I make that a function, since it's used in two methods?

Comment (glouppe) [inline skopt/space.py]: If you want, but I am fine with duplicating this very small piece of code (a single line!). I find it easier to read.

**Comment by nfcampos** [comment]
@glouppe `Categorical([None, 0, 1, 2, 3, "a", "b", "c", 10.0])` in your comment above is currently impossible, even without the `None` category because the LabelEncoder raises `ValueError: Mix of label input types (string and number)`. What can we do about that?

**Comment by glouppe** [comment]
> @glouppe Categorical([None, 0, 1, 2, 3, "a", "b", "c", 10.0]) in your comment above is currently impossible, even without the None category because the LabelEncoder raises ValueError: Mix of label input types (string and number). What can we do about that?

Oh right. Well the easiest way would be to remap all 'symbolic' values (the ones provided in `categories`) to numbers (from 0 to len(categories)-1), feed that to `LabelEncoder`, and inverse map those integers to the original string values. Might be best to do it this way, since we wouldn't have to make an exception of `None`.

File: skopt/space.py [hunk 3]
```diff
@@ -10,6 +10,16 @@
...
 
 
+def encode_none(a):
+    return a if a is not None else '__None__'
+encode_none = np.vectorize(encode_none, otypes='U')
+
+
+def decode_none(a):
+    return a if a != '__None__' else None
+decode_none = np.vectorize(decode_none, otypes='O')
```
glouppe: Those can be removed now :)

nfcampos: ah yes :) in the meantime I've done the 'remap anything to numbers' approach you mentioned above, which should be better than this, so I'll push that

glouppe:
Very simple solution at the end! Thanks for your patience. That's +1 for me.

nfcampos:
@glouppe I believe there is one other bug with the `transformed_size` property of the `Categorical` dimension, if there are repeated categories `transformed_size` will report the size with the duplicates but `_CategoricalEncoder` removes duplicates, thus the sizes won't match. This already true before the changes in this PR because `LabelBinarizer` internally removes duplicate labels. This should probably be fixed in another PR right?

glouppe:
Yeah, i would not worry too much about this, for now we have been assuming that the categories are unique.

skopt/space.py
```diff
@@ -51,7 +51,9 @@ def fit(self, X):
         * `X` [array-like, shape=(n_categories,)]:
             List of categories.
         """
-        self._lb.fit(X)
+        self.mapping = {v: i for i, v in enumerate(X)}
```
MechCoder: How about just

```
python
self.mapping = dict(zip(range(len(X)) ,X))
```

MechCoder: Should we suffix this with an underscore?

MechCoder:
Just a minor issue about prefixing the `mapping` and `inverse_mapping` attributes with underscores. We may not want to expose these attributes to users. I'll merge after that.

glouppe:
Merging this, I'll make the changes with the underscores afterwards.

nfcampos:
Thanks for merging this!

glouppe:
Thanks for your contribution Nuno :)

betatim:
As it is still summer where I am: 🍨 for successful PRs!

PR #144, opened 2016-07-22
[MRG] Update number of random points to be sampled to optimize the acquisition function.

MechCoder:
Here we go (on hart6):

Before:
index

hart6


MechCoder:
Tests fail. This is non-intuitive to me :(

glouppe:
Maybe we can slightly adjust the allowed margin error of those tests? 

But the main reason i think is that optimization gets stucked in a local minima and never escapes from it because of a lack of exploration (which is partly due to how the returned std behaves in regions that have not been seen by the forest).

MechCoder:
I'll have a look at the tests tomorrow.

glouppe:
Set to `n_points=1000` as part of #174. Closing.

PR #146, opened 2016-07-22
Return type should be a list of lists

MechCoder:


glouppe:
Thanks for this! +1 modulo Travis' veto

MechCoder:
We should probably refactor most of the tests in `tests_common.py` similar to sklearn's. But among other things, refactoring is not my idea of fun :p

glouppe:
Sure, we can do that later though.

betatim:
@MechCoder you should join me for the party over in #139 🎉

Issue #67, opened 2016-04-22
Plot functions from `OptimizeResult` object


glouppe:
To avoid cluttering the examples with tons of code to produce nice plots, I think it would make sense to provide a few plot functions taking directly an `OptimizeResult` object. 

See e.g. the `plot_acquisition` and `plot_convergence` functions in GPyOpt (at the end of http://diana-hep.org/carl/notebooks/Parameterized%20inference%20with%20nuisance%20parameters.html)

betatim:
In addition to the GPyOpt plots it would be good to have something to show where the samples are for N>2 dimensions. Maybe a [corner.py](https://github.com/dfm/corner.py) style plot.

glouppe:
I'll work on `plot_convergence`, this would be convenient for all our notebooks. I am thinking about designing it so that `plot_convergence` can either accept a single `OptimizeResult` object, a list of them (hence plotting the average and the individual traces), or a list of lists of them.

glouppe:
What could be another interesting plot? in GPyOpt they have this plot which shows the distance between consecutive points. This is nice to study the exploitation/exploration trade-off, no matter the dimension of the search space. The caveat is that not all distance metrics are appropriate...

betatim:
Listing some ideas:

* https://github.com/scikit-optimize/scikit-optimize/issues/67#issuecomment-232316611 I'd love this or other tools to help me see if the optimiser got stuck/is doing the right thing.
* I also found the 2D plot from @kegl useful as well as the 1D-convex-hull one. [plots towards bottom of the ipynb](https://gist.github.com/betatim/41d2c4375d1869d03ded76578e6c25ab)

glouppe:
Closing this, we have some nice ones now :)

PR #147, opened 2016-07-22
[MRG] Scatter plot matrix

betatim:
Create a scatter plot from a `OptimizeResult`.

Do you know how to make the top triangle vanish or something useful we could display there?

This is the output from a run on `hart6`:
download 7

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -92,12 +92,43 @@ def plot_convergence(*args, **kwargs):
     return ax
 
 
-def plot_scatter_matrix(result, bins=20):
-    """Plot scatter plot matrix of samples.
+def _format_scatter_plot_axes(ax, space):
+    # Deal with formatting of the axes
+    for i in range(space.n_dims):
+        for j in range(space.n_dims):
+            ax_ = ax[i, j]
 
-    The diagonal shows a histogram for each dimension. Create pairwise
-    scatter plots of each dimension of the search space. Later
-    samples are shown in a darker color.
+            if j > i:
+                ax_.axis("off")
+
+            # adjust bounds for every off-diagonal axis
+            if i != j:
+                ax[i, j].set_ylim(*space.dimensions[i].bounds)
+                ax[i, j].set_xlim(*space.dimensions[j].bounds)
+
+            ax_.xaxis.set_major_locator(MaxNLocator(6, prune='both'))
+            ax_.yaxis.set_major_locator(MaxNLocator(6, prune='both'))
+
+            if i < space.n_dims - 1:
+                ax_.set_xticklabels([])
+            # bottom row
+            else:
+                [l.set_rotation(45) for l in ax_.get_xticklabels()]
+
+            if j > 0:
+                ax_.set_yticklabels([])
+
+    return ax
+
+
+def plot_objective_function(result, levels=10):
+    """Pairwise scatter plot of objective function
+
+    Pairwise scatter plots are shown on the off-diagonal for each
+    dimension of the search space. A red point indicates the minimum.
+
+    Note: the objective function contours are obtained by interpolating
+    between samples. The surrogate model is not used.
 
     Note: search spaces that contain `Categorical` dimensions are
     currently not supported by this function.
@@ -105,7 +136,62 @@ def plot_scatter_matrix(result, bins=20):
     Parameters
     ----------
     * `result` [`OptimizeResult`]
-        The result for which to plot the scatter plot matrix.
+        The result for which to create the scatter plot matrix.
+
+    * `levels` [int, default=10]
+        Number of levels to draw on the contour plot, passed directly
+        to `plt.contour()`
+
+    Returns
+    -------
+    * `ax`: [`Axes`]:
+        The matplotlib axes.
+    """
+    space = result.space
+    samples = np.asarray(result.x_iters)
+    order = range(samples.shape[0])
+    fig, ax = plt.subplots(space.n_dims, space.n_dims, figsize=(8, 8))
+
+    fig.subplots_adjust(left=0.05, right=0.95, bottom=0.05, top=0.95,
+                        hspace=0.1, wspace=0.1)
+
+    for i in range(result.space.n_dims):
+        for j in range(result.space.n_dims):
+            # lower triangle
+            if i > j:
+                # define grid
+                # XXX use linspace(*args, 100) after python2 support ends
+                bounds = space.dimensions[j].bounds
+                xi = np.linspace(bounds[0], bounds[1], 100)
+                bounds = space.dimensions[i].bounds
+                yi = np.linspace(bounds[0], bounds[1], 100)
+                zi = griddata((samples[:, j], samples[:, i]), result.func_vals,
+                              (xi[None, :], yi[:, None]), method='cubic')
+
+                ax[i, j].contour(xi, yi, zi, levels, linewidths=0.5, colors='k')
+                ax[i, j].contourf(xi, yi, zi, levels, cmap='viridis_r')
+                ax[i, j].scatter(samples[:, j], samples[:, i], c='k', s=10, lw=0.)
+                ax[i, j].scatter(result.x[j], result.x[i], c=['r'], s=20, lw=0.)
+
+    return _format_scatter_plot_axes(ax, space)
+
+
+def plot_sampling_order(result, bins=20):
+    """Visualize order in which points where sampled
+
+    Pairwise scatter plots are shown on the off-diagonal for each
+    dimension of the search space. The order in which samples were
+    evaluated is as the colour of each point. The diagonal shows a
+    histogram of sampled values for each dimension. A red point
+    indicates the minimum.
+
+    Note: search spaces that contain `Categorical` dimensions are
+    currently not supported by this function.
+
+    Parameters
+    ----------
+    * `result` [`OptimizeResult`]
+        The result for which to create the scatter plot matrix.
 
     * `bins` [int, bins=20]:
         Number of bins to use for histograms on the diagonal.
@@ -118,9 +204,7 @@ def plot_scatter_matrix(result, bins=20):
     space = result.space
     samples = np.asarray(result.x_iters)
     order = range(samples.shape[0])
-    fig, ax = plt.subplots(result.space.n_dims,
-                           result.space.n_dims,
-                           figsize=(8, 8))
+    fig, ax = plt.subplots(space.n_dims, space.n_dims, figsize=(8, 8))
 
     fig.subplots_adjust(left=0.05, right=0.95, bottom=0.05, top=0.95,
                         hspace=0.1, wspace=0.1)
@@ -133,44 +217,8 @@ def plot_scatter_matrix(result, bins=20):
             # lower triangle
             elif i > j:
                 ax[i, j].scatter(samples[:, j], samples[:, i], c=order,
-                                 s=40, lw=0., alpha=0.996, cmap='viridis')
+                                 s=40, lw=0., cmap='viridis')
                 ax[i, j].scatter(result.x[j], result.x[i],
-                                 c=['r'], s=10, lw=0.)
+                                 c=['r'], s=20, lw=0.)
 
-            # upper triangle
-            elif i < j:
-                # define grid.
-                xi = np.linspace(*space.dimensions[j].bounds, 100)
-                yi = np.linspace(*space.dimensions[i].bounds, 100)
-                zi = griddata((samples[:, j], samples[:, i]), result.func_vals,
-                              (xi[None, :], yi[:, None]), method='cubic')
-
-                ax[i, j].contour(xi, yi, zi, 15, linewidths=0.5, colors='k')
-                ax[i, j].contourf(xi, yi, zi, 15, cmap='viridis')
-                ax[i, j].scatter(samples[:, j], samples[:, i], c='k', s=10, lw=0.)
-                ax[i, j].scatter(result.x[j], result.x[i], c=['r'], s=10, lw=0.)
-
-    # Deal with formatting axes
-    for i in range(result.space.n_dims):
-        for j in range(result.space.n_dims):
-            ax_ = ax[i, j]
-
-            # adjust bounds for every off-diagonal axis
-            if i != j:
-                ax[i, j].set_ylim(*space.dimensions[i].bounds)
-                ax[i, j].set_xlim(*space.dimensions[j].bounds)
-
-            ax_.xaxis.set_major_locator(MaxNLocator(6, prune='both'))
-            ax_.yaxis.set_major_locator(MaxNLocator(6, prune='both'))
-
-            if i < result.space.n_dims - 1:
-                ax_.set_xticklabels([])
-            # bottom row
-            else:
-                [l.set_rotation(45) for l in ax_.get_xticklabels()]
-
-            # first column
-            if j > 0:
-                ax_.set_yticklabels([])
-
-    return ax
+    return _format_scatter_plot_axes(ax, space)
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -117,7 +117,7 @@ def _format_scatter_plot_axes(ax, space, ylabel):
                 if j > 0:
                     ax_.set_yticklabels([])
                 else:
-                    ax_.set_ylabel("$X_{%i,%i}$" % (i, j))
+                    ax_.set_ylabel("$X_{%i}$" % i)
             else:
                 ax_.set_ylim(*diagonal_ylim)
                 ax_.yaxis.tick_right()
@@ -134,7 +134,7 @@ def _format_scatter_plot_axes(ax, space, ylabel):
             # ... the bottom row
             else:
                 [l.set_rotation(45) for l in ax_.get_xticklabels()]
-                ax_.set_xlabel("$X_{%i,%i}$" % (i, j))
+                ax_.set_xlabel("$X_{%i}$" % j)
 
     return ax
 
@@ -233,7 +233,7 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
         return xi, yi, zi
 
 
-def plot_objective_function(result, levels=10, n_points=40, n_samples=100):
+def plot_objective(result, levels=10, n_points=40, n_samples=100):
     """Pairwise partial dependence plot of the objective function.
 
     The diagonal shows the partial dependence for dimension `i` with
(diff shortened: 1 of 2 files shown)


glouppe:
I think you need to call `ax.axis("off")` for the corresponding axes.

glouppe:
That's nice! Would be good to plot the min in red or something, what do you think?

betatim:
With all these `i` and `j`s could you check that I didn't screw up any indexing please?

```
from functools import partial
from skopt.benchmarks import hart6 as _hart6
from skopt import plots

def hart6(x, noise_level=0.):
    return _hart6(x) + noise_level * np.random.randn()

bounds = [(0., 1.), (0., 1.), (0., 1.), (0., 1.), (0., 1.), (0., 1.)]
func = partial(hart6, noise_level=2.0)
n_calls = 80*2
et_minimize = partial(forest_minimize, base_estimator="et")
et_res = et_minimize(func, bounds, n_calls=n_calls, random_state=1)
ax = plots.plot_scatter_matrix(et_res, bins=10)
```

download 9

betatim:
Interesting to see how much these distributions change just by rerunning things 😕 😟

betatim:
same problem, but with `dummy_minimize`:

download 10

betatim:
One more plot for #67

skopt/plots.py
```diff
@@ -88,3 +89,65 @@ def plot_convergence(*args, **kwargs):
...
+
+    * `bins` [int, bins=20]:
+        Number of bins to use for histograms on the diagonal.
+
+    Returns
+    -------
+    * `ax`: [`Axes`]:
+        The matplotlib axes.
+    """
+    samples = np.asarray(result.x_iters)
```
glouppe: Shall we document that this function cannot be used for `Categorical` dimensions?

skopt/plots.py
```diff
@@ -88,3 +89,65 @@ def plot_convergence(*args, **kwargs):
...
+    fig.subplots_adjust(left=0.05, right=0.95, bottom=0.05, top=0.95,
+                        hspace=0.1, wspace=0.1)
+
+    for i in range(result.space.n_dims):
+        for j in range(i + 1):
+            if i == j:
+                ax[i, i].hist(samples[:, i], bins=bins)
+            else:
+                ax[i, j].scatter(samples[:, j], samples[:, i],
+                                 c=order, s=40, lw=0., alpha=0.6, cmap='Blues')
```
glouppe: Regarding `c=order`, do you think it is that useful to color-code the points by their iteration number? I think it is interesting to have a plot somehow describing in time the optimization process, but in this case, wouldn't it make more sense if those points were instead color-coded by their corresponding objective value?

skopt/plots.py
```diff
@@ -88,3 +89,65 @@ def plot_convergence(*args, **kwargs):
...
+    fig, ax = plt.subplots(result.space.n_dims,
+                           result.space.n_dims,
+                           figsize=(8, 8))
+    fig.subplots_adjust(left=0.05, right=0.95, bottom=0.05, top=0.95,
+                        hspace=0.1, wspace=0.1)
+
+    for i in range(result.space.n_dims):
+        for j in range(i + 1):
+            if i == j:
+                ax[i, i].hist(samples[:, i], bins=bins)
```
glouppe: Then if points correspond to their objective value, a better plot on the diagonal might be the partial dependence plots of those dimensions? what do you think?

glouppe:
> With all these i and js could you check that I didn't screw up any indexing please?

Looks good

skopt/plots.py
betatim: It looks pretty but for any large number of samples I think you end up with "points on top of points on top of points" so the utility of using `c=order` diminishes.

I like the idea of colour coding by objective function. Should we add it (upper triangle) or replace or make it a parameter? I will try the upper triangle version unless there is opposition.

glouppe: Yup, give it a try. Not sure that will make the plot easy to read though.

skopt/plots.py
betatim: Independent of the colouring of the off-diagonals I think this is a good idea. I would add it as a line ontop of the histogram. I'd like to be able to see where the samples are (the histogram) and what the value of the objective function is (line). To verify that most samples are where the suspected minimum is.

Should we make a new function which shows on the diagonal the 1D partial dependence and on the off-diagonal the 2D?

betatim: If possible, can we make these in two new PRs? One for adding partial dependence on the diagonal to `plot_scatter_matrix` and a second one to make a matrix that shows the 1D and 2D partial dependence

glouppe: I am really not sure we should mix these two sources of information, i.e. have instead
- a function for the scatter plots color-coded by iteration number + histograms of those locations on the diagonal. This is the function we have in this PR. 
- a function for the scatter plots color-coded by objective value + partial dependence plots on the diagonal.

If we go that way, then `plot_scatter_matrix` should be renamed to be more descriptive.

betatim:
download 11

The lower triangle shows the same as before but different cmap. Upper triangle shows where samples are and the objective function.

(plus several bug fixes and tweaks to handle axis ranges better)

Conclusion from my side: [do what Gilles said](https://github.com/scikit-optimize/scikit-optimize/pull/147#discussion_r72048730)

betatim:
Split things into two functions. One to show the order of samples and histograms, and another to show the objective function value. Below the output of each on `branin` with two extra dimensions (that are meaningless).

`plot_sampling_order`
download 12

`plot_objective_function`
download 13

Thoughts on the names?

What is missing is how to calculate the partial dependence for the diagonal.

Should we introduce a switch that allows using the model instead of the samples for `plot_objective_function` or should it be a new function?

glouppe:
Let `F(X=(X_1, ..., X_p))` be the prediction of our model. For the 1d partial depenpence of plot of feature `X_i`, we want to plot `f(x) = E[F(X)|X_i=x]` where the expectation is taken over the joint distribution of all the other features `j != i`. 

In our case, there is no distribution defined over those features (it is in fact a search space, not  a distribution), so partial dependence plots are not really defined either. However, we could still  artificially assume that these feature are uniformly distributed and  plot `f(x) = E[F(X)|X_i=x]` as empirically approximated using the mean prediction of the last model over uniformly randomly drawn points (i.e `space.rvs()`) where the value of feature `X_i` is changed to `x`.

This also extends to 2d plots where two feature values are fixed instead of one.

betatim:
I worked on this a bit [here](https://gist.github.com/betatim/52905c4d01a9185c81de2b0afd72b05e). Seems like "marginal" is the wrong word. This computes 1d versions of a 2d function using the samples.

(I now wonder if you have access to lhcb.slack.com and watched the discussion there.)

glouppe:
Hmm, I am not sure whether we should either reuse `x_iters` as the data we marginalize or instead sample uniformly at random within the bounds. After all, maybe it makes sense to reuse `x_iters`, since this is the data that was used for training the model.

betatim:
I think using the model is a better idea. The whole idea of SMBO is that the model adds something to the data points in terms of getting a good estimate of the true shape of the function you are optimising. It also has the advantage that we can generate more samples from the model than we can from the true objective function.

(Practically I wonder if we could see much difference because the plots are fairly zoomed out.)

I will implement (essentially):
```python

def partial_dependence(dimension, bins=10):
    edges = mquantiles(X[:, dimension], prob=np.linspace(0, 1, bins, endpoint=False)[1:])
    edges = np.array([low] + list(edges) + [high])

    q_x = []
    q_y = []
    q_up = []
    q_down = []
    for i in range(edges.shape[0] - 1):
        idx = (edges[i] < X[:, dimension]) & (X[:, dimension]< edges[i+1])
        q_x.append(edges[i] + (edges[i+1] - edges[i])/2.)
        # average it! no median
        q_y.append(np.median(ys[idx]))
        q_down.append(np.percentile(ys[idx], 16))
        q_up.append(np.percentile(ys[idx], 84))

    plt.plot(X[:, dimension], ys, '.')
    plt.plot(q_x, q_y, 'r-', lw=2)
    plt.fill_between(q_x, q_down, q_up, color='r', alpha=0.3)
```

Adding the ability to do 2d and using samples from the model.

>  approximated using the mean prediction of the last model over uniformly randomly drawn points (i.e `space.rvs()`) where the value of feature `X_i` is changed to `x`.

This isn't quite clear to me. In the [notebook](https://gist.github.com/betatim/52905c4d01a9185c81de2b0afd72b05e) I draw random samples, then average them per quantile in `X_i`. Are you suggesting to draw random samples and then set the value of `X_i` to a certain value for each sample? I don't think you can do that 😕

glouppe:
> This isn't quite clear to me. In the notebook I draw random samples, then average them per quantile in X_i. Are you suggesting to draw random samples and then set the value of X_i to a certain value for each sample? I don't think you can do that 

Why not? This is what a partial dependence plot is.

betatim:
Oh wow 😮 ! I thought I had implemented the partial dependence plot, but no. Now that I understand (I think):

```python
def f(x):
    return np.sin(x[0])*x[0] + np.cos(x[1])*x[1] + x[2]

def partial(dimension):
    # dimension we want to see
    xx = np.linspace(low, high, 100)

    rvs = np.random.uniform(low, high, size=(100, 3))
    
    y = []
    for x in xx:
        rvs_ = rvs.copy()
        rvs_[:, dimension] = x
        y.append(np.mean([f(r) for r in rvs_]))

    return xx, y

plt.plot(*partial(0), 'b-')
plt.plot(*partial(1), 'r-')
plt.plot(*partial(2), 'g-')
```
gives
download 19

📔 📚 every day is a school day with @glouppe! Merci.

glouppe:
Yeah, that is it :) 

> rvs = np.random.uniform(low, high, size=(100, 3))

Now my mumbling above might make more sense to you: I was wondering whether we should marginilize over uniformly random points, as you do here, or rather use the points collected in `res.x_iters` (and evaluate them through `res.models[-1]`).

betatim:
download 20

This is what you get as partial dependence plot for a ET model applied to 
```python
def f(x):
    return np.sin(x[0])*x[0] + np.cos(x[1])*x[1]
# one spurious/random/useless dimension
bounds = [(-3.141, 3*3.141), (-3.141, 3*3.141), (-10., -4.)]

n_calls = 160
n_random_starts = 15
```

It is nice to see that the model realises that the third dimension is a red herring. Note that the plot in column 1, row 2 only roughly looks like the real objective function. This confused me for a while but it makes sense because you only evaluate the expensive function at points close to the minimum -> the model doesn't learn anything about how the function looks like far away. If you increase the number of random samples to 150, thing start to look a lot more like the real objective function:
download 21

A big question for me is how to write good/useful tests for this. Right now I am "debugging"/checking things as I go along by thinking hard about each plot and comparing things with little toy implementations outside of `skopt`. For example [this notebook](https://gist.github.com/betatim/3cec4f7fc884538ebcf38a1a014f446c) which is only really useful for me today. Ideas welcome.

skopt/plots.py
```diff
@@ -88,3 +90,157 @@ def plot_convergence(*args, **kwargs):
...
+
+    Returns
+    -------
+    * `ax`: [`Axes`]:
+        The matplotlib axes.
+    """
+    space = result.space
+    samples = np.asarray(result.x_iters)
+    order = range(samples.shape[0])
+    rvs = space.rvs(n_samples=10)
```
glouppe: I think you can increase this to `100` or `500`. `n_samples=10` might lead to too noisy empirical estimates.

glouppe: also, have tried to compare visually replacing this with `rvs = results.x_iters`?

skopt/plots.py
```diff
@@ -88,3 +90,157 @@ def plot_convergence(*args, **kwargs):
...
+    for i in range(result.space.n_dims):
+        for j in range(result.space.n_dims):
+            if i == j:
+                bounds = space.dimensions[i].bounds
+                xi = np.linspace(bounds[0], bounds[1], 40)
+                values = []
+
+                for x in xi:
+                    rvs_ = np.array(rvs)
+                    rvs_[:, i] = x
```
glouppe: `rvs` should be `space.transform`'d (even when using only int or floats, since we have this optional prior argument)

skopt/plots.py
betatim: Right now I tuned the numbers so that they aren't super painful when you want to plot often. I had started with `100` points but this was super slow. +1 for 100 as default and we should definitely make it a parameter that can be adjusted by the user.

glouppe: Yeah, I guess this should be tuned in combination with the number of steps from lower to upper bounds (set to 40 at the moment).

betatim:
Another iteration of 🎨. Getting there. The following four pictures are `plot_sampling_order` and `plot_objective_function` for a ET model that has few and many `n_random_starts`.

The language in the docstrings could be improved. Not sure how to refer to partial dependence. Are you calculating the partial dependence of the model wrt dim1 and dim2? All reads a bit clunky 😢 

download 23
download 22
download 25

glouppe:
Giving detailed explanations about those plots and how to read would make for a great notebook :)

glouppe:
Could add labels on the left and bottom, to make it easier that a subplot concerns X_i vs X_j?

skopt/plots.py
```diff
@@ -88,3 +90,245 @@ def plot_convergence(*args, **kwargs):
...
+            # bottom row
+            else:
+                [l.set_rotation(45) for l in ax_.get_xticklabels()]
+
+    return ax
+
+
+def partial_dependence(space, model, i, j=None, sample_points=None,
+                       n_samples=100, n_points=40):
+    """Calculate partial dependence of `model` for dimensions `i` and `j`
```
glouppe: Maybe instead something like "Calculate the partial dependence for dimensions `i` and `j` with respect to the objective value".

skopt/plots.py
```diff
@@ -88,3 +90,245 @@ def plot_convergence(*args, **kwargs):
...
+
+    The idea is that the partial dependence plot tells us how the
+    value of the dimensions `i` and `j` influence the `model` predictions
+    after we have "averaged out" the influence of all other dimensions.
+
+    Parameters
+    ----------
+    * `space` [`Space`]
```
glouppe: Can you add colons after the brackets? (here and everywhere)

skopt/plots.py
```diff
@@ -88,3 +90,245 @@ def plot_convergence(*args, **kwargs):
...
+        Surrogate model for the objective function.
+
+    * `i` [int]
+        The first dimension for which to calculate the partial dependence.
+
+    * `j` [int, default=None]
+        The second dimension for which to calculate the partial dependence.
+        To calculate the 1D partial dependence on `i` alone set `j=None`.
+
+    * `sample_points` [list of lists, shape=(n_points, n_dims), default=None]
```
glouppe: If those are transformed, we should require an np.array instead.

skopt/plots.py
```diff
@@ -88,3 +90,245 @@ def plot_convergence(*args, **kwargs):
...
+    * `yi`: [np.array]:
+        The value of the model at each point `xi`.
+
+    For 2D partial dependence:
+
+    * `xi`: [np.array]:
+        The points at which the partial dependence was evaluated.
+    * `yi`: [np.array]:
+        The points at which the partial dependence was evaluated.
+    * `zi`: [np.array]:
```
glouppe: Would be good to document the shape, as `zi` is 2d array.

skopt/plots.py
```diff
@@ -88,3 +90,245 @@ def plot_convergence(*args, **kwargs):
...
+
+
+def plot_objective_function(result, levels=10, n_points=40, n_samples=100):
+    """Pairwise scatter plot of objective function
+
+    Pairwise scatter plots are shown on the off-diagonal for each
+    dimension of the search space. A red point indicates the minimum.
+
+    Note: the objective function contours are obtained by interpolating
+          between samples. The surrogate model is not used.
```
glouppe: This is no longer true, right?

glouppe:
For the plots on the diagonal, I think it would be better if all were using the y-scale for the number of samples / objective value. E.g. for the 1d partial dependence plot of X3, it is not obvious that there is almost no correlation, simply because the plot is zoomed in.

skopt/plots.py
betatim: Correct, got removed.

skopt/plots.py
betatim: Done. Actually it is a list of lists. I think this is ok (but easy to turn into an array if that's what the world wants).

skopt/plots.py
betatim: Done.

betatim:
> Giving detailed explanations about those plots and how to read would make for a great notebook :)

Easy, you read them just like you read tealeaves ;) 🍵 

Plots below are `branin` without noise and two extra dimensions that are useless. ET model, with 15 random points.

download 27
download 26

glouppe:
Looks great! Just one last nitpick about the x and y-axis labels. Wouldnt it be easier if those were simply `X_i` (on the y-axis?) and `X_j` (on the other axis). E.g, at the moment it is immediate what `X_3,2` means for the x-axis vs `X_3,0` on the y-axis.

skopt/plots.py
```diff
@@ -88,3 +90,267 @@ def plot_convergence(*args, **kwargs):
...
+            for y_ in yi_transformed:
+                rvs_ = np.array(sample_points)
+                rvs_[:, (i, j)] = (x_, y_)
+                row.append(np.mean(model.predict(rvs_)))
+            zi.append(row)
+
+        return xi, yi, zi
+
+
+def plot_objective_function(result, levels=10, n_points=40, n_samples=100):
```
glouppe: Or simply, `plot_objective`?

skopt/plots.py
```diff
@@ -88,3 +90,267 @@ def plot_convergence(*args, **kwargs):
...
+
+                ax[i, j].contour(xi, yi, zi, levels, linewidths=0.5, colors='k')
+                ax[i, j].contourf(xi, yi, zi, levels, cmap='viridis_r')
+                ax[i, j].scatter(samples[:, j], samples[:, i], c='k', s=10, lw=0.)
+                ax[i, j].scatter(result.x[j], result.x[i], c=['r'], s=20, lw=0.)
+
+    return _format_scatter_plot_axes(ax, space, "Partial dependence")
+
+
+def plot_sampling_order(result, bins=20):
```
glouppe: `plot_evaluations`?

glouppe:
Shall we also update one of the notebooks to include of one those plots?

betatim:
> Shall we also update one of the notebooks to include of one those plots?

betatim:
(I just wanted to cancel my comment ... #fatfingers)

betatim:
Will investigate including it in an existing notebook. #162 is about creating notebooks that explain how to read these plots in general. We should have those too.

glouppe:
That can also be done later.

For me this PR is already good enough to be merged.

betatim:
Then I vote merge now, would be great to get this done before I leave for holiday 🌴

glouppe:
Boum!

betatim:
😃 Thanks for the comments and patience, this PR just kept getting bigger and bigger.

PR #155, opened 2016-07-26
List packages for setup.py

### LINKED ISSUE #152: Setup.py install does not work
Not sure why, but locally `python setup.py install` never seems to work. The `skopt` module can never found afterwards. By contrast  `python setup.py develop` does work.

Can anybody try to reproduce this in some test env?

betatim:
Fixes #152 

Added a few more fields as well. Who is the author of skopt? What about "The scikit-optimize contributors"? Should we make an email address (what doe sklearn do for this?)?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,4 +12,5 @@ setup(name='skopt',
                         ' contexts.'),
       url='https://scikit-optimize.github.io/',
       license='BSD',
+      author='The scikit-optimize contributors',
       packages=['skopt', 'skopt.learning'])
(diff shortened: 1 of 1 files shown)


betatim:
Explicitly listing the packages like this means we avoid having to deal with `find_packages` not being available in `distutils`. If one day we have lots of packages and get bored of listing them we can switch.

glouppe:
> What about "The scikit-optimize contributors"?

+1

glouppe:
and +1 for this

betatim:
I'll merge once travis is happy.

PR #156, opened 2016-07-26
fix Categorical dimension transformed_bounds for len(categories) == 2

nfcampos:
related to #142 
to be compatible with `Space.transformed_bounds`, Categorical.transformed_bounds` needs to output a single tuple when `transformed_size` is 1, instead of a list with 1 tuple.


glouppe:
Good catch! Thanks

Issue #131, opened 2016-07-21
Add call specifications to results


glouppe:
For inspection purposes of the results (e.g. by a plotting function), I think it would be nice to record in the `OptimizeResult` object the arguments that were passed to the `_minimize` function (e.g. by copying `locals()` at the beginning of the call).

PR #157, opened 2016-07-27
[MRG] Save call specifications in results

glouppe:
My attempt at #131 

This adds a `specs` dictionary to `results`, such that `results.specs["function"]` and `results.specs["args"]` respectively store the function name and a copy of its arguments.


betatim:
Looks sensible.

To restart the optimisation process from this information, should we also store the random state at the end of the function? Or should we leave that as something each `*_minimize` to implement if they want to support warm starts?

betatim:
I am assuming that warm starting is one of the primary uses cases for having this.

glouppe:
> To restart the optimisation process from this information, should we also store the random state at the end of the function? 

Just did that

glouppe:
> I am assuming that warm starting is one of the primary uses cases for having this.

Yup, also:
- self documenting results
- can be used for introspection by plotting functions (e.g. to recover what was the acquisition function and make those plots)

betatim:
For the `rng` state, does it make a difference whether we persist it at the start or the end of the call? Actually not sure which one you'd want:

* the state as passed to the minimizer?
* the state after the minimizer has made all the calls to it?

or does it make no difference because we don't get a copy of the RNG state so it keeps updating and we get the state at the end of the minimizer?

glouppe:
The original value as it was passed (e.g. an int) is stored with the other arguments, i.e. in `specs["args"]["random_state"]`. Then what I do in addition is to store a reference to (not a copy of) the `RandomState` object, whose state gets updated by the subsequent instructions.

betatim:
Ok. Sounds like we have all the bases covered. Waiting for travis and then merge.

(We need more docs that explain to people how to use all the stuff we have built :-/)

betatim:
Quick and efficient, like a 🚄.

MechCoder:
Sorry for being late, could you clarify a bit more on why this is useful? (possibly with a minor example) I can see the usefulness of storing the state of the random state before and after, but it is not completely obvious as to why we need to store all the other arguments passed to the function. (Since none of them are modified in-place)

MechCoder:
Oh I see, Is it just to plot stuff etc directly from the `OptimizeResult` instance without having to rely on the original functions? Nvm, I was just afraid that we are overloading the `OptimizeResult` instance with too much stuff ;)

Issue #160, opened 2016-07-28
Normalization to [0,1]?


glouppe:
For a project I am working on, I just found that normalizing input values to [0,1] yields better results. 

In general, I think this is not necessary, as the GP kernel should be able to learn the proper scaling, but I do see differences in practice. 

We shall investigate more systematically on few examples whether this is necessary and either i) pick a better default kernel or ii) systematically rescale so that the transformed space is [0,1]^p, no matter what. In a very early version of the codebase, we were indeed doing that.

Also, option ii has some advantages, as it allows not shooting oneself in the foot when playing with custom `base_estimator`.

betatim:
Indeed this got lost when fiddling with the spaces API.

Are there any downsides to always transforming? It seems like it should be ok. Just wondering for tree based models which don't have a problem with varying scales if there is a penalty they end up paying for getting normalised dimensions only.

MechCoder:
Should we postpone this to after 0.1?

glouppe:
Yes, this is just an exploratory branch at the moment

PR #161, opened 2016-07-28
[WIP] Normalize transformed space to [0,1]^p

glouppe:
This is an exploratory branch for #160

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -23,6 +23,23 @@ class _Identity:
         return Xt
 
 
+class _Normalize:
+    """Normalize transform."""
+
+    def __init__(self, lower, upper):
+        self.lower = float(lower)
+        self.upper = float(upper)
+
+    def fit(self, X):
+        return self
+
+    def transform(self, X):
+        return (np.asarray(X, dtype=np.float) - self.lower) / (self.upper - self.lower)
+
+    def inverse_transform(self, Xt):
+        return Xt * (self.upper - self.lower) + self.lower
+
+
 class _Log10:
     """Base 10 logarithm transform."""
 
@@ -110,7 +127,7 @@ class Dimension(object):
         """
         rng = check_random_state(random_state)
         samples = self._rvs.rvs(size=n_samples, random_state=rng)
-        return self.inverse_transform(samples)
+        return samples
 
     def transform(self, X):
         """Transform samples form the original space to a warped space."""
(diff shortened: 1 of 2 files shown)


glouppe:
For further reference, Spearmint uses Beta warping by default. See https://github.com/HIPS/Spearmint/blob/master/spearmint/models/gp.py#L319

MechCoder:
Should we keep both changes independent? I was hacking on the beta warp before the package was created, but then I had abandoned the effort. I would like to continue on that right now, now that we have a stable API.

MechCoder:
> Spearmint uses Beta warping by default.

No surprises with regards to that. The authors of Spearmint wrote that paper as well :-)

betatim:
Could be separate PRs. Would you have to wait for this to be merged?

MechCoder:
Not really. The beta warp transforms the space to [0, 1] (since it is a CDF)

betatim:
I only had a quick glance at the paper, it looked like beta warping needed the variables to be in [0, 1] already.

We should provide a simple transform which scales everything to [0, 1] linearly and have beta warping as a separate tool.

MechCoder:
Seems like it. Would you like to continue this PR?

On a side note, do you understand how to normalize to [0, 1], when the bounds are say (10**-5, 10**5)?

MechCoder:
Hmm. I think we just provide the input to kernel after sampling uniformly from [0, 1] and do whatever is necessary in the inverse transform.

betatim:
> Would you like to continue this PR?

Wasn't planning to

glouppe:
I'll resume working on this in the next days or so.

MechCoder:
While you are it, do you mind recording the values that you get while comparing the performance. I feel it will be useful while handling (https://github.com/scikit-optimize/scikit-optimize/issues/205)

skopt/space.py
```diff
@@ -23,6 +23,23 @@ def inverse_transform(self, Xt):
         return Xt
 
 
+class _Normalize:
```
MechCoder: Should we rename this to maybe `HypercubeTransform` and make it public? `Normalize` can be confused with unit norm normalization.

MechCoder:
@glouppe This seems quite ready. What is left to do?

betatim:
Do you want to complete it @glouppe or can someone else cherry-pick the commits?

glouppe:
You can go ahead and take it, but given the other changes we had meanwhile, it might be simpler to just start from scratch.

MechCoder:
just fyi, working on this now

MechCoder:
see https://github.com/scikit-optimize/scikit-optimize/pull/237

Issue #108, opened 2016-07-16
Conditional search spaces


betatim:
Specifying and searching spaces where some parameters depend on others, for example a classifier's hyper-parameters which depend on your choice of classifier.

nfcampos:
do you guys have any idea about wha sort of api you'd like for this?
just a function that validates `Space` samples? or something more involved than that?

nfcampos:
do you guys want to do this before or after 0.1?

glouppe:
after

felipeportella:
Hi, there is any design or plan around this feature? I'm in need for it and I'm willing to spend some time to code something ... so maybe is better is I follow some design already planned to make it easier a future PR ...

PR #164, opened 2016-07-28
[MRG + 1] Tuning hyperparameters in a pipeline

cmmalone:
I didn't see an example of optimizing over the hyperparams of several sklearn objects assembled into a pipeline (i.e. a pipeline generalization `examples/hyperparameter-optimization.py`).  I took a couple hours and made a notebook showing how to do this, if that example seems useful.

It's kinda drafty and terse compared to the other (really nice!) example notebooks.  I'm happy to add/modify if you can think of more stuff that might be interesting to show here.


glouppe:
Thanks! 

A couple of comments:

- `pipeline.set_params(pca__n_components=N)` is the recommended syntax for setting parameters.
- instead of creating a new notebook, shall we rather expand or add a note in `hyperparameter-optimization`?

betatim:
I would put links between the two notebooks ("if you liked this notebook on HP opt, maybe you want to keep reading and learn about pipelines" or something) but keep them separate. I think it makes it easier to discover what examples we have and keeps them a bit more self-contained.

@cmmalone can you add some plots, everyone loves plots 📈 📊. Not sure exactly what to plot ... how the performance changes over the iterations of the optimisation? Dependence on the parameters?

Who generates the warnings you are silencing? Is it from `skopt` or `sklearn`? If they are from us could you post them here/new issue so we can fix what ever they are warning about?

glouppe:
@betatim Well, in this case I dont agree. Tuning pipelines is not different from tuning a regular estimator. You just have to know the syntax. Everything else remains exactly the same.

betatim:
I am +0 on splitting, if you guys have stronger opinions on this I go with your decision.

cmmalone:
Thanks for the comments!  I'm working on them now.

@betatim the warnings I see are coming from the call to `gp_minimize`; here's the message:

`/Users/civisemployee/src/scikit-optimize/src/sklearn/sklearn/utils/extmath.py:368: UserWarning: The number of power iterations is increased to 7 to achieve higher precision.`

betatim:
That warning is generated by something inside `sklearn`. Silencing that warning is being tracked in scikit-learn/scikit-learn#6746

glouppe:
Sorry if I was not clear, but my suggestion was not to concatenate the two notebooks together. At the moment the two are almost identical and I am not sure it is very helpful to repeat things like you propose. I think we could rather simply add a paragraph near the definition of the objective function to explain how to adapt the code to pipelines. The rest of the original content would stay the same.

cmmalone:
@glouppe For sure, I think it can be merged much more elegantly--I got halfway through my work yesterday and then something else came up and I wasn't able to finish.  

For how to actually make it clear, though, how simple it is to do pipelines once you've done e.g. a classifier--I've personally always found code snippets more helpful than prose explanations, but like you say, the code that would demonstrate this repeats a lot of what's already there.  What if I do something like this--keep the first part of the notebook largely the same, with just a note to see below if you're interested in pipeline stuff.  Then, where my pipeline example is now, replace the digits dataset with Boston housing (so it feels more like the same problem), build a pipeline with PCA and the same classifier that you used before, and then the convergence plot would show 6 traces: 3 classifier only, and 3 transformer+classifier.   Honestly I'd be a little curious to see how convergence changes as we toggle PCA on/off.

Let me know how that sounds to you.

betatim:
Could we make the pipeline example (second half of the notebook) something that optimises parameters of algorithms and contents of the pipeline? At its simplest this would be one step of preprocessing (with parameters to tune) plus one step of classifications (with parameters to tune). You would end up optimising over the choice of preprocessing step and classifier as well as their respective parameters. I'd find that quite exciting.

Caveat: our API doesn't yet explicitly support conditional search spaces (#108) but maybe that isn't a problem in practice for this example.

glouppe:
> For how to actually make it clear, though, how simple it is to do pipelines once you've done e.g. a classifier--I've personally always found code snippets more helpful than prose explanations

I agree with you on this. Actually, couldn't we maybe add a paragraph on this, then have below some code snippet defining a second objective function and then document that this new objective function can be plugged right in the minimizer? As Tim describes, you could even show that it may even be more complicated than that. E.g. something along the line of:

```
def objective(params):
    decomposition = params[0]   # categorical, either "pca" or "ica"
    if decomposition == "pca":
        p = make_pipeline(PCA(), SomeClassifier(...))
    else:
        p = make_pipeline(ICA(), SomeClassifier(...))
    
    # then set params of the classifier
    p.set_params(clf__XXX=params[1])
    p.fit(X, y)
    ...
    return score
```

cmmalone:
Oh I really like that idea, of playing around with the fundamental pieces in the pipeline.  I was thinking to go in that direction anyway once this PR was complete, so yeah, I'm on board.

So where we stand now is, I've got my local version of the code into a place where it's comparing 6 pipelines (3 optimization schemes x switching PCA on/off) and it's close to being ready for a commit.  One thing I've found, though, is that PCA on the Boston dataset doesn't seem very methodologically well-founded, since (among other things) that data is fairly low-dimensionality to begin with.  Any objections to moving the example over to the digits dataset, where PCA makes more sense?

I should have an updated version over the weekend that finishes up what I've started, and then we can talk more about if we want to add @glouppe 's suggestion in this round or merge the PR and circle back for that later.

cmmalone:
Ok, I've cleaned things up a bit, and compared the performance (on the Boston dataset) between GBR alone and PCA+GBR.  Hopefully the flow is a little better, although feel free to disagree :)

A few notes I made as I was working on this.  

1) PCA isn't a particularly well-motivated tool to use on this dataset/problem.  As a result (presumably), the pipelines do worse than pipeline(PCA+GBR).  

2) I played around with switching over to the digits dataset (and adjusting from GBR to GBC) because this seems like an instance where PCA is more theoretically well-motivated, but I found that the optimization then got _really_ slow--sometimes up to 5 seconds per iteration.  Running 50 iterations was clearly a no-go, especially for demonstration/development purposes.  I'm not sure if this is because PCA was slow, or digits is higher-dimensional, or GBC is slower than GBR, or something else, but I wanted to note it--would be interesting (to me) to understand the cause a little better.

3) Also something weird in digits + GBC--the dummy optimization seemed to do really well?  Here's the convergence plot I got for GBC only (no pipeline/PCA):
screen shot 2016-08-07 at 2 49 34 pm
Possibly this was just a coincidence, but it was so counterintuitive that I wanted to mention it.

Now that this commit should basically complete all the stuff I wanted to do with the original pipeline PR, I'd also be happy to talk more about optimizing over components of the pipeline (e.g. PCA vs. ICA).  It's really interesting and should be done.  But I might suggest that we save that for a future PR.  Thoughts?

betatim:
The trouble with trees is that they are pretty robust wrt their inputs -> you don't benefit a lot from preprocessing. One thing that might help (a tree based model) is building additional, new features from existing ones (even constructing `a + b` requires a fairly deep tree). For [polynomial features](http://scikit-learn.org/stable/modules/preprocessing.html#generating-polynomial-features) there is a tool in `sklearn`. Just an idea.

What happens if you increase `n_components`'s range up to 13 (the original number of features in boston)?

Did you run it for more than 50 iterations to see if it eventually catches up with the no PCA setup?

Before going through the example for formatting and word smithing we (@cmmalone, @glouppe) should decide: are we happy with the general structure/problem setup? GBR followed by a pipeline example of PCA+GBR on boston.

I am kind of happy with PCA+GBR. My preference would be PolynomialFeatures+GBR.

Thanks for the patience @cmmalone.
---

It is strange that random search outperforms the others. Before drawing any conclusions I would [run it several times](https://scikit-optimize.github.io/notebooks/strategy-comparison.html) because a single instance means nothing when random numbers are involved 😀 Though it seems in general, on low dimensional problems random search is hard to improve on.

In this particular run it looks like random search got lucky and picked a good point early on (and then doesn't improve on it any further)

cmmalone:
Heya,

I ran a bunch of little studies.   Here's what I was finding:

1) Boston data, GBR alone, error = 2.78 (n_calls=30, GP optimizer)
2) Boston data, polynomial features + GBR, error = 3.05 
3) Boston data, PCA + GBR, error = 3.67

Running out to n_calls = 75 doesn't really change the picture.

I also tried a bunch of analogous stuff with the digits dataset and a classification approach.  The same general trends held, albeit much more slowly because for some reason the GBC takes FOREVER on the digits dataset.  

So, none of these approaches seems to really illustrate the point that we want to communicate, namely that doing feature engineering might be a good idea.  Upon some reflection I can see intuitively why this makes some sense for PCA (a well-validated model in general should do worse when you start doing dimensionality reduction) but I'm still scratching my head about polynomial features.  

Now, here's one idea of where we can take this.  Sure, dimensionality reduction might not win you any advantages over just using the full dataset.  However, say we have a given feature budget, like max number of features = 10 or whatever--I bet that we'd see a difference between e.g. PCA for preprocessing vs. univariate feature selection, and of course n_components/n_features should make some difference.  I'd be happy to hack on something like we discussed earlier, where we optimize over which pieces to actually assemble in the pipeline, as well as the parameters of the pieces.  But I think that might be getting more toward a regime where we'd want a separate notebook for that, so as not to invite comparisons like the ones we're doing now (i.e. why even mess around with the pipeline if the raw estimator works better?) 

Just an idea, I'm happy to keep wrangling this for as long as you guys don't mind working with me on it.

One last thing, that became viscerally clear in this work--imagine a pipeline where you have polynomial features + classifier.  And suppose that the optimal degree of polynomial features is 3.  In the case I was examining (digits--probably not benefitted by poly features, but that's not the point here), doing the poly features transformation for degree = 3 is SLOW.  But, the optimization is going to do it a lot, because it's giving the best results.  So you end up with a bottleneck where it spends a ton of time repeating the same poly feature transformation while it's trying to find the best parameters for the classifier--not ideal.  I'm imagining a possible algorithm especially for pipeline optimizations like this (well-defined interfaces between the pieces) where you can do the transformations once and then repeatedly just reload them while optimizing the downstream stuff.  I'm sure this is something you've thought about, and it seems really nontrivial to do for a general use case, but it would have been really nice to have here.

betatim:
🚧 Extreme pragmatism ahead 🚧 
Are @glouppe and @cmmalone happy with the example we have right now?

I would vote to merge the example as is. Yes it might not make the most amount of sense to assemble things this way. IMHO our examples should be more about the technical stuff (how do I do this?) and less about recommending the best approach. "We give you the fork and show you how to use it. You can use it to eat fries or steak or vegetables. Which one you should be eating is up to you."

In the end SMBO is about discovering things that you didn't know. So we can formulate the example as "Naively you might think polynomial features would help, turns out it doesn't. Who knew? Good thing you tried it."
🚧 Extreme pragmatism over 🚧 

A bigger example that swaps in and out pipeline components and finds their best parameters would be interesting to me. Definitely a new notebook though. Let's start a new issue if we ant to discuss this further.

I would say that caching results that are repeatedly computed is a smart move, building the infrastructure for that is outside of the scope of this particular library though.

Personally I find both very interesting, however as `scikit-optimize` maintainer I would say "too specific to the hyperparameter tuning problem, `skopt` is about so much more. Probably useful and popular though so we might consider building an example to get people interested in `skopt`."

MechCoder:
@cmmalone You might want to inherit `cross_val_score` from `model_selection` to avoid the deprecation warning.

MechCoder:
Uggh I hate the fact that I cannot comment on the diff :/

MechCoder:
I haven't read all the commentary but I'm fine with merging this as is. I agree that the purpose isn't to show that a pipeline with so and so performs better but just how to use a pipeline in combination with the minimize functions.

(Btw, this is a low dimensional problem with enough data and not sure we should do any dimensionality reduction except for showcase purposes)

@cmmalone Thanks for your work! This gets my +1!

MechCoder:
@cmmalone Oh! Could you set `pipe_x0` to [13] + `x0`? and search across `range(1, 14)`. I'm curious to see what happens.

betatim:
We should just remove that second import of `cross_val_score`.

cmmalone:
Thanks for the comments, everyone, and I'm happy to merge too.  Appreciate your patience.

@MechCoder I've changed the n_components for PCA as you suggested, and gave the pipeline the same starting point for the pipeline and the solo regressor.  Also cleaned up the `cross_val_score` as suggested by @betatim

MechCoder:
We probably want a graph instead of a NameError at the end :-)

betatim:
@cmmalone if you re-run the notebook once more from scratch and update your last commit that would be great. Then merging 🎉 !

cmmalone:
@betatim @MechCoder does it look better now on your end?

MechCoder:
The start point should be [13], instead of 5 no?

cmmalone:
@MechCoder good catch

MechCoder:
Can you rerun it. for one last time.

betatim:
Merging this and using my powers to update the notebook after.

betatim:
🍝 ! Good job everyone and thanks for the patience on all sides.

There are a few ideas in this thread that we wanted to follow up. Create new issues or PRs for those :)

MechCoder:
Umm. did you push the commit to upstream yet?

betatim:
Oups! Done now

PR #165, opened 2016-07-28
[MRG+1] Support parallel fitting to GradientBoostingQuantileRegressor

### LINKED ISSUE #141: n_jobs parameter to GradientBoostingQuantileRegressor
Fitting across the three quantiles (or provided) should be embarassingly parallel.

nel215:
I attempted to fix #141.

### DIFF
--- a/skopt/learning/gbrt.py
+++ b/skopt/learning/gbrt.py
@@ -29,17 +29,17 @@ class GradientBoostingQuantileRegressor(BaseEstimator, RegressorMixin):
         of `GradientBoostingRegressor` are supported. Use this to change
         the hyper-parameters of the estimator.
 
-    * `random_state` [int, RandomState instance, or None (default)]:
-        Set random state to something other than None for reproducible
-        results.
-
     * `n_jobs` [int, default=1]:
         The number of jobs to run in parallel for `fit`.
         If -1, then the number of jobs is set to the number of cores.
+
+    * `random_state` [int, RandomState instance, or None (default)]:
+        Set random state to something other than None for reproducible
+        results.
     """
 
     def __init__(self, quantiles=[0.16, 0.5, 0.84], base_estimator=None,
-                 random_state=None, n_jobs=1):
+                 n_jobs=1, random_state=None):
         self.quantiles = quantiles
         self.random_state = random_state
         self.base_estimator = base_estimator
--- a/skopt/learning/gbrt.py
+++ b/skopt/learning/gbrt.py
@@ -72,10 +72,13 @@ class GradientBoostingQuantileRegressor(BaseEstimator, RegressorMixin):
                 raise ValueError('base_estimator has to use quantile'
                                  ' loss not %s' % base_estimator.loss)
 
+        # The predictions for different quantiles should be sorted.
+        # Therefore each of the regressors need the same seed.
+        base_estimator.set_params(random_state=rng)
         regressors = []
         for q in self.quantiles:
             regressor = clone(base_estimator)
-            regressor.set_params(alpha=q, random_state=rng)
+            regressor.set_params(alpha=q)
 
             regressors.append(regressor)
 
--- a/skopt/learning/gbrt.py
+++ b/skopt/learning/gbrt.py
@@ -79,7 +79,7 @@ class GradientBoostingQuantileRegressor(BaseEstimator, RegressorMixin):
 
             regressors.append(regressor)
 
-        self.regressors_ = Parallel(n_jobs=self.n_jobs)(
+        self.regressors_ = Parallel(n_jobs=self.n_jobs, backend='threading')(
             delayed(_parallel_fit)(regressor, X, y)
             for regressor in regressors)
 
--- a/skopt/learning/gbrt.py
+++ b/skopt/learning/gbrt.py
@@ -4,6 +4,11 @@ from sklearn.base import clone
 from sklearn.base import BaseEstimator, RegressorMixin
 from sklearn.ensemble import GradientBoostingRegressor
 from sklearn.utils import check_random_state
+from sklearn.externals.joblib import Parallel, delayed
+
+
+def _parallel_fit(regressor, X, y):
+    return regressor.fit(X, y)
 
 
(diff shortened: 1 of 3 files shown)


betatim:
Thanks a lot for the PR. I will take a look at it now.

@MechCoder @glouppe should we have our own `externals` or just use the `sklearn` one or ...? Not sure what is the normal thing to do.

MechCoder:
I think for joblib it is all right to use `sklearn.externals`

skopt/learning/tests/test_gbrt.py
```diff
@@ -94,3 +94,22 @@ def test_gbrt_with_std():
...
+
+    rgr = GradientBoostingQuantileRegressor(n_jobs=1)
+    rgr.fit(X, y)
+    estimates = rgr.predict(X)
+
+    rgr.set_params(n_jobs=2)
+    rgr.fit(X, y)
+    estimates_parallel = rgr.predict(X)
+
+    assert_almost_equal(np.mean(estimates, axis=0),
```
MechCoder: You can check `estimates` directly right?

MechCoder:
looks ok to me

skopt/learning/gbrt.py
```diff
@@ -27,13 +32,18 @@ class GradientBoostingQuantileRegressor(BaseEstimator, RegressorMixin):
...
         results.
+
+    * `n_jobs` [int, default=1]:
+        The number of jobs to run in parallel for `fit`.
+        If -1, then the number of jobs is set to the number of cores.
     """
 
     def __init__(self, quantiles=[0.16, 0.5, 0.84], base_estimator=None,
-                 random_state=None):
+                 random_state=None, n_jobs=1):
```
betatim: Could you swap the order of `random_state` and `n_jobs`?

skopt/learning/tests/test_gbrt.py
```diff
@@ -94,3 +94,22 @@ def test_gbrt_with_std():
...
+                        np.mean(estimates_parallel, axis=0))
```
betatim: If you set (and reset) `random_state` for the regressor the return value from `fit()` should be exactly the same, which you can compare with `assert_almost_equal` instead comparing the summary statistic (the mean).

betatim:
I think we can use the [`threading`](https://pythonhosted.org/joblib/parallel.html#using-the-threading-backend) backend with GBT which is more lightweight than `multiprocessing`.

Can you add a parameter to `gbrt_minimize` (and `forest_minimize` for ET and RF) that allows the user to control the number of jobs?

nel215:
Thanks for the comments! I'll fix these.

nel215:
I made some fixes.

Due to the change of handling random_state, the `test_tree_based_minimize` test failed.
So, I increased `n_call` of the test. Do you have a better idea than this?

betatim:
Why do you think we need [different seeds](https://github.com/scikit-optimize/scikit-optimize/pull/165/files#diff-011506ca7d0237d223cbe785bf810be7R80) for each of the regressors?

The thinking behind using the same was to try and ensure that the quantiles will always be sorted. The predictions for different quantiles (here the Xth quantile and the X+eps'th quantile) should be sorted: `q(X) < quantile(X+epsilon)` which if the two regressors are "independent" (because they have different seeds) might not be true. I never benchmarked this to see how big a problem this is in practice.

nel215:
I may misunderstand it. So let me ask you some questions.

If it is necessary to use the same seed for each of the regressors, the master branch wrong?
Should the following code output the same integer values?

```
> cat a.py
from sklearn.ensemble import GradientBoostingRegressor
import numpy as np

random_state = np.random.RandomState(1)

base_estimator = GradientBoostingRegressor(loss='quantile')
quantiles=[0.16, 0.5, 0.84]

for q in quantiles:
    regressor = clone(base_estimator)
    regressor.set_params(alpha=q, random_state=rng)
    print(regressor.random_state.randint(10))
> python a.py
5
8
9
```

If it is necessary to output same integer values, I think that the following code is required.

```
> cat b.py
from sklearn.ensemble import GradientBoostingRegressor
import numpy as np

random_state = np.random.RandomState(1)

base_estimator = GradientBoostingRegressor(loss='quantile')
quantiles=[0.16, 0.5, 0.84]

base_estimator.set_params(random_state=rng)
for q in quantiles:
    regressor = clone(base_estimator)
    regressor.set_params(alpha=q)
    print(regressor.random_state.randint(10))
> python a.py
5
5
5
```

Is this correct?

betatim:
Yes indeed `master` is wrong in this case! Good catch. My intention was definitely to have `regressor.random_state.randint(10)` print the same number for each regressor.

When you change the code to do that, can you add a comment that summarizes our discussion so that in the future people reading the code know why we want each regressor to produce the same random numbers.

skopt/tests/test_tree_opt.py
```diff
@@ -91,6 +91,6 @@ def test_tree_based_minimize():
         yield (check_minimize, minimizer, bench4, 0.0,
                [("-2", "-1", "0", "1", "2")], 0.05, 10)
         yield (check_minimize, minimizer, branin, 0.39,
-               [(-5.0, 10.0), (0.0, 15.0)], 0.1, 100)
+               [(-5.0, 10.0), (0.0, 15.0)], 0.1, 160)
```
glouppe: I did some changes recently on this, you will need to rebase

nel215:
Ok. I'll fix it.

nel215:
I fixed.

glouppe:
Thanks a lot for your contribution @nel215, this looks good to me! What do you think @betatim ?

betatim:
Merged! Thanks for the contribution and thanks for spotting&squashing that bug.

I added you to [`AUTHORS.md`](https://github.com/scikit-optimize/scikit-optimize/blob/master/AUTHORS.md) and ['CHANGELOG.md`](https://github.com/scikit-optimize/scikit-optimize/blob/master/CHANGELOG.md)

nel215:
Thanks for the review :)

PR #167, opened 2016-07-29
[MRG] Fix 2d partial dependence plots

glouppe:
Before:
download

After:
download 3

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -140,7 +140,7 @@ def _format_scatter_plot_axes(ax, space, ylabel):
 
 
 def partial_dependence(space, model, i, j=None, sample_points=None,
-                       n_samples=100, n_points=40):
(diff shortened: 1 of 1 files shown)


glouppe:
ping @betatim

skopt/plots.py
```diff
@@ -226,14 +226,14 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
             row = []
             for y_ in yi_transformed:
                 rvs_ = np.array(sample_points)
-                rvs_[:, (i, j)] = (x_, y_)
+                rvs_[:, (j, i)] = (x_, y_)
```
glouppe: This was the issue :)

betatim:
Thanks for figuring this out! 🔍  and next time something is weird we should remember to use simple functions (polynomials) but with different ranges for each dimension.

PR #168, opened 2016-07-29
x axis tick labels, log scale and zi.T

betatim:
All changed are related to plotting.

Add `x` axis tick labels for plots on the diagonal

Add option to switch between `linear` and `log` scale for
contour plots.

Switching to log scale for `branin` revealed that we
also need to transpose the `zi` for the contour plots. After the fix
in #167 and seeing `branin` in log-scale it was obvious we needed to also transpose it.

Before:
screen shot 2016-07-29 at 13 45 16

After:
screen shot 2016-07-29 at 13 46 25

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -239,7 +239,7 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
 
 
 def plot_objective(result, levels=10, n_points=40, n_samples=250,
-                   z_scale='linear'):
+                   zscale='linear'):
     """Pairwise partial dependence plot of the objective function.
 
     The diagonal shows the partial dependence for dimension `i` with
@@ -272,7 +272,7 @@ def plot_objective(result, levels=10, n_points=40, n_samples=250,
         Number of random samples to use for averaging the model function
         at each of the `n_points`.
 
-    * `z_scale` [str, default='linear']
+    * `zscale` [str, default='linear']
         Scale to use for the z axis of the contour plots. Either 'linear'
         or 'log'.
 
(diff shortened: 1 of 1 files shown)


skopt/plots.py
```diff
@@ -230,10 +235,11 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
                 row.append(np.mean(model.predict(rvs_)))
             zi.append(row)
 
-        return xi, yi, zi
+        return xi, yi, np.array(zi).T
```
betatim: You were right @glouppe we needed to `zi.T` after all.

glouppe: Combo bug! <3 matplotlib

betatim: This is like the multi kill of FPS games right?

betatim: On a more serious note. This used to be a `list of lists` as return type. Now it is an array. I had justified `list of lists` because the points were in the original space not the transformed space. Now I think that was a stupid justification and it should just be a `np.array`.

skopt/plots.py
```diff
@@ -230,10 +235,11 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
...
 
 
-def plot_objective(result, levels=10, n_points=40, n_samples=250):
+def plot_objective(result, levels=10, n_points=40, n_samples=250,
+                   z_scale='linear'):
```
glouppe: `zscale` for consistency  with `plot_convergence`? (where `yscale` exists, and corresponds to the param name in matplotlib)

glouppe:
Haha! now this finally makes sense :) +1 for merge, except the small comment

skopt/plots.py
betatim: Done.

glouppe:
Great! Then please merge and squash this bug! :)

betatim:
🐞 meet 🔨

PR #169, opened 2016-07-30
[MRG] Ilustration of SMBO using gp_minimize

MechCoder:
This notebook was previously `gp_minimize_1d.py`. I thought it would still be useful to show it in comparison with `forest_minimize`


MechCoder:
We have discussed this before, but the RF variances do not have the same effect of GP variances.

Anything that I am doing obviously wrong here? cc: @betatim @glouppe

glouppe:
> We have discussed this before, but the RF variances do not have the same effect of GP variances. (i.e the variances at the points close to the points previously queried do not reduce)

They would reduce if you had more points. At the scale of the notebook (a mere dozen), the effect of grouping observations together in the same leaves (due to `min_samples_leaf`) is too strong to obtain accurate local estimates.

MechCoder:
So would we need a default of `min_samples_leaf` such that the variance estimates do not become zero and it should solve the problem that I describe?

MechCoder:
Not sure that is easy to obtain though :(

betatim:
I would phrase it like this: in a GP based model you add information by hand that says "make the variance ~zero at a sampled point" where as in a tree based setup you don't provide any information "by hand" and instead use only the data (in this case the samples) speak for themselves.

Conclusion: there isn't much we can do other than maybe recommend that if the user is in a situation where things are driven by their prior beliefs (kernel choice) they should use a GP and if they have a problem with a large number of samples use trees?

Side note: in this case it doesn't seem to matter much as the tree based model finds the correct minimum anyway?

glouppe:
The name of the notebook file is too long. It is not ideal, but the name should be short, so that it renders well on the left menu of the documentation. It should also be in lowercase with hyphens between words.

betatim:
Now that this is a notebook we need to invent some narrative around the examples.

Can we refactor the two plotting functions to be shared between GP and forest?

betatim:
Thoughts on what needs doing:

* add a narrative:
  - Title, author
  - intro paragraph outlining what will be discussed
  - one sentence about what you will learn
  - one paragraph each explaining what you can see in each plot
  - final paragraph summarising what you saw and that you now learnt what ever we promised you would learn 😀 
* file name: `simple_1d_minimize.ipynb`? `simple_1d_minimization.ipynb`?
* refactor the plotting for gp and forest into one function
* on the plots, should we call it "forest based ..." or "Extra trees based ...". From a language lawyer perspective: is ET a forest or its own thing?

Would it be easier to understand/illustrate that the forest based models need a few samples before they can make a reasonable estimate of the variance if we used a simple DT? It should be directly visible which `min_samples_leaf` samples are grouped together. Or maybe show it with a DT first to get the concept across and then ET "for real applications"?

Is https://github.com/scikit-optimize/scikit-optimize/pull/169#issuecomment-236514575 a correct summary? "GP does the more intuitive thing in this case as it profits from the human guidance in the form of the kernel. Trees have no 'human input' so don't look nice for small `n`. In this case both find the minimum."

Looks like a lot of comments but I think really we are in good shape here: 👍

MechCoder:
I shall address your comments, when I am less sleepier.

Btw, I may not be completely right but here is my interpretation on why it finds the minimum in this particular example

 Before the first RF fit, the points randomly sampled are already evenly spaced throughout the search region. So the RF is able to fit a good approximation to the original function. In case, the points are sampled are from just one particular region, (This can happen if the search space is too huge). Then the `forest_minimize` will be stuck in a sub-optimal space since the variance everywhere is typically constant. And since the variance is constant, it explores the unexplored space by chance, (when we do random sampling) rather than guided by the acquisition function.

In the current state, IMHO we should do some pseudo-random sampling such that the first few random points are evenly distributed in the search space. (I think this is what is done in Spearmint as well)

MechCoder:
See: https://github.com/HIPS/Spearmint/blob/master/spearmint/grids/sobol_grid.py

betatim:
Quasi random sequences are cool. I thought replacing `RandomGridSearchCV` with one using Sobol sequences would make me famous but then I read [Section 4](http://www.jmlr.org/papers/volume13/bergstra12a/bergstra12a.pdf) and concluded it would be better to build `skopt` 😁 It seems the improvement over random search for hyper-parameter tuning is marginal.

betatim:
Feels like we are going a bit off-topic here. The original idea behind the example was to show how easy it is to switch between methods and give an idea on what SMBO is. To discuss the trade offs between tree based models and GPs etc I think we should move to a new issue/PR. What do you think?

MechCoder:
All right, let us hold this thought. I wanted to do some work on this today but my net connection died. Hopefully, tomorrow there will be some time.

MechCoder:
See: https://github.com/scikit-optimize/scikit-optimize/pull/187 for another implication of pseudo-random sampling.

MechCoder:
Ok, I initally wanted to point out the drawbacks of `forest_minimize` in a notebook. But now I took the easier route of just showing SMBO on a 1-D model using `gp_minimize`

glouppe:
How different is that from the bayesian-optimization notebook we already have? shouldnt we rather improve over this one?

MechCoder:
Yes, I guess, which parts which you would like integrating? Would you prefer replacing replacing the last graph in your notebook with the last graph in mine?

(I really like the last graph in my notebook and would like to show it in some way or the other)

betatim:
👍 on using your plot in the existing notebook. Seeing BO step-by-step helped me understand what it was doing so I also like your plot.

I have a slight preference for the plotting style (colour scheme and all in one figure) of the `bayesian-optimization` notebook. It uses LCB which lends itself to being plotted in the same figure as the GP mean.

PR wise it would be easier to stick with LCB, but then we recommend people use EI everywhere 😕 so should we update the example? This would be a bit more work and would need careful checking that all the text is still valid. I would also "transpose" the axes @MechCoder so that you have GP and EI next to each other and each new row is a new step of the SMBO. This way we can have bigger figures.

Also fine adding it to the `README` once merged.

MechCoder:
Thanks! I've refactored the example. The reason we did not use EI for illustrative purposes, was that it converges rather quickly. (Not sure how much that applies to this example though)

betatim:
Travis is unhappy because of a "stalled build", so I restarted the build 😕 

Otherwise LGTM, @glouppe?

@MechCoder can you also modify the `README` in this PR?

glouppe:
Looks nice! thanks for this! I am merging, we can modify the readme afterwards

betatim:
🎉

Issue #163, opened 2016-07-28
Repeating discrete (integer/categorical) entries


ardunn:
Does anyone here have a workaround for repeated integer/categorical entries? For example, if I am searching the space 
`dimensions = [(1,10), (1,10),["red","green","blue"]]`
With a local optimum of:
`my_input = [1,1,"red"]`
Skopt will keep testing the same entries at or around this optimum. If anyone knows a way around these repeated calculations, it would be much appreciated!

MechCoder:
Hi,

Typically `gp_minimize` should not get stuck in a local minima because values far away from the local minima should have a high variance, hence one of these values with high exploration should be picked.

Could you try setting `n_restarts_optimizer` to something higher along with `search='lbfgs'` and let us know if the problem persists?

Also, is it possible to let us know the function that you are trying to optimize?

Thanks!

ardunn:
Sure, I will try that and get back to you.

The function I am trying to optimize is not visible to Skopt (I removed the need for a function call by just passing the input and output of a black-box function to Skopt instead) but the search space is basically of the form I was describing above: Two integer dimensions and a categorical dimension. The integer dimensions are both 52 integers and the categorical dimension is 7 categories, giving a total of 18,928 possible combinations. 

I am running into a repeat issue within 200-300 iterations, which is surprising because the highest scorers of my function are well spread throughout the entire space, and Skopt has not exhaustively found all high scorers.

ardunn:
For what I am doing I needed to modify the skopt file to move the optimization loop outside skopt since each function evaluation is highly expensive. I reassigned `yi` to `my_input`, `Xi` to `my_output` and made `gp_minimize` return `next_x` instead of `res` - everything else is default skopt. 

A simplified version of my problem is this. 

```
dim = [(1,10),(2,20),("red","blue","green","orange")]
my_input = [[9, 2,"green"]]
my_output = [12]

def BBfun (x):
    "highest score is 10, 20, red"
    score = 0
    score+= x[0]*x[1]
    if x[2] == "red":
        score+=10
    elif x[2]=="orange":
        score+=5
    elif x[2]=="green":
        score+=1
    else:
        score-=5
    return score*-1

for i in range(20):
    print i
    guess = gp_minimize(my_input,my_output,dim3, search='lbfgs', n_restarts_optimizer=10)
    my_input.append(guess)
    my_output.append(BBfun(guess))
```
[...]

PR #170, opened 2016-07-31
[WIP] Avoid costly function evaluations for points that have been evaluated before

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/163

In the case of discrete spaces, I modified the optimization algorithm from choosing the point that has the best acquisition value to a point that has the best acquisition value and not been evaluated before.

### DIFF
--- a/skopt/gp_opt.py
+++ b/skopt/gp_opt.py
@@ -17,6 +17,8 @@ from sklearn.utils import check_random_state
 
 from .acquisition import _gaussian_acquisition
 from .space import Space
+from .space import Categorical
+from .space import Integer
 from .utils import in2d
 
 def _acquisition(X, model, y_opt=None, method="LCB", xi=0.01, kappa=1.96):
@@ -168,6 +170,8 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
     # Check params
     rng = check_random_state(random_state)
     space = Space(dimensions)
+    is_discrete_space = all(
+        [isinstance(dim, (Categorical, Integer)) for dim in space.dimensions])
 
     # Default GP
     if base_estimator is None:
@@ -241,20 +245,21 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
             X = space.transform(space.rvs(n_samples=n_points,
                                           random_state=rng))
 
-            # Choose the point that gives the best acquisition
-            # value AND has not been reevaluated at before.
-            # This prevents costly function reevaluations.
-            # while exploring more of the search space.
-            X_in_Xi_transform = in2d(X, Xi_transform)
+            if is_discrete_space:
+                # Choose the point that gives the best acquisition
+                # value AND has not been reevaluated at before.
+                # This prevents costly function reevaluations.
+                # while exploring more of the search space.
+                X_in_Xi_transform = in2d(X, Xi_transform)
 
-            # Highly unlikely corner case.
-            if np.all(X_in_Xi_transform):
-                warnings.warn(
-                    "Optimization procedure ended prematurely since the "
-                    "search space has been exhaustively searched. Set "
-                    "'n_points' to a higher value.")
-                break
-            X = X[~X_in_Xi_transform]
+                # Highly unlikely corner case.
+                if np.all(X_in_Xi_transform):
+                    warnings.warn(
+                        "Optimization procedure ended prematurely since the "
+                        "search space has been exhaustively searched. Set "
+                        "'n_points' to a higher value.")
+                    break
+                X = X[~X_in_Xi_transform]
             values = _gaussian_acquisition(
                 X=X, model=gp, y_opt=np.min(yi), method=acq,
                 xi=xi, kappa=kappa)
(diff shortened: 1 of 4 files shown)


MechCoder:
ping @glouppe @betatim

skopt/gp_opt.py
```diff
@@ -161,13 +163,15 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
...
     """
-    # Save call args 
+    # Save call args
     specs = {"args": copy.copy(inspect.currentframe().f_locals),
              "function": inspect.currentframe().f_code.co_name}
 
     # Check params
     rng = check_random_state(random_state)
     space = Space(dimensions)
+    is_discrete_space = all(
```
glouppe: Not sure I agree with this definition of discrete space. We should be clearer. `Integer` are discrete. Same if only one of the dimensions (not all) is `Categorical` or `Integer`.

skopt/gp_opt.py
```diff
@@ -190,7 +194,7 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
     n_init_func_calls = len(x0) if y0 is None else 0
     n_total_init_calls = n_random_starts + n_init_func_calls
 
-    if n_total_init_calls <= 0:
+    if n_random_starts + len(x0) <= 0:
```
glouppe: why this change?

skopt/tree_opt.py
```diff
@@ -19,13 +20,18 @@
 from .learning import GradientBoostingQuantileRegressor
 from .learning import RandomForestRegressor
 from .space import Space
+from .space import Categorical
+from .space import Integer
+from .utils import in2d
```
glouppe: what is this function?

skopt/utils.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+
+def in2d(arr1, arr2):
```
glouppe: we can find a much better name

skopt/gp_opt.py
```diff
@@ -227,28 +231,48 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
...
                                           random_state=rng))
+
+            if is_discrete_space:
+                # Choose the point that gives the best acquisition
+                # value AND has not been reevaluated at before.
+                # This prevents costly function reevaluations.
+                # while exploring more of the search space.
+                X_in_Xi_transform = in2d(X, Xi_transform)
+
+                # Highly unlikely corner case.
```
glouppe: Shall we really stop? if there are remaining calls, we might still find an unvisited point.

skopt/gp_opt.py
```diff
@@ -257,9 +281,23 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
...
-
-                if a < best:
-                    next_x, best = x, a
+                    X.append(x)
+                    values.append(a)
+
+            X = np.asarray(X)
+            values = np.asarray(values)
+
+            if is_discrete_space:
```
glouppe: I am not convinced this is a good strategy for `lbfgs`. It that optimization converges, it may always find the same (visited) point, even if the space has not been exhaustively searched.

glouppe:
`dummy_minimize` should also be changed

glouppe:
What do we do when the function is noisy? It this case, evaluating it more than once might be helpful.

skopt/gp_opt.py
glouppe: Also, couldnt we rather extend the space API with a property or something?

glouppe:
Shall we focus on that for 0.1? I would personally postpone it, as it is not as trivial as it appears.

skopt/gp_opt.py
MechCoder: On a side note, not sure it is a good strategy to use `lbfgs` for these spaces. I sent a PR at https://github.com/scikit-optimize/scikit-optimize/pull/172 to fix that.

betatim:
I agree with https://github.com/scikit-optimize/scikit-optimize/pull/170#issuecomment-236412433 so I think we should try and find some cases (and make examples for them) to study when/why this happens. If only so we can give a recommendation to the user, who will most likely be confused that it is happening and think it is a bug (like we did when we first saw it). So 👍 for post-0.1

MechCoder:
By a noisy function, do you mean that which can have more than one value at a given point?

MechCoder:
If we do consider those, then I would just add a warning if a point is chosen more than once and not do anything special. Let us hold the discussion off to after 0.1

glouppe:
> By a noisy function, do you mean that which can have more than one value at a given point?

Yes, and I believe we should assume it is stochastic unless otherwise specified. (E.g. when tuning hyper-parameters of a model, the CV score is a noisy estimate of the true generalization of the model)

betatim:
I would prefer to merge this now with code that assumes that the objective is noisy -> give a warning when a point is selected again. Then create an issue to discuss how to signal that the objective is not noisy in which case this warning might be an error. Having the warnings under the assumption of noisiness is useful and we shouldn't delay it until we figure out the API for "not a noisy objective".

(also rebase: 🔁 🎸)

MechCoder:
Yeah cool. I'll do that tomorrow.

Btw, I think we should tune the noise hyperparameter instead of setting it to a constant here (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/gp_opt.py#L35)

skopt/gp_opt.py
betatim: How about instead of trying to decide if a space is "finite", why don't we check if a new sample is already in the list of previously evaluated samples (using `np.allclose` and friends)?

PR #171, opened 2016-07-31
Remove nose from dependencies

MechCoder:
Nose should not be required to build the project.


betatim:
Should we start having a `dev-requirements.txt` that lists `nose` and `pep8` etc?

MechCoder:
That sounds reasonable.

PR #172, opened 2016-07-31
[MRG] Allow search=auto in gp_minimize

MechCoder:
Clearly, lbfgs does not make sense for Integer and Categorical spaces since even though the acquisition function may be continuous, the actual function itself isn't. The lbfgs optimization procedure might lead to values in the transformed space in between 0 and 1, which might break inverse transform.

In this PR, search is set to lbfgs automatically only when all the dimensions are continuous.

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -324,6 +324,13 @@ class Space:
 
         self.dimensions = _dimensions
 
+    @property
+    def is_real(self):
+        """
+        Returns true if all dimensions are Real
+        """
+        return all([isinstance(dim, Real) for dim in self.dimensions])
+
     def rvs(self, n_samples=1, random_state=None):
         """Draw random samples.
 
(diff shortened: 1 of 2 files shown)


MechCoder:
Ok, this seems a bug in `inverse_transform` of `LabelBinarizer` in sklearn which is why it does not break.

```python
r = ["m", "a", "n", "o", "j"]
lb = LabelBinarizer()
lb.fit_transform(r)
t = np.asarray([[0.9, 0.1, 0, 0, 0]])
lb.inverse_transform(t)
# ???
array(['a'], 
      dtype='<U1')
```

MechCoder:
Even for an `Integer` space, we wouldn't have the desired behaviour for `inverse_transform` because of casting to int. We should round it to the nearest integer.

skopt/space.py
```diff
@@ -324,6 +324,13 @@ def __init__(self, dimensions):
 
         self.dimensions = _dimensions
 
+    @property
+    def is_real_space(self):
```
glouppe: `is_real` should be enough? (since typically this property will be checked through something like `space.is_real`)

glouppe:
+1 once my small comment is addressed

glouppe:
Set to `n_points=1000` as part of #174. Closing.

Issue #166, opened 2016-07-29
Double check all default parameters


glouppe:
All default parameters should be double checked, as I believe some of our choices are currently suboptimal.

- [x] gp_minimize: default kernel?
- [x] gp_minimize: default noise level alpha for the GP? or prefer alpha=0.0 + a WhiteKernel() component?
- [x] gp_minimize: search=sampling vs search=lbfgs?
- [x] gp_minimize: when search=lbfgs, maxiter? n_restarts_optimizer?
- [x] gp_minimize: n_points?
- [x] forest_minimize: n_points?
- [x] forest_minimize: default base estimator?
- [x] default acquisition function? from the literature and from our tests, EI seems to be the best choice among the functions we have. 
- [x] default values for kappa and xi?

I am tagging this as major, as these tweaks can change results from poor to great. We should fix that before 0.1.

glouppe:
We should also double check with has being done in other BO packages.

MechCoder:
For `forest_minimize`, I think the default base estimator, that is the extra trees regressor comes closest to SMAC's defaults as it smoothens the decision surface.
Also the number of points are 10000

glouppe:
Fixed by #174 until the next iteration :)

Issue #173, opened 2016-08-01
Solvers: DIRECT and CMA-ES


glouppe:
From looking at several implementations of BO, many are running optional solvers in replacement of LBFGS, including:

- CMA-ES: https://pypi.python.org/pypi/cma
- DIRECT: https://pythonhosted.org/DIRECT/

Those are certainly worth adding, given the hurdles we have encountered so far with LBFGS.

glouppe:
At the same time, might be worth factoring out the minimization logic into a separate `optimizer` module.

PR #174, opened 2016-08-01
[MRG] Better defaults

glouppe:
Exploration for #166, supersedes #144.

- [x] gp_minimize: default kernel?
- [x] gp_minimize: default noise level alpha for the GP? or prefer alpha=0.0 + a WhiteKernel() component?
- [x] gp_minimize: search=sampling vs search=lbfgs? fix by #172 
- [x] gp_minimize: when search=lbfgs, maxiter? n_restarts_optimizer?
- [x] gp_minimize: n_points?
- [x] forest_minimize: n_points?
- [x] forest_minimize: default base estimator?
- [x] default acquisition function? from the literature and from our tests, EI seems to be the best choice among the functions we have. 
- [x] default values for kappa and xi?


glouppe:
All done, I compared with @jmetzen's BO package and with GPyOpt. Not much to change in the end. I think the main source of improvement will be to implement #173.

skopt/gp_opt.py
```diff
@@ -78,6 +78,11 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
     * `base_estimator` [a Gaussian process estimator]:
         The Gaussian process estimator to use for optimization.
 
+    * `alpha` [float, default=1e-10]:
+        Value added to the diagonal of the kernel matrix during fitting.
+        Larger values correspond to increased noise level in the observations
```
betatim: "to increased" -> "to an increased"

skopt/gp_opt.py
```diff
@@ -78,6 +78,11 @@ def gp_minimize(func, dimensions, base_estimator=None, acq="EI", xi=0.01,
...
+        and reduce potential numerical issue during fitting.
```
betatim: issueS

betatim:
How did you pick the values? If you did any kind of benchmarking or other quantitative thing we should store those scripts/notebooks in the repo somewhere so we can use them in the future. (If this is mainly gut feeling and what others do that is also fine :))

Did you look at the examples to see if any of them are affected by the changes in defaults?

Otherwise: LGTM

glouppe:
> How did you pick the values? If you did any kind of benchmarking or other quantitative thing we should store those scripts/notebooks in the repo somewhere so we can use them in the future. (If this is mainly gut feeling and what others do that is also fine :))

I mostly compared with other packages. I did not do a very thorough analysis to be honest. I think we have reasonably good defaults.

glouppe:
> Did you look at the examples to see if any of them are affected by the changes in defaults?

Good point, I'll regenerate all of them.

glouppe:
Okay hop, merging.

betatim:
>> Did you look at the examples to see if any of them are affected by the changes in defaults?

> Good point, I'll regenerate all of them.

Worth making an issue to find out if we can have travis upload the re-rendered notebooks somewhere for each PR?

skopt/tree_opt.py
```diff
@@ -112,7 +112,7 @@ def _tree_minimize(func, dimensions, base_estimator, n_calls,
 
 
 def gbrt_minimize(func, dimensions, base_estimator=None, n_calls=100,
-                  n_points=20, n_random_starts=10, x0=None, y0=None,
+                  n_points=1000, n_random_starts=10, x0=None, y0=None,
```
MechCoder: Since `n_points` is random anyway, why not set it to 10000 for consistency with Spearmint and SMAC?

MechCoder:
@betatim Please do so..

Issue #162, opened 2016-07-28
Notebook explaining how to read `plot_evaluations` and `plot_objective`


betatim:
Following on from https://github.com/scikit-optimize/scikit-optimize/pull/147#issuecomment-235736195

Create a notebook that explains how to use and understand `plot_evaluations`.

Create a notebook that explains how to use and understand `plot_objective`

(This is could be a two PR issue.)

betatim:
I will try and work on this tomorrow.

PR #179, opened 2016-08-02
[MRG] Example for `plot_evaluations` and `plot_objective`

betatim:
An example on how to use `plot_evaluations()` and how to read its output.

Related to #162


betatim:
This also fixes up a mistake in the layout of the 2D plot matrix with the position of the `x` label of the on-diagonal plots and chooses the figure size based on the number of dimensions. (could refactor this into a separate PR if you want)

glouppe:
Will you also combine this notebook with explanations for `plot_objective`?

betatim:
Yes, `plot_objective` in the the same notebook is the plan (but my train arrived :)). This also still needs some tuning in terms of being a good example.

glouppe:
These functions are really great! Love the notebook so far. Ping us once you want reviews :)

glouppe:
Some comments as I read along:

- visualizing-samples.ipynb -> visualizing-results ?
- Visualizing the objective function -> Visualizing optimization results?
- objecitves  -> objectives
- in the intro, maybe recall that the `*_minimize` all return a dict-like object which can inspected. The `plot_*` functions are helpers designed for that purpose. 
-  is it necessary to redefine `branin` and `hart` with an optional noise level if this later argument is not used afterwards? 
-  might be helpful to remind readers what is a partial dependence plot ;)

betatim:
Used my super powers to edit your comment adding checkboxes.

`hart6` is still redefined but without noise so that I can add "noisy" dimensions.

Working on a explanation of partial dependence.

betatim:
All done.

glouppe:
This is really cool! +1 for me, we can still brush things later

betatim:
Merging! 💚

PR #184, opened 2016-08-04
Fixed a minor logic fault related to n_random_starts

yunjie-yang:
When the *_minimize() is fed with x0 and y0, the "n_init_func_calls" should take len(x0) only when y0 is _not_ empty. There was a minor logic fault in the original code.

### DIFF
--- a/skopt/gp_opt.py
+++ b/skopt/gp_opt.py
@@ -198,7 +198,7 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
     if not isinstance(x0, list):
         raise ValueError("`x0` should be a list, but got %s" % type(x0))
 
-    n_init_func_calls = len(x0) if y0 is None else 0
+    n_init_func_calls = len(x0) if y0 is not None else 0
     n_total_init_calls = n_random_starts + n_init_func_calls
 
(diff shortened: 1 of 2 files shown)


MechCoder:
Not sure what this fixes.

If `y0 is None` and `x0` is provided, I need to subtract `len(x0)` from `n_calls` since I have to figure out `y0` as well.

MechCoder:
Oh you might be confused because `n_init_func_calls` does not refer to the `len(y0)` but the number of func cals to initialize `y0`. Does that make it clearer?

Thanks!

yunjie-yang:
Hi MechCoder,

Yes, you are right. I was confused about the purpose of `n_init_func_calls`. My intent was trying to initialize the optimization with given x0 and y0 and allowing `n_random_starts` to be 0. However, I did see you have a recent merge to fix this. So my issue was completely solved. Thank you very much!

yunjie-yang:
@MechCoder I think I forgot to mention you... but I just wanted to say sorry for the confusion and thank you very much for fixing the issue!

glouppe:
Hi Yunjie! Nice to see you posting here :) Do not hesitate to create more issues with comments, suggestions or strange other things. We will be happy to fix them!

Issue #127, opened 2016-07-20
`check_minimize` should pass all the times instead of a majority number of times


MechCoder:
I understand the intention but it is a minimization problem and hence we would like `check_minimize` pass all the times instead of a majority number of times.

(Hopefully that day should come soon)

glouppe:
Honest question: can we really guarantee convergence of a stochastic process, within a finite number of iterations?

MechCoder:
No: But the first 3 benchmark problems should at least pass, don't you think? They are really easy.

On a related note, should we change the way we test `gp_minimize` currently (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/tests/test_gp_opt.py#L18). It tests only for one random state.

MechCoder:
I was concerned that if we make some breaking changes, the present tests may not be strong enough to catch those.

betatim:
One idea would be to [mock](https://docs.python.org/3/library/unittest.mock.html) `np.random` so that it returns a sequence of numbers that we hand crafted and have verified lead to a certain state of the minimiser. This would be faster, give you more fine grained control/tests, and a lot of work.

Then running an overall minimisation like we do right now would become more of an integration test.

PR #187, opened 2016-08-07
[MRG] Check_minimize should pass all times instead of a majority

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/127
The tricky test for passing was bench 2. On examining one of the random states, for that particular random state the initial points were in the region [-6, 5) and hence the RF was not able to be a good surrogate model for bench 2.

The fix was to warm start by pseudo sampling from [-6, 0] and [0, 6]

### DIFF
--- a/skopt/tests/test_forest_opt.py
+++ b/skopt/tests/test_forest_opt.py
@@ -49,20 +49,19 @@ def check_minimize(minimizer, func, y_opt, dimensions, margin,
         assert_less(r.fun, y_opt + margin)
 
 def test_tree_based_minimize():
-    rng = np.random.RandomState(0)
     for name, minimizer in MINIMIZERS:
         yield (check_minimize, minimizer, bench1, 0.,
                [(-2.0, 2.0)], 0.05, 25, 5)
 
-        # This benchmark contains 2 different functions
-        # when x < 0 and otherwise.
-        # Hence sample uniformly from [-6, 0] and
-        # [0, 6] and provide a warm-start.
-        X1 = (-6 + 6 * rng.rand(5, 1)).tolist()
-        X2 = (6 * rng.rand(5, 1)).tolist()
-        X0 = X1 + X2
+        # XXX: We supply points at the edge of the search
+        # space as an initial point to the minimizer.
+        # This makes sure that the RF model can find the minimum even
+        # if all the randomly sampled points are one side of the
+        # the minimum, since for a decision tree any point greater than
+        # max(sampled_points) would give a constant value.
+        X0 = [[-5.6], [-5.8], [5.8], [5.6]]
         yield (check_minimize, minimizer, bench2, -5,
-               [(-6.0, 6.0)], 0.05, 100, 0, X0)
+               [(-6.0, 6.0)], 0.1, 100, 10, X0)
 
         yield (check_minimize, minimizer, bench3, -0.9,
                [(-2.0, 2.0)], 0.05, 25)
--- a/skopt/tests/test_forest_opt.py
+++ b/skopt/tests/test_forest_opt.py
@@ -40,33 +40,40 @@ def test_forest_minimize_api():
                          base_estimator=DecisionTreeClassifier())
 
 
-def check_minimize(minimizer, func, y_opt, dimensions, margin, n_calls):
-    # The result depends on random sampling so run the test several
-    # times and pass if majority of tests converge. Any single instance
-    # converging might just be luck.
-    success = 0
-    N = 3
-
-    for n in range(1, N + 1):
+def check_minimize(minimizer, func, y_opt, dimensions, margin,
+                        n_calls, n_random_starts=10, x0=None):
+    for n in range(3):
         r = minimizer(
-            func, dimensions, n_calls=n_calls, random_state=n)
-        if r.fun <= y_opt + margin:
-            success += 1
-
-    assert_less(N * 0.5, success)
-
+            func, dimensions, n_calls=n_calls, random_state=n,
+            n_random_starts=n_random_starts, x0=x0)
+        assert_less(r.fun, y_opt + margin)
 
 def test_tree_based_minimize():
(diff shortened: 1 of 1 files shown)


betatim:
I would be ok with splitting the search space by hand like proposed here. Though it does assume you have some prior knowledge about the function. Though I suspect we are getting lucky here, we could find a seed for which all samples are on one side of the minimum again. So is the real solution to "know" where the minimum is and use that as split point (instead of zero)?

Another solution is to increase the number of random points to sample before fitting the model (thereby reducing the chances of being unlucky)

Yet another one would be for us to "by hand" always evaluate a set of points close to the boundaries of the search space for tree based models. At least then we would only have to interpolate between points instead of extrapolating.

skopt/tests/test_tree_opt.py
```diff
@@ -40,40 +40,45 @@ def test_forest_minimize_api():
...
-                   np.tile((0.0, 1.0), (6, 1)), 1.0, 100)
+                   np.tile((0.0, 1.0), (6, 1)), 1.0, 75)
         else:
-            yield (check_minimize, minimizer, branin, 0.39,
-                   [(-5.0, 10.0), (0.0, 15.0)], 0.1, 125)
             yield (check_minimize, minimizer, hart6, -3.32,
-                   np.tile((0.0, 1.0), (6, 1)), 1.0, 30)
+                   np.tile((0.0, 1.0), (6, 1)), 1.0, 50)
+
+        if name == "et":
```
betatim: We used to have 125 iterations for all models except DT for which we had 200. So I think you need to reverse the `if`. Why not leave it grouped as it was to make clear that it is only DT that gets more iterations?

MechCoder: ET passes for all with a lesser number of iterations, while the other forest models passes all with more. I reduced the number of iterations for ET for time constraints,

MechCoder:
> Though I suspect we are getting lucky here, we could find a seed for which all samples are on one side of the minimum again.

I agree to that.

> So is the real solution to "know" where the minimum is and use that as split point (instead of zero)?

You can also view it as pseudo-sampling, where each interval gets the same probability of picking points, which is what the test attempts to demonstrate, sampling uniformly from both intervals.

> Yet another one would be for us to "by hand" always evaluate a set of points close to the boundaries of the search space for tree based models.

That would be a great idea to investigate.

MechCoder:
@glouppe @betatim Any objectiions in merging this?

betatim:
>> So is the real solution to "know" where the minimum is and use that as split point (instead of zero)?

> You can also view it as pseudo-sampling, where each interval gets the same probability of picking points, which is what the test attempts to demonstrate, sampling uniformly from both intervals. (thought this is likely to be useless for higher dimesions)

It seems the problem is not sampling evenly on either side of zero but getting a sample >5 no? At least that is how I understood your first comment.

There should be no difference between uniformly drawing 10 samples from [-6, 6] or five from [-6, 0) and five from [0, 6]. If we observe a difference it is just luck/seed. Or is there something I am missing here? I might be, but right now I can't see it :-/

So I would prefer we try a few different seeds until we can find one that works instead of making the code more complicated.

betatim:
*bump*

MechCoder:
So hope the last commit makes you happy :-p

skopt/tests/test_forest_opt.py
```diff
@@ -40,33 +40,44 @@ def test_forest_minimize_api():
...
         yield (check_minimize, minimizer, hart6, -3.32,
-               np.tile((0.0, 1.0), (6, 1)), 1.0, 30)
+               np.tile((0.0, 1.0), (6, 1)), 1.0, 50)
+
+        if name == "et":
+            yield (check_minimize, minimizer, branin, 0.39,
+                   [(-5.0, 10.0), (0.0, 15.0)], 0.15, 125)
+        else:
+            yield (check_minimize, minimizer, branin, 0.39,
+                   [(-5.0, 10.0), (0.0, 15.0)], 0.15, 200)
```
betatim: Can't we use `n_calls=125` for all minimizers? From https://scikit-optimize.github.io/notebooks/strategy-comparison.html my intuition would be that you should be "done" by 125 iterations

skopt/tests/test_forest_opt.py
```diff
@@ -40,33 +40,44 @@ def test_forest_minimize_api():
...
 def test_tree_based_minimize():
     for name, minimizer in MINIMIZERS:
         yield (check_minimize, minimizer, bench1, 0.,
-               [(-2.0, 2.0)], 0.05, 25)
+               [(-2.0, 2.0)], 0.05, 25, 5)
+
+        # XXX: This benchmark contains 2 different functions
+        # when x < 0 and otherwise.
+        # Hence we supply points at the edge of the search
+        # space as an initial point to the minimizer.
```
betatim: Not sure the reason is that the benchmark contains two functions. I think the more important part is getting samples on either side of the minimum. So I'd rephrase the comment to:

"Having samples on either side of the minimum is helpful for tree based methods. For this benchmark
the minimum is close to the edge and so we explicitly provide points at the boundaries of the search
space."

WDYT?

betatim:
Beyond these two, looks good to me.

skopt/tests/test_forest_opt.py
MechCoder: Yeah, I meant to remove that :/

skopt/tests/test_forest_opt.py
MechCoder: lemme see

MechCoder: It seems like the RF model gets stuck in a local optima :(

MechCoder:
Will merge when tests pass.

skopt/tests/test_forest_opt.py
betatim: ok

PR #196, opened 2016-08-12
[MRG] Add one to x-axis for convergence plot

betatim:
It seems a bit weird to have a y axis that says "minimum after n calls" and then start with n=0.


MechCoder:
Thanks!

betatim:
🌴

Issue #140, opened 2016-07-21
dict-like search space


nfcampos:
Would you guys consider support for a dict-like search space, eg.
```py
def objective(params):
  pipeline.set_params(**params)
  return -np.mean(cross_val_score(pipeline, ...))

space = {
  'one_param': (1, 10),
  'other_param': ('a', 'b', 'c'),
}

result = gp_minimize(objective, space)
```
I can put together a PR if you think this makes sense.

glouppe:
I would prefer waiting for the first release before considering extensions
to the API. Worth thinking about though.

betatim:
If we dig into this #108 is related.

PR #198, opened 2016-08-12
Dict-like space

nfcampos:
this PR adds a dict-like Space class (closes issue #140 )

changes: 
- added `DictSpace` class
- added `make_space` function for turning dimensions passed to `*_minimize` into a Space
- added tests for `DictSpace`
- added tests for `make_space`
- updated hyperparameter optimization example to use a dict space (it's much better for passing parameters to a scikit estimator)

questions:
- do you guys want the `dimensions` property of a dict space to be a dictionary or a list? for now I kept it as a list, let me know if i should change that (whatever is decided for this should will also decide whether `bounds` are a dict or a list)

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -486,3 +486,36 @@ class Space:
                 b.extend(dim.transformed_bounds)
 
         return b
+
+
+class DictSpace(Space):
+
+    def __init__(self, dimensions):
+        self.keys = sorted(list(dimensions.keys()))
+        super(DictSpace, self).__init__([dimensions[k] for k in self.keys])
+
+    def rvs(self, n_samples=1, random_state=None):
+        return [
+            dict(zip(self.keys, sample))
+            for sample in super(DictSpace, self).rvs(n_samples, random_state)
+        ]
+
+    def transform(self, X):
+        return super(DictSpace, self).transform([
+            [sample[k] for k in self.keys] for sample in X
+        ])
+
+    def inverse_transform(self, Xt):
+        return [
+            dict(zip(self.keys, sample))
+            for sample in super(DictSpace, self).inverse_transform(Xt)
+        ]
+
+
+def make_space(dimensions):
+    if isinstance(dimensions, Space):
+        return dimensions
+    elif type(dimensions) in (list, tuple):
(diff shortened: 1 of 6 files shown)


skopt/space.py
```diff
@@ -486,3 +486,39 @@ def transformed_bounds(self):
...
+    def inverse_transform(self, Xt):
+        return [
+            dict(zip(self.keys, sample))
+            for sample in super(DictSpace, self).inverse_transform(Xt)
+        ]
+
+
+def make_space(dimensions):
+    if isinstance(dimensions, Space):
+        return dimensions
```
nfcampos: with this change it is now possible to pass to *_minimize a `Space` instance directly, which was not possible before, let me know if you don't like that for some reason and I can remove this branch from the function

betatim:
Travis is unhappy 😢. Some quick thoughts/questions:

* why do we need `DictSpace`? Can't `make_space` take care of the different ways of specifying a space and create `Space` instances for all of them?
* if we get into modifying how to specify spaces we should consider how to specify (in the future, not this PR) a conditional space. Say one dimension is: "no preprocessing" and "PCA" then if you choose "PCA" there will be extra dimensions that don't exist if you choose "no preprocessing".
* I like that we can now specify names for dimensions, I think the names should be stored in the `Dimension` object

I will take a closer look this evening.

nfcampos:
@betatim 
- re: Travis being unhappy, I didn't realise some tests define `dimensions` as a numpy array, I've added explicit support for that to `make_space`
- re: the point of `DictSpace` is that its samples are `dict`s instead of `list`s, which in lots of cases is much better for use inside the objective function
- re: conditional spaces I wrote something for that some weeks ago but never opened a PR, it's now here #199
  - this only supports conditional spaces where the dimensions are always the same but some combinations of values are invalid

nfcampos:
do you guys have any interest in something like this, ie. should I resolve the conflicts with the latest master?

pavelkomarov:
Should we be deleting these old pull requests that never got merged?

nfcampos:
I won’t be updating this so I’m closing it myself.

PR #200, opened 2016-08-12
[MRG+1] added __repr__ and __eq__ for all Dimension classes and Space

### LINKED ISSUE #183: Pretty printing for Space, Real, Integer etc
Definitely not important but would be pretty to have (no pun intended)

nfcampos:
closes #183 
the `__eq__` methods definitely could use some tests, I'll add those, but do the `__repr__` methods need tests as well?

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -177,6 +177,16 @@ class Real(Dimension):
                 "Prior should be either 'uniform' or 'log-uniform', "
                 "got '%s'." % self._rvs)
 
+    def __eq__(self, other):
+        return (type(self) is type(other)
+                and self._low == other._low
+                and self._high == other._high
+                and self.prior == other.prior)
+
+    def __repr__(self):
+        return "Real(low={}, high={}, prior={})".format(
+            self._low, self._high, self.prior)
+
     def inverse_transform(self, Xt):
         """Inverse transform samples from the warped space back into the
            orignal space.
@@ -213,6 +223,14 @@ class Integer(Dimension):
         self._rvs = randint(self._low, self._high + 1)
         self.transformer = _Identity()
 
+    def __eq__(self, other):
+        return (type(self) is type(other)
+                and self._low == other._low
+                and self._high == other._high)
+
+    def __repr__(self):
+        return "Integer(low={}, high={})".format(self._low, self._high)
+
     def inverse_transform(self, Xt):
         """Inverse transform samples from the warped space back into the
            orignal space.
@@ -246,12 +264,22 @@ class Categorical(Dimension):
         self.categories = categories
         self.transformer = _CategoricalEncoder()
         self.transformer.fit(self.categories)
+        self._prior = prior
 
         if prior is None:
             prior = np.tile(1. / len(self.categories), len(self.categories))
 
         # XXX check that sum(prior) == 1
         self._rvs = rv_discrete(values=(range(len(self.categories)), prior))
+
+    def __eq__(self, other):
+        return (type(self) is type(other)
+                and self.categories == other.categories
+                and self._prior == other._prior)
+
+    def __repr__(self):
+        return "Categorical(categories={}, prior={})".format(
+            self.categories, self._prior)
 
     def rvs(self, n_samples=None, random_state=None):
         choices = self._rvs.rvs(size=n_samples, random_state=random_state)
@@ -323,6 +351,13 @@ class Space:
                 raise ValueError("Invalid grid component (got %s)." % dim)
 
         self.dimensions = _dimensions
+
+    def __eq__(self, other):
+        return all([a == b for a, b in zip(self.dimensions, other.dimensions)])
+
+    def __repr__(self):
+        return "Space([{}])".format(
+            ',\n       '.join(map(str, self.dimensions)))
 
     @property
     def is_real(self):
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -10,6 +10,7 @@ from sklearn.utils import check_random_state
 from sklearn.utils.fixes import sp_version
 
 
+# helper class to be able to print [1, ..., 4] instead of [1, '...', 4]
 class _Ellipsis:
     def __repr__(self):
         return '...'
@@ -287,8 +288,14 @@ class Categorical(Dimension):
             cats = self.categories[:3] + [_Ellipsis()] + self.categories[-3:]
         else:
             cats = self.categories
+
+        if len(self.prior) > 7:
+            prior = self.prior[:3] + [_Ellipsis()] + self.prior[-3:]
+        else:
+            prior = self.prior
+
         return "Categorical(categories={}, prior={})".format(
-            cats, self.prior)
+            cats, prior)
 
     def rvs(self, n_samples=None, random_state=None):
         choices = self._rvs.rvs(size=n_samples, random_state=random_state)
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -185,8 +185,8 @@ class Real(Dimension):
 
     def __eq__(self, other):
         return (type(self) is type(other)
-                and self.low == other.low
-                and self.high == other.high
+                and np.allclose([self.low], [other.low])
+                and np.allclose([self.high], [other.high])
                 and self.prior == other.prior)
 
     def __repr__(self):
@@ -231,8 +231,8 @@ class Integer(Dimension):
 
     def __eq__(self, other):
         return (type(self) is type(other)
-                and self.low == other.low
-                and self.high == other.high)
+                and np.allclose([self.low], [other.low])
+                and np.allclose([self.high], [other.high]))
 
     def __repr__(self):
         return "Integer(low={}, high={})".format(self.low, self.high)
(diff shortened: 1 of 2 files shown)


MechCoder:
Looks fine as is. Thanks! Don't think we need tests for the repr's

skopt/space.py
```diff
@@ -246,13 +264,23 @@ def __init__(self, categories, prior=None):
...
         self._rvs = rv_discrete(values=(range(len(self.categories)), prior))
 
+    def __eq__(self, other):
+        return (type(self) is type(other)
+                and self.categories == other.categories
+                and self._prior == other._prior)
+
+    def __repr__(self):
+        return "Categorical(categories={}, prior={})".format(
+            self.categories, self._prior)
```
MechCoder: I have a minor concern here. If there are too many categories, this would not toll great right?

Should we just print the first say 3 - 4 and the last 3 - 4 categories?

skopt/space.py
```diff
@@ -324,6 +352,13 @@ def __init__(self, dimensions):
 
         self.dimensions = _dimensions
 
+    def __eq__(self, other):
+        return all([a == b for a, b in zip(self.dimensions, other.dimensions)])
+
+    def __repr__(self):
+        return "Space([{}])".format(
```
MechCoder: Should we limit the number of dimensions as well?

skopt/space.py
nfcampos: yep limiting the categories sounds reasonable, how many do we want? first 3 and last 3?

skopt/space.py
nfcampos: limiting the number of dimensions I'm not so sure about, even if you have a space with lots of dimensions you probably still want to see all, no?

skopt/space.py
```diff
@@ -246,13 +264,23 @@ def __init__(self, categories, prior=None):
         self.categories = categories
         self.transformer = _CategoricalEncoder()
         self.transformer.fit(self.categories)
+        self._prior = prior
```
betatim: Thanks for finding this 🐛. We should make this `self.prior`, nothing to hide :)

betatim:
All the dimensions inherit from `Dimension`. Can we adopt something generic like in [`BaseEstimator`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/base.py#L290-L293?

Should an instance of `class MyReal(Real)` compare equal to a `Real`? I'm not sure because I am unsure about use cases. (If the answer is yes we should use `isinstance` instead of `type`.

skopt/space.py
nfcampos: should arguments passed to the constructor always be assigned to `self` without the `_` for dimensions? any reason why `min` and `max` have the `_`?

betatim: Not sure actually. I thought we adopted the `sklearn` idea that each argument to the constructor becomes `self.name = value` and things which are computed from them `self.name_ = computed_value`. Right now it seems inconsistent (comparing `self.categories` and `self._low`). Let's see what @MechCoder's opinion is but I'd vote for the `sklearn` style even thought it might be a bit OTT.

skopt/space.py
MechCoder: Sounds fine.

skopt/space.py
MechCoder: Not sure as well. What if there are a 1000 dimensions? I would say let's increase the limit to first 20-25 and last 20-25?

Would appreciate what others have to think as well. cc: @betatim

skopt/space.py
MechCoder: >  I thought we adopted the sklearn idea

Yes indeed, that was the idea. (Though the same reasons, don't apply here)

MechCoder:
> Can we adopt something generic like in BaseEstimator?

The idea seems good, but I feel it is overkill in this situation ;)

skopt/space.py
betatim: So the conclusion is to change `self._low` to `self.low`?

MechCoder: Yes!

nfcampos:
I've made all the changes discussed here I believe

skopt/space.py
```diff
@@ -246,13 +269,27 @@ def __init__(self, categories, prior=None):
...
         # XXX check that sum(prior) == 1
         self._rvs = rv_discrete(values=(range(len(self.categories)), prior))
 
+    def __eq__(self, other):
+        return (type(self) is type(other)
+                and self.categories == other.categories
+                and self.prior == other.prior)
+
+    def __repr__(self):
+        if len(self.categories) > 7:
```
MechCoder: This should be `>=` 7, no?

skopt/space.py
```diff
@@ -246,13 +269,27 @@ def __init__(self, categories, prior=None):
...
+            cats = self.categories[:3] + [_Ellipsis()] + self.categories[-3:]
+        else:
+            cats = self.categories
+        return "Categorical(categories={}, prior={})".format(
+            cats, self.prior)
```
MechCoder: You should also change `self.prior` appropriately. (Not that we use it a lot)

skopt/space.py
```diff
@@ -10,6 +10,11 @@
 from sklearn.utils.fixes import sp_version
 
 
+class _Ellipsis:
```
MechCoder: Surely just printing out `...` is easier than the burden of maintaining a separate class? :p

skopt/space.py
```diff
@@ -324,6 +361,17 @@ def __init__(self, dimensions):
 
         self.dimensions = _dimensions
 
+    def __eq__(self, other):
+        return all([a == b for a, b in zip(self.dimensions, other.dimensions)])
+
+    def __repr__(self):
+        if len(self.dimensions) > 31:
```
MechCoder: `>=` 31?

MechCoder:
Just some minor comments. Thanks for the addition!

@betatim Please merge if you are happy.

skopt/space.py
nfcampos: ah yes, forgot about that

skopt/space.py
nfcampos: we show the first 3 and last 3; but if there are 7 I would just be replacing the middle element with `...` and taking up the same space. so I only truncate if there are more than 7. makes sense?

skopt/space.py
nfcampos: (see comment in the other diff)

skopt/space.py
MechCoder: I see, cool

skopt/space.py
```diff
@@ -158,25 +163,35 @@ def __init__(self, low, high, prior="uniform"):
...
+                np.log10(self.high) - np.log10(self.low))
             self.transformer = _Log10()
 
         else:
             raise ValueError(
                 "Prior should be either 'uniform' or 'log-uniform', "
                 "got '%s'." % self._rvs)
 
+    def __eq__(self, other):
+        return (type(self) is type(other)
```
MechCoder: Feel free not to change this, but we can just check type(other) == "whatever"

skopt/space.py
```diff
@@ -158,25 +163,35 @@ def __init__(self, low, high, prior="uniform"):
...
+                and self.low == other.low
+                and self.high == other.high
+                and self.prior == other.prior)
```
betatim: Worth using `np.allclose` to compare the prior which will be weird floating point numbers?

betatim: probably worth using everywhere we compare floats.

nfcampos: good point. is there an equivalent of `np.allclose` for scalars?

skopt/space.py
nfcampos: ah, so there's a reason i did this :)
if i create a list like `[1, 2, 3, '...', 5, 6, 7]` python stringifies it exactly like that, so we get those quote around the ellipsis which is definitely not what we want. The easiest way I could think of to get `[1, 2, 3, ..., 5, 6, 7]` without the quotes is for that ellipsis to be an object whose `__repr__` method returns `...`. Since the only way to create objects in python is to create a class I had to create a class to achieve that.
if you guys can think of a better way to achieve this I'm all ears

skopt/space.py
betatim: AFAIK `np.allclose([a], [b])` is the way 😀

skopt/space.py
MechCoder: That seems like a good solution. Please add a comment to prevent people like me from asking that question again. :-)

MechCoder:
@nfcampos Can you just address the inline comments and I'll happy to merge?

nfcampos:
@MechCoder I've addressed all comments now

skopt/space.py
```diff
@@ -246,13 +270,43 @@ def __init__(self, categories, prior=None):
         self.categories = categories
         self.transformer = _CategoricalEncoder()
         self.transformer.fit(self.categories)
+        self.prior = prior
 
         if prior is None:
             prior = np.tile(1. / len(self.categories), len(self.categories))
```
MechCoder: It should be easier to introduce a `self.prior_` attribute and equate that to the `np.tile(` here. In that way you can directly use `np.allclose` to compare `self.prior_` with `other.prior_` and do away with the `priors_are_equal` method.

glouppe:
Happy with those changes! Thanks a lot @nfcampos :)

nfcampos:
Thanks @glouppe @MechCoder and @betatim :)

MechCoder:
I addressed my final inline comment over here (https://github.com/scikit-optimize/scikit-optimize/pull/207)

PR #203, opened 2016-08-17
[WIP] Add TensorFlow MNIST example

ibab:
@betatim asked me whether I could make an example that optimizes the hyperparameters of a neural network using scikit-optimize.

I've written a short convolutional neural network that can classify handwritten digits from the MNIST dataset.
I've then exposed various hyperparameters like the number of layers, the dropout probability, etc. and used scikit-optimize to improve them.
Is everyone happy with the example?
I'll try to improve it a bit further, so this is only [WIP] for now.

### DIFF
--- a/examples/tensorflow-convnet.ipynb
+++ b/examples/tensorflow-convnet.ipynb
@@ -1,0 +1,228 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Optimizing convolutional neural networks with scikit-optimize\n",
+    "\n",
+    "The goal of this example is to train a classifier that can label handwritten digits from the MNIST dataset.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


MechCoder:
Thanks! I'll try to have a look over the weekend.

MechCoder:
Sorry about the delay, note that the default `GaussianProcessRegressor` in sklearn does not tune the kernel (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/gaussian_process/gpr.py#L155)   You can remove the `alpha` parameter since now the noise should also be tuned pending https://github.com/scikit-optimize/scikit-optimize/pull/225

MechCoder:
It might also be worth trying `forest_minimize` since it is claimed that it is highly beneficial when the number of categorical parameters are high.

What are your thought about integrating this? @betatim @glouppe

betatim:
I'll close this, we can re-open/think if @ibab comes back.

PR #207, opened 2016-08-23
Remove priors are equal function

MechCoder:


MechCoder:
Nothing great but the present solution just seems excessively verbose to me.

glouppe:
Feel free to merge.

Issue #32, opened 2016-03-29
Run examples as part of the CI


betatim:
Avoid broken examples like what happened in #29 by running them as part of travis. Not sure if there is anything more useful we can do than to check that they run with exit code == 0.

glouppe:
Shall we leave this open, provided examples will be eventually replaced with the notebooks?

betatim:
Yes. One way would be calling `nbconvert --execute` and then thinking about how to detect failure.

MechCoder:
Reopened since circleci could not catch the error in https://github.com/scikit-optimize/scikit-optimize/pull/209

MechCoder:
Fixed by https://github.com/scikit-optimize/scikit-optimize/commit/8d804453ea5a2499ef405ca13eff86c9312fa8eb

PR #208, opened 2016-08-23
Execute notebooks as part of build

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/32 

I'm not completely sure how the timeout in Circle Ci works, so just to check.

### DIFF
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -45,7 +45,7 @@ pip install pdoc==0.3.2 pygments
 
 # Generating documentation
 for nb in examples/*ipynb; do
-    jupyter nbconvert "$nb" --to markdown | tee -a ~/nb_to_md.txt
+    jupyter nbconvert --execute "$nb" --to markdown | tee -a ~/nb_to_md.txt
 done
 
 cd ~


MechCoder:
Merging for now. Will revert it if the website breaks.

betatim:
👍 

did you gain any insight on how we could upload notebooks somewhere/nbviewer them as part of a PR? Should we fail the CI if the notebooks raise warnings/exceptions?

MechCoder:
> Should we fail the CI if the notebooks raise warnings/exceptions?

We already do, I followed the sklearn approach of piping the output to stderr, and then grepping for tracebacks here (https://github.com/scikit-optimize/scikit-optimize/blob/master/circle.yml#L9). You can see the build failing here because I forgot to do an import. (https://circleci.com/gh/scikit-optimize/scikit-optimize/58). What we can do is modify the build script to fail early if there are exceptions. RIght now it checks all examples.

> did you gain any insight on how we could upload notebooks somewhere/nbviewer them as part of a PR?

No, I did not try that, but I like the idea. Feel free to open two issues.

PR #209, opened 2016-08-24
[MRG] Set noise tunable by default

MechCoder:
https://papers.nips.cc/paper/4522-practical-bayesian-optimization-of-machine-learning-algorithms.pdf advocates the use of setting noise as a tunable hyperparameter on the GP kernel.


skopt/gp_opt.py
```diff
@@ -194,12 +194,14 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
+                        nu=2.5)
+        noise = WhiteKernel()
         base_estimator = GaussianProcessRegressor(
-            kernel=(ConstantKernel(1.0, (0.01, 1000.0)) *
-                    Matern(length_scale=np.ones(space.transformed_n_dims),
-                           length_scale_bounds=[(0.01, 100)] * space.transformed_n_dims,
-                           nu=2.5)),
-            normalize_y=True, alpha=alpha, random_state=random_state)
+            kernel=cov_amplitude * matern + noise,
+            normalize_y=True, random_state=random_state)
```
glouppe: By default `alpha` in `GaussianProcessRegressor` is positive, which means noise can be controlled in two ways. Shall fix `alpha=0.0` instead here, since we have the `noise` component?

MechCoder:
I am yet to see the implications of this on any example but I'm unlikely to get there before tomorrow.

MechCoder:
For the first time, I'm happy to see a build fail...

skopt/gp_opt.py
MechCoder: Yes. But 10**-10 is still a small value no?. It is added to the covariance here (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/gaussian_process/gpr.py#L382)

MechCoder:
I think I broke something on sklearn master with this https://github.com/scikit-learn/scikit-learn/commit/c31ad7ada3d46f8b8119685b91d1562f8e710b09 :/ 

Could you check if you are still able to run the bayesian optimisation notebook? The repr breaks.

glouppe:
Meh... :/ I'll check later...

MechCoder:
Couldn't help it.

htismpc

MechCoder:
Minimal code to reproduce:

```python
matern = Matern(length_scale=np.ones(1))
ConstantKernel() * matern
```

MechCoder:
My gut is that it has something to do with this (https://github.com/scikit-learn/scikit-learn/pull/6963#discussion-diff-75063452) but I'll verify when I get to work.

MechCoder:
See: https://github.com/scikit-learn/scikit-learn/pull/7259 for a fix

MechCoder:
@glouppe I've pushed a temporary workaround and I've modified the example, which makes me think. should we keep alpha (maybe renamed to noise) with an option="auto"? It would be easier for people to play with their custom kernels without having to add a `WhiteKernel`

skopt/tests/test_gp_opt.py
```diff
@@ -26,6 +26,6 @@ def test_gp_minimize():
         yield (check_minimize, bench4, 0.0,
                [("-2", "-1", "0", "1", "2")], search, acq, 0.05, 10)
         yield (check_minimize, branin, 0.39, [(-5.0, 10), (0.0, 15.)],
-               search, acq, 0.1, 50)
```
MechCoder: I changed this because tests were failing.

glouppe:
Looks ok. Feel free to merge if green.

PR #212, opened 2016-08-31
Update Authors

MechCoder:


betatim:
Should we have a separate list or should we have just one list of authors? I have a slight preference for the latter. Overall looks good to me, so feel free to merge.

PR #213, opened 2016-09-01
[MRG] Time tests and report coverage

betatim:
This prints a list of the slowest tests and uploads the coverage (from py3.5) to codecov so we can actually look at it.

To see if we can speed up the tests we need to see what is slow, otherwise we will never think about it. As we already collect the coverage info it also makes sense to put it somewhere so we can see it -> we might start improving coverage.

* [x] exclude `skopt/tests/` from codecov report

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+coverage:
+  ignore:
+    - "skopt/tests/*"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ env:
 install: source build_tools/travis/install.sh
 
 script:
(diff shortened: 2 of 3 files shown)


MechCoder:
Yes, this will be highly useful. But where can we see the output? In the travis log?

betatim:
A summary of the coverage will be posted in the PR as a comment. The timing will be in the travis log.

betatim:
Bam! 87% coverage, not bad considering we never explicitly checked it!

MechCoder:
Awesome, thanks!

betatim:
I sneaked in a minor modification with  336f762

PR #214, opened 2016-09-01
[MRG+1] Add results obtained for Branin and Hart6

MechCoder:
An attempt to reproduce the extensive benchmarks described in http://www.automl.org/papers/13-BayesOpt_EmpiricalFoundation.pdf

TLDR; We are worse than Spearmint but better than others (for 2 benchmark functions)


MechCoder:
Note that this was before Spearmint implemented the beta transform and other improvements.

MechCoder:
Scripts to reproduce: https://github.com/MechCoder/Notebooks/blob/master/Function%20benchmarks.ipynb

betatim:
I'd store the results in a subdirectory.

Can we turn the notebook into a script that can live next to the results? Otherwise I fear we will "forget" how the numbers were produced or how to reproduce them or add more benchmarks.

MechCoder:
Sure, but I would prefer https://github.com/scikit-optimize/scikit-optimize/pull/215 to be merged before this. It exposes the time taken for each iteration to the user, so with that we can keep a track of the time as well and log that as well. WDYT?

MechCoder:
y@betatim Addressed your comments. Let me know if there is anything else.

MechCoder:
How about the other minimize functions. Do we still bench with `n_random_starts=1` despite knowing that this will default to random search until we cover the edges of the search space?

betatim:
> Do we still bench with n_random_starts=1 despite knowing that this will default to random search until we cover the edges of the search space?

why would we use `n_random_starts=1`? I'd use our default values as much as possible.

benchmarks/RESULTS.md
```diff
@@ -0,0 +1,10 @@
+These are the results got by optimizing the particular set of benchmark functions provided in Table 2 of
+(http://www.automl.org/papers/13-BayesOpt_EmpiricalFoundation.pdf) with our defualt values. All optimizers are run 10 times with `n_calls=200`.
+
+## gp_minimize
+
+
+|Blackbox Function| Valid loss | Best loss |
```
betatim: Valid loss? Isn't "average minimum" better? As well as "best minimum"?

MechCoder: I just followed the convention in the paper.

MechCoder:
Running benchmarks for `forest` and `gbrt`

MechCoder:
One quick look that I haven't done anything silly with the benchmark scripts and it should be good for merge. (Surprised that `forest_minimize` performs better on hart6)

benchmarks/RESULTS.md
betatim: Ah ok,  ¯\_(ツ)_/¯

betatim:
I'll take a look tomorrow morning.

MechCoder:
also opened an issue here (https://github.com/automl/HPOlib/issues/114)

betatim:
LGTM. If you wanted to you could rename `RESULTS.md` to `README.md` so it gets rendered automagically when you navigate to the directory. Merge with or without that change 😃

PR #215, opened 2016-09-02
[MRG] Log time in VerboseCallback and return callbacks in OptimizeResult instance

MechCoder:
- [x] Introduce attribute `total_time` in `VerboseCallback`
- [x] Allow `return_callbacks` argument that sets the attribute callbacks in the returned `OptimizeResult` instance.

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -41,12 +41,16 @@ class VerboseCallback(object):
     ----------
     * `iter_no`: [int]:
         Number of iterations of the optimization routine.
+
+    * `total_time`: [list]:
+        total_time[n_iter-1] gives the time of that particular iteration.
     """
 
     def __init__(self, n_total, n_init=0, n_random=0):
         self.n_init = n_init
         self.n_random = n_random
         self.n_total = n_total
+        self.total_time = []
         self.iter_no = 1
         self._start_time = time()
         self._print_info(start=True)
@@ -80,6 +84,7 @@ class VerboseCallback(object):
             The optimization as a OptimizeResult object.
         """
         time_taken = time() - self._start_time
+        self.total_time.append(time_taken)
         self._print_info(start=False)
 
         curr_y = res.func_vals[-1]
--- a/skopt/dummy_opt.py
+++ b/skopt/dummy_opt.py
@@ -16,7 +16,7 @@ from .utils import create_result
 
 def dummy_minimize(func, dimensions, n_calls=100,
                    x0=None, y0=None, random_state=None, verbose=False,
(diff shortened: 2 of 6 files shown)


MechCoder:
Is it useful to split the time taken for function evaluation and the time taken to fit the surrogate model? In some cases, it is useful when the time taken for a function evaluation completely varies (for example, while tuning regularization etc.)

Not sure how to do it with the current API though.

betatim:
Do we really need the `return_callbacks` argument? I'm not in love. My vote would be to always attach the info from the callbacks (if they provide any).

Should we attach the callbacks themselves or have a way of specifying which bit of info from the callbacks to add to the result? Not sure how easy that is to do technically but from a user perspective it would be much nicer to have `res.info.total_time` (maybe change `info` to another name). So we store additional info computed by the callbacks in the `info` attribute.

MechCoder:
Sounds reasonable. How about just `res.iter_time` instead of `res.total_time`?

betatim:
Think we should prefix it with something `res.name_of_callback.blah`?? The attribute should be named the same as the attribute from the callback.

How would you automatically pick which attributes of a callback to store? Should we adopt the convention that `self.attribute_` is stored but `self.otherattr` isn't? Or add a method to the callbacks to return "the summary to be attached" to each callback object?

glouppe:
> My vote would be to always attach the info from the callbacks (if they provide any).

+1

MechCoder:
I'm fine with just doing `res.name_of_the_callback.blah` with all the (public) attributes from the callback instance. Is that the consensus?

glouppe:
You cannot know the name of the callback in advance, since these may be user supplied. Just fill the list `res.callbacks`.

MechCoder:
Well, you can always do `callback.__class__.__name__` and such stuff. Taking a step back, what advantage does the user have in the optimization procedure returning all the callbacks? (I know I wrote it, but still)

glouppe:
I dont know. What was the use case you had in mind? why do you need to keep track of them? I am +0 for returning callbacks at all.

MechCoder:
I just wanted to get the time for every iteration.

MechCoder:
(Not the first time I've over - engineered something). I would be okay with just storing a `iter_time` attribute on the OptimizeResult instance if verbose is set to be True.

glouppe:
I am not convinced by adding anything at all to `OptimizeResult`. If you want to inspect attributes from a callback, then inspect that callback, not some other object.

MechCoder:
See https://github.com/scikit-optimize/scikit-optimize/pull/216

PR #216, opened 2016-09-02
[MRG] TimerCallback

MechCoder:


skopt/callbacks.py
```diff
@@ -91,3 +91,23 @@ def __call__(self, res):
         if self.iter_no <= self.n_total:
             self._print_info(start=True)
             self._start_time = time()
+
+
+class TimerCallback(object):
+    """
+    A timer callback that logs the time for every iteration.
```
betatim: "Log the elapsed time between each iteration of the optimization loop."?

betatim: How about:
```
"""Log the elapsed time between each iteration of the minimization loop.

The time for each iteration is stored in the `iter_time` attribute which you
can inspect after the minimization has completed.

$short_example??
"""
```

MechCoder:
We do not add class examples anywhere else, but apart from that addressed your comment and added a small note on `iter_time` attribute

betatim:
⌚

PR #217, opened 2016-09-05
cluttered around the minimum -> clustered around the minimum

andreh7:
'clutter' is actually a synonym of 'litter'

### DIFF
--- a/examples/bayesian-optimization.ipynb
+++ b/examples/bayesian-optimization.ipynb
@@ -428,7 +428,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Finally, as we increase the number of points, the GP model approaches the actual function. The final few points are cluttered around the minimum because the GP does not gain anything more by further exploration."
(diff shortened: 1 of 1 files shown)


betatim:
Thanks for this! Will have to find a minute later today to investigate how to get circleCI to rerun the test. Seems like it just timed out because running the examples took too long.

andreh7:
In principle, this is just a one letter change so it would be easy for you to do the modification yourself. But I guess the problem with circleCI timing out is worth solving anyway because it might affect more than just this pull request...

betatim:
As a rule we only merge things if travis and circle are green, independent of who proposes the change 😄. As you said, fixing the CircleCI timeout now is going to be worth it in the long term.

Thanks for your patience and also for making the effort of creating a PR for a small change!

glouppe:
All checks are now passing. Thanks @andreh7 !

betatim:
The magic of "retry build" :)

Issue #158, opened 2016-07-27
Refactor `*_minimize()` to reduce duplication


betatim:
There are large parts of shared/duplicated code and doc strings in all the different `*_minimize()` functions. The interface of the underlying surrogate models is (nearly) compatible. We should try and refactor things to avoid the duplication and bugs sneaking in because we have to do everything three times.

betatim:
A prime example is #157 where we have the same code several times. Sad 🐼

PR #219, opened 2016-09-08
[MRG] Use base_opt abstraction and refactor existing minimize functions

MechCoder:
Partial refactor of the entire module. Fixes https://github.com/scikit-optimize/scikit-optimize/issues/158

1. Refactor all existing minimize functions (apart from `dummy_opt`), such that only the base_estimator is reimplemented in different minimize functions.
2. Rename `search` to `acq_optimizer` and `acq` to `acq_func` for clarity.
3. Move `gbrt_minimize` to `gbrt_opt.py`
4. User should be able to implement custom optimization techniques just by implementing `base_estimator`


MechCoder:
cc: @betatim @glouppe

skopt/optimizer/__init__.py
```diff
@@ -0,0 +1,4 @@
+from . dummy_opt import dummy_minimize
+from . gbrt_opt import gbrt_minimize
+from . forest_opt import forest_minimize
+from . gp_opt import gp_minimize
```
betatim: What is the difference to `from .gpopt import gp_min` which we use in other places? #stylepolice

betatim:
Travis is unhappy because you missed one test that uses `dummy_opt`.

skopt/optimizer/base_opt.py
```diff
@@ -187,31 +161,18 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
         For more details related to the OptimizeResult object, refer
         http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
     """
-    # Save call args
     specs = {"args": copy.copy(inspect.currentframe().f_locals),
              "function": inspect.currentframe().f_code.co_name}
 
+    if not is_regressor(base_estimator):
+        raise ValueError("The base_estimator parameter has to either"
+                         " be a string or a regressor instance."
```
betatim: We don't support strings

betatim:
General thought: if all tests pass, (almost) doesn't change any public APIs -> merge?

A few comments:

* `skopt/optimizer/base_opt.py` -> `skopt/optimizer/base.py`? As well for the others.
* Should `base_opt` be public or private?
* I will make a few comments on the docs later today

skopt/optimizer/forest_opt.py
```diff
@@ -0,0 +1,166 @@
...
+        For more details related to the OptimizeResult object, refer
+        http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
+    """
+    rng = check_random_state(random_state)
+
+    # Default estimator
+    if isinstance(base_estimator, str):
+        if base_estimator not in ("rf", "et"):
+            raise ValueError(
+                "Valid values for the base_estimator parameter"
```
nfcampos: `Valid string values...`?

skopt/optimizer/__init__.py
MechCoder: Should we rename?

MechCoder:
I think `base_opt` should be kept public. Users can make use of the SMBO abstraction in by just re-implementing the `base_estimator`.

I have no strong opinion about others. Using `gp_min` etc will be similar to scipy's `fmin`, (and less verbose) but it breaks the public API.

skopt/optimizer/__init__.py
betatim: Misunderstanding, I was referring to the space between `.` and `gp_opt`.

betatim:
`base_opt` stays public then :) I think it is good to show how easy it is to plugin your own estimator.

> Using gp_min etc will be similar to scipy's fmin, (and less verbose) but it breaks the public API.

Not with you, can you elaborate please?

skopt/optimizer/__init__.py
MechCoder: ok then the difference is the space.

MechCoder:
Sorry, I was referring to `gp_min` in your comment over here (https://github.com/scikit-optimize/scikit-optimize/pull/219#discussion_r77953093)

MechCoder:
Moving the `_*minimize` functions to an `optimizer` submodule breaks `from skopt.gp_opt import gp_minimize`. Instead I would have to do `from skopt import gp_minimize` or `from skopt.optimizer.gp_opt import gp_minimize` :-/

MechCoder:
Whoops. The parameters are ordered differently in all the minimize functions. We should be be consistent. WDYT?

skopt/__init__.py
```diff
@@ -76,9 +76,6 @@ def f(x):
...
            "callbacks",
            "learning",
+           "optimizer",
            "plots",
-           "space",
-           "gp_minimize",
-           "dummy_minimize",
-           "forest_minimize",
-           "gbrt_minimize")
+           "space")
```
glouppe: Shouldnt we leave them here? We want `from skopt import gp_minimize` to work.

glouppe:
> Moving the _*minimize functions to an optimizer submodule breaks from skopt.gp_opt import gp_minimize. Instead I would have to do from skopt import gp_minimize or from skopt.optimizer.gp_opt import gp_minimize :-/

` from skopt import gp_minimize ` <-- this should be the import we advertise

skopt/__init__.py
MechCoder: Yes, I'll fix that. But I think `__all__` affects only what gets imported during `from skopt import *`. `from skopt import gp_minimize` should still work since I"ve imported from `optimizer` above. (http://stackoverflow.com/a/64130/1170730)

skopt/optimizer/__init__.py
betatim: 😀 can we remove it to satisfy my inner OCD? Pleeeaasse

betatim:
My comment here:

> `skopt/optimizer/base_opt.py` -> `skopt/optimizer/base.py`?

was just about renaming the file. It seems having optimizer and opt in the path is redundant.

MechCoder:
I've addressed the comments and made the arguments consistent. Anything else?

skopt/optimizer/forest_opt.py
```diff
@@ -0,0 +1,167 @@
+"""Forest based minimization algorithms."""
+
+from sklearn.utils import check_random_state
+
+from .base import base_minimize
+from ..learning import ExtraTreesRegressor
+from ..learning import RandomForestRegressor
+
+
+def forest_minimize(func, dimensions, base_estimator=None,
```
betatim: The old default was `base_estimator='et'` this is what is making travis unhappy. I think I prefer having `'et'` explicitly instead of using `None`. WDYT?

MechCoder:
Done. I would prefer a careful look before merging.

skopt/optimizer/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .dummy_opt import dummy_minimize
+from .forest_opt import forest_minimize
+from .gbrt_opt import gbrt_minimize
+from .gp_opt import gp_minimize
```
glouppe: Since those now live in the `optimizer` module, shouldnt we drop the `_opt` suffix?

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
 
     * `base_estimator` [a Gaussian process estimator]:
-        The Gaussian process estimator to use for optimization.
-        By default, a Matern kernel is used with the following
-        hyperparameters tuned.
-        - All the length scales of the Matern kernel.
-        - The covariance amplitude that each element is multiplied with.
-        - Noise that is added to the matern kernel. The noise is assumed
-          to be iid gaussian.
+        Should be a sklearn Regressor. In addition, should have an optional
```
glouppe: I would say that it should implement the scikit-learn `RegressorMixin` interface.

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
+    * `n_random_starts` [int, default=10]:
+        Number of evaluations of `func` with random initialization points
+        before approximating the `func` with `base_estimator`.
+
+    * `acq_func` [string, default=`"EI"`]:
+        Function to minimize over the posterior distribution. Can be either
 
         - `"LCB"` for lower confidence bound,
         - `"EI"` for expected improvement,
         - `"PI"` for probability of improvement.
```
glouppe: This way already like this, but I realize we should mention "(negative) expected improvement" and (negative) probability of improvement", since the docstring says we are minimizing those.

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
-    * `kappa` [float, default=1.96]:
-        Controls how much of the variance in the predicted values should be
-        taken into account. If set to be very high, then we are favouring
-        exploration over exploitation and vice versa.
-        Used when the acquisition is `"LCB"`.
-
-    * `search` [string, `"auto"`, `"sampling"` or `"lbfgs"`, default=`"auto"`]:
-        Searching for the next possible candidate to update the Gaussian prior
+    * `acq_optimizer` [string, `"auto"`, `"sampling"` or `"lbfgs"`, default=`"auto"`]:
+        Searching for the next possible candidate to update the prior
```
glouppe: This sentence is not very clear.

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
-        Number of points to sample to determine the next "best" point.
-        Useless if search is set to `"lbfgs"`.
-
-    * `n_random_starts` [int, default=10]:
-        Number of evaluations of `func` with random initialization points
-        before approximating the `func` with `base_estimator`.
-
-    * `n_restarts_optimizer` [int, default=10]:
-        The number of restarts of the optimizer when `search` is `"lbfgs"`.
+        If acq_optimizer is set to `"lbfgs"`, then a point is sampled randomly, and
```
glouppe: add backticks around `acq_optimizer` (here and everywhere else)

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
+        lbfgs is run for 5 iterations optimizing the acquisition function
```
glouppe: `maxiter` is set to `20` when we call LBFGS

skopt/tests/test_callbacks.py
```diff
@@ -3,7 +3,7 @@
 
 from skopt.benchmarks import bench1
 from skopt.callbacks import TimerCallback
-from skopt.dummy_opt import dummy_minimize
+from skopt.optimizer.dummy_opt import dummy_minimize
```
glouppe: `from skopt import dummy_minimize`

glouppe:
Thanks for the refactoring! This is indeed much better. I commented on a few minor things, but otherwise +1 once these are addressed.

MechCoder:
Good to go?

skopt/optimizer/base.py
```diff
@@ -82,60 +60,39 @@ def gp_minimize(func, dimensions, base_estimator=None, alpha=10e-10,
...
-        `"sampling"` for all other cases.
-
-        If search is set to `"sampling"`, `n_points` are sampled randomly
-        and the Gaussian Process prior is updated with the point that gives
-        the best acquisition value over the Gaussian prior.
-
-        If search is set to `"lbfgs"`, then a point is sampled randomly, and
-        lbfgs is run for 10 iterations optimizing the acquisition function
-        over the Gaussian prior.
+        Should inherit from `linear_model.base.RegressorMixin`.
```
betatim: Why not `sklearn.base.RegressorMixin`?

skopt/optimizer/base.py
```diff
@@ -75,24 +75,24 @@ def base_minimize(func, dimensions, base_estimator,
...
-        If acq_optimizer is set to `"sampling"`, `n_points` are sampled randomly
-        and the Gaussian Process prior is updated with the point that gives
-        the best acquisition value over the Gaussian prior.
-
-        If acq_optimizer is set to `"lbfgs"`, then a point is sampled randomly, and
-        lbfgs is run for 5 iterations optimizing the acquisition function
-        over the Gaussian posterior.
+        Method to minimize the acquistion function. The prior over `func`
+        is updated with the optimal value obtained by optimizing `acq_func`
+        with `acq_optimizer`.
```
betatim: For my education: we have the objective function `func` and the search space for the parameters. What is the prior?

MechCoder: Hmm. You are right, it might be better to say the `fit model`

MechCoder:
Will merge when tests pass.

betatim:
Yay! Getting this done is great and thank you for doing this somewhat tedious work 🚧 🎆

PR #222, opened 2016-09-12
[MRG] Handle warnings

### LINKED ISSUE #221: Suppress warning in auto-generated notebooks
https://scikit-optimize.github.io/notebooks/hyperparameter-optimization.html
https://scikit-optimize.github.io/notebooks/bayesian-optimization.html

Yuck :(

betatim:
Fixes #221

Remove deprecation warnings by changing our scoring parameter
names.

Try and handle the warning from matplotlib about building the
font cache.

### DIFF
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -49,7 +49,7 @@ python -c "import matplotlib.pyplot as plt"
 
 # Generating documentation
 for nb in examples/*ipynb; do
-    jupyter nbconvert --ExecutePreprocessor.timeout=900 --execute "$nb" --to markdown |& tee -a nb_to_md.txt
+    jupyter nbconvert --ExecutePreprocessor.timeout=1024 --execute "$nb" --to markdown |& tee -a nb_to_md.txt
 done
 
 cd ~
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -53,3 +53,4 @@ mkdir -p ./doc/skopt/notebooks
 cp ${SKOPT_HOME}/examples/*md ${HOME}/doc/skopt/notebooks
 cp -r ${SKOPT_HOME}/examples/*_files ${HOME}/doc/skopt/notebooks
 python ${SKOPT_HOME}/build_tools/circle/make_doc.py --overwrite --html --html-dir ./doc --template-dir ${SKOPT_HOME}/build_tools/circle/templates --notebook-dir ./doc/skopt/notebooks skopt
+cp -r ./doc ${CIRCLE_ARTIFACTS}
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -43,6 +43,10 @@ export SKOPT_HOME=$(pwd)
 conda install --yes jupyter
 pip install pdoc==0.3.2 pygments
 
+# importing matplotlib once builds the font caches. This avoids
+# having warnings in our example notebooks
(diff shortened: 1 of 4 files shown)


betatim:
Is there a way to see the output of the notebooks from the circle CI run?

Reproducing the matplotlib warning locally will be tricky as you only see if the very first time you import matplotlib.

betatim:
Voila: https://circleci.com/gh/scikit-optimize/scikit-optimize/100#artifacts/containers/0 how cool is that? If we made the links relative you could navigate the whole website for the build. Tres cool.

Unfortunately it seems to suggest that the warnings are still there. 😕

build_tools/circle/install.sh
```diff
@@ -43,6 +43,10 @@ export SKOPT_HOME=$(pwd)
 conda install --yes jupyter
 pip install pdoc==0.3.2 pygments
 
+# importing matplotlib once builds the font caches. This avoids
+# having warnings in our example notebooks
+python -c "import matplotlib.pyplot as plt"
```
glouppe: Clever! :elephant:

betatim: old 🐶 new trick 😄

betatim:
It works!

https://103-54340642-gh.circle-artifacts.com/0//tmp/circle-artifacts.6Kr9NjK/doc/skopt/notebooks/bayesian-optimization.html

glouppe:
Hmm, look at the last plot. What happened? bad initial random state?

betatim:
Good question. I can reproduce it locally.

```python
# Plot GP(x) + concours
gp = res.models[12]
y_pred, sigma = gp.predict(x, return_std=True)
```

gives a flat prediction, but `gp = res.models[11]` still works:
image

but already looks a bit weird. Something strange about the kernel we use?

betatim:
From looking at commits related to the kernel choice the only idea I have is #209 ?

MechCoder:
Uggh. not sure what I broke :/

MechCoder:
I figured out the bug. Thinking of an efficient way to fix it.

MechCoder:
The fix is more subtle, decribed as TODO item no 2 in https://github.com/scikit-optimize/scikit-optimize/pull/225. (and may or may not have a huge impact) However, the bottom line is that you need more samples such that the noise prediction becomes stabilized. A better function approximation is obtained after 50 iterations (or even before around 25). Here is a plot of the attribute `noise` from https://github.com/scikit-optimize/scikit-optimize/pull/225 . I am truly amazed at how the GP picks up the `noise` parameter!
noise

betatim:
I will try and make a summary to see if I understand what is happening: the kernel has a parameter that represents the amount of noise (the `np.random.randn() * noise_level` term in the toy model). The value of this parameter is plotted in your plot (is the x axis number of samples or number of models?). For the first few iterations there aren't enough samples to really set that parameter to a good value. Then at 15 it suddenly picks up on it (noise ~ 0.025) and by the time we get to ~55 iterations it is tuned to the "true" value.

I don't quite understand why it causes the prediction (mean of the GP) to become flat in the example though.

betatim:
Another question: should we merge this with the justification that this was already a problem before this PR and is being addressed in #225?

glouppe:
Yes we can merge this now

MechCoder:
@betatim Yes, that is right. My assumption is that the prediction is wrong because keeping the noise term (or the WhiteKernel) at prediction time gives an estimate of `y`, (that is f(x) with the noise) while we would like to model just `f(x)`

PR #225, opened 2016-09-16
[MRG] Add noise attribute to skopt.GaussianProcessRegressor

MechCoder:
- [x] Adds a noise attribute directly to `skopt.learning.GaussianProcessRegressor`
- [x] At prediction time, to predict `P(f(x) | y)` the kernel `K(X_train, X_new)` and `K(X_new, X_new)`, should not include the noise or white kernel argument, this has been described in Eq, 2.24 of http://www.gaussianprocess.org/gpml/chapters/RW2.pdf . Note that `K(X_train, X_new)` and `K(X_new, X_new)` do not have the `sigma**2` term.

### DIFF
--- a/skopt/learning/gpr.py
+++ b/skopt/learning/gpr.py
@@ -1,5 +1,16 @@
 from sklearn.gaussian_process import GaussianProcessRegressor as sk_GaussianProcessRegressor
 from sklearn.gaussian_process.kernels import WhiteKernel
+from sklearn.gaussian_process.kernels import Sum
+
+
+def _check_WhiteKernel_in_Sum(kernel):
+    if isinstance(kernel, Sum):
+        return (
+            _check_WhiteKernel_in_Sum(kernel.k1) or
+            _check_WhiteKernel_in_Sum(kernel.k2)
+        )
+    else:
+        return isinstance(kernel, WhiteKernel)
 
 
 class GaussianProcessRegressor(sk_GaussianProcessRegressor):
@@ -37,9 +48,9 @@ class GaussianProcessRegressor(sk_GaussianProcessRegressor):
         # http://www.gaussianprocess.org/gpml/chapters/RW2.pdf
         if isinstance(self.kernel_, WhiteKernel):
             self.kernel_.set_params(noise_level=0.0)
-        else:
+        elif _check_WhiteKernel_in_Sum(self.kernel_):
             for param, value in self.kernel_.get_params().items():
                 if isinstance(value, WhiteKernel):
                     self.kernel_.set_params(
-                        **{param:WhiteKernel(noise_level=0.0)})
+                        **{param: WhiteKernel(noise_level=0.0)})
         return self
--- a/skopt/learning/gpr.py
+++ b/skopt/learning/gpr.py
@@ -30,7 +30,10 @@ class GaussianProcessRegressor(sk_GaussianProcessRegressor):
         # The noise component of this kernel should be set to zero
         # while estimating K(X, X_test) and K(X_test, X_test)
         # Note that the term K(X, X) should include the noise but
-        # this (K(X, X))^-1y is precomputed as the attribute alpha.
+        # this (K(X, X))^-1y is precomputed as the attribute `alpha_`.
+        # (Notice the underscore).
+        # This has been described in Eq 2.24 of
+        # http://www.gaussianprocess.org/gpml/chapters/RW2.pdf
         if isinstance(self.kernel_, WhiteKernel):
             self.kernel_.set_params(noise_level=0.0)
         else:
--- a/skopt/learning/gpr.py
+++ b/skopt/learning/gpr.py
@@ -3,20 +3,40 @@ from sklearn.gaussian_process.kernels import WhiteKernel
 from sklearn.gaussian_process.kernels import Sum
 
 
-def _check_WhiteKernel_in_Sum(kernel):
+def _param_for_white_kernel_in_Sum(kernel, kernel_str=""):
+    """
+    Check if a WhiteKernel exists in a Sum Kernel
+    and if it does return the corresponding key in
+    `kernel.get_params()`
+    """
+    if kernel_str != "":
+        kernel_str = kernel_str + "__"
     if isinstance(kernel, Sum):
-        return (
-            _check_WhiteKernel_in_Sum(kernel.k1) or
-            _check_WhiteKernel_in_Sum(kernel.k2)
-        )
-    else:
-        return isinstance(kernel, WhiteKernel)
+        for param, child in kernel.get_params(deep=False).items():
+            if isinstance(child, WhiteKernel):
+                return True, kernel_str + param
+            else:
+                present, child_str = _param_for_white_kernel_in_Sum(
+                    child, kernel_str + param)
+                if present:
+                    return True, child_str
+    return False, "_"
 
 
 class GaussianProcessRegressor(sk_GaussianProcessRegressor):
     """
     GaussianProcessRegressor that allows noise tunability.
     """
+    def __init__(self, kernel=None, alpha=1e-10,
+                 optimizer="fmin_l_bfgs_b", n_restarts_optimizer=0,
+                 normalize_y=False, copy_X_train=True, random_state=None,
+                 noise=None):
+        self.noise = noise
+        super(GaussianProcessRegressor, self).__init__(
+            kernel=kernel, alpha=alpha, optimizer=optimizer,
+            n_restarts_optimizer=n_restarts_optimizer,
+            normalize_y=normalize_y, copy_X_train=copy_X_train,
+            random_state=random_state)
 
     def fit(self, X, y):
         """Fit Gaussian process regression model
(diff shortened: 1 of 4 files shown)


skopt/learning/gpr.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    GaussianProcessRegressor that allows noise tunability.
+    """
+
+    def fit(self, X, y):
+        super(GaussianProcessRegressor, self).fit(X, y)
+        for param, value in self.kernel_.get_params().items():
+            # XXX: Should return this only in the case where a
+            # WhiteKernel is added.
+            if param.endswith('noise_level'):
```
MechCoder: This seems a bit hacky to me. What do you think are the other alternatives?

MechCoder:
What are your thoughts about decoupling the noise argument from the `base_estimator` argument? This makes it explicit that I have added a `WhiteKernel` whose noise component I can set to zero while estimating `K(X_new, X_new)` and `K(X_train, X_new)`

betatim:
Where would you want to introduce the `noise=` keyword? As an argument to `gp_minimize`? As a naive user I think I'd prefer that either I give a kernel that I've designed myself and it gets used or that `skopt` uses its default. Having a parameter that then (maybe) modifies the kernel that I passed seems like it would make things more complicated.

Not knowledgable enough about this whole magic of brewing a kernel :(

cc @glouppe

glouppe:
or simply use the `alpha` parameter from the original scikit-learn interface? (instead of adding the `WhiteKernel` component)

MechCoder:
> or simply use the alpha parameter from the original scikit-learn interface?

We can do that but that assumes I need to know the amount of noise before hand right?

The `alpha` parameter does the right thing, but using a `WhiteKernel` means I need to set the noise component of the kernel to zero while doing the kernel product here (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/gaussian_process/gpr.py#L285) and here (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/gaussian_process/gpr.py#L298). What are some other ways to achieve that?

MechCoder:
Should be ready for reviews. I would consider this as a bug.

Before this PR (`n_calls=50`)

gp_before

After this PR (`n_calls=50`)

gp_after

skopt/learning/gpr.py
```diff
@@ -0,0 +1,41 @@
...
+        super(GaussianProcessRegressor, self).fit(X, y)
+        for param, value in self.kernel_.get_params().items():
+            if param.endswith('noise_level'):
+                self.noise_ = value
+                break
+
+        # The noise component of this kernel should be set to zero
+        # while estimating K(X, X_test) and K(X_test, X_test)
+        # Note that the term K(X, X) should include the noise but
+        # this (K(X, X))^-1y is precomputed as the attribute alpha.
```
glouppe: What `alpha` attribute are you referring to?

glouppe: Can you add the reference to Eqn. 2.24 in the comment?

glouppe:
Also, not convinced by the before/after figure. Both look fine to me.

MechCoder:
> but we should be aware this trick only holds for identically distributed noise, right?

Yeah, :/. I had added that point in the initial PR (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/gp.py#L69)

> Also, not convinced by the before/after figure. Both look fine to me.

Why do you think so, given a good number of function calls, the mean prediction should approximate the actual function right, and the first one seems a bit off?

skopt/learning/gpr.py
MechCoder: Sorry I meant this `alpha_` (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/gaussian_process/gpr.py#L229)

glouppe:
> Why do you think so, given a good number of function calls, the mean prediction should approximate the actual function right, and the first one seems a bit off?

This would be true for uniformly distributed samples, but this assumption does not hold in BO. We should only expect to be accurate in regions close of candidate optima. In the first figure, the region of high uncertainty is never ever sampled again because its LCB is larger than other parts of the input space. This is an expected behaviour of the algorithm.

MechCoder:
Yes, that's true. Sorry about the "noise"

MechCoder:
But I think the second cluster of points should be closer to the second local minima, but seems to be displaced by a bit in the first graph. But that just might be me nitpicking

skopt/learning/gpr.py
```diff
@@ -0,0 +1,41 @@
+from sklearn.gaussian_process import GaussianProcessRegressor as sk_GaussianProcessRegressor
+from sklearn.gaussian_process.kernels import WhiteKernel
+
```
MechCoder: done

MechCoder:
anything else?

betatim:
I'm happy, but don't really understand much of the GP mumbojumbo so deferring to @glouppe 😃

skopt/learning/gpr.py
```diff
@@ -0,0 +1,45 @@
...
+        # Note that the term K(X, X) should include the noise but
+        # this (K(X, X))^-1y is precomputed as the attribute `alpha_`.
+        # (Notice the underscore).
+        # This has been described in Eq 2.24 of
+        # http://www.gaussianprocess.org/gpml/chapters/RW2.pdf
+        if isinstance(self.kernel_, WhiteKernel):
+            self.kernel_.set_params(noise_level=0.0)
+        else:
+            for param, value in self.kernel_.get_params().items():
+                if isinstance(value, WhiteKernel):
```
glouppe: Isnt this only correct when `WhiteKernel` is a term in a sum kernel? (e.g. in the case where `kernel = K*WhiteKernel`, i dont believe this trick holds, but your if-statement will still be true)

MechCoder: I never saw that operations like `+` inherited from `Sum`, so I pushed a workaround that and added a test

skopt/learning/gpr.py
```diff
@@ -0,0 +1,56 @@
...
+        # (Notice the underscore).
+        # This has been described in Eq 2.24 of
+        # http://www.gaussianprocess.org/gpml/chapters/RW2.pdf
+        if isinstance(self.kernel_, WhiteKernel):
+            self.kernel_.set_params(noise_level=0.0)
+        elif _check_WhiteKernel_in_Sum(self.kernel_):
+            for param, value in self.kernel_.get_params().items():
+                if isinstance(value, WhiteKernel):
+                    self.kernel_.set_params(
+                        **{param: WhiteKernel(noise_level=0.0)})
```
glouppe: Sorry for nitpicking again, but this still does not solve completely the issue. If you have `kernel = K*WhiteKernel + WhiteKernel`, you would change both of them. 

The cleanest solution might be to instantiate the kernel yourself in this new class, adding in the `WhiteKernel` term to the user-provided kernel, and then keep a reference to this term.

MechCoder: > adding in the WhiteKernel term to the user-provided kernel,

This goes against the principles of sklearn to not set logic at init time, right?

glouppe: Not if you have something like:

```
def __init__(self, ..., kernel, ...):
    self.kernel = kernel

def fit(self, X, y):
    self.noise_ = WhiteKernel()
    self.kernel_ = self.kernel + self.noise_
    ...
    self.gp_ = sklearn.GaussianProcessRegressor(self.kernel_).fit(X, y)
    ...
    self.noise_.set_params(noise_level=0.0)
    ...
```

MechCoder: But you wouldn't always want to set the "noise" option on, no? Which is why I suggested the `noise="auto"` option to suggest if we wanted to add the `WhiteKernel` or not.

MechCoder: `noise="auto"` -> `noise="gaussian"` (to be more accurate)

MechCoder: fixing this now

MechCoder:
@glouppe Your solution will not work because the provided kernel is not modified at fit time, but rather a clone of it. I have pushed another fix (with tests). Let me know what you think.

skopt/learning/gpr.py
```diff
@@ -0,0 +1,193 @@
...
+    alpha_: array-like, shape = (n_samples,)
+        Dual coefficients of training data points in kernel space
+
+    log_marginal_likelihood_value_: float
+        The log-marginal-likelihood of ``self.kernel_.theta``
+    """
+    def __init__(self, kernel=None, alpha=1e-10,
+                 optimizer="fmin_l_bfgs_b", n_restarts_optimizer=0,
+                 normalize_y=False, copy_X_train=True, random_state=None,
+                 noise=None):
```
glouppe: So by default, `noise=None`, which amounts to reuse scikit-learn's GaussianProcessRegressor. Sorry to be blunt, but in the end, what is the added advantage of this PR? simply providing a shortcut for adding a `WhiteKernel` component? i.e. `noise="gaussian"` versus `kernel=K+WhiteKernel()`?

glouppe: ... oh no, forget that comment. It also deals with disabling that `WhiteKernel` part when making predictions. Accordingly, it might be good to add a test highlighting the difference between `skopt.GP(kernel=K, noise="gaussian").predict` and `sklearn.GP(kernel=K+WhiteKernel()).predict`.

glouppe:
LGTM!

MechCoder:
@glouppe Anything else?

MechCoder:
Phew, this took more time then I had expected.

MechCoder:
squashed and mergred

betatim:
:shipit: good job!

PR #228, opened 2016-09-18
Fix PyQt4 import error

MechCoder:
Travis on master creates a `PyQt4` import error when trying to import matplotlib.pyplot. Not sure what the issue is but checking if this fixes the issue.

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -26,7 +26,7 @@ popd
 # Configure the conda environment and put it in the path using the
 # provided versions
 conda create -n testenv --yes python=$PYTHON_VERSION pip nose \
-   numpy scipy cython matplotlib
+   numpy scipy cython pyqt matplotlib
 source activate testenv
 pip install git+http://github.com/scikit-learn/scikit-learn.git
 


betatim:
It seems wrong that we'd have to add `PyQT` to our list of dependencies. Do you know if there was a recent release of `matplotlib`? Trying to work out what change caused this to break.

betatim:
Even more weird is that it seems to work on CircleCI??

betatim:
[Last successful build](https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/160425499) and [failing build](https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/160767690) can anyone spot the difference? Seems like both install the same `matplotlib` version. Does the backend get selected via some environment variable?

betatim:
We used to install `qt-4.8.7-4` but now `conda` installs `qt-5.6.0-0`. So at some point the default switched from qt4 to qt5 and I guess `matplotlib` hasn't caught up with that yet. Same on CircleCI but for some reason the import error there doesn't fail the build.

Will start a new PR where the qt version is pinned to qt4.

PR #229, opened 2016-09-19
[MRG] Pin PyQT to 4.11 temporarily

betatim:
This is a takeover from #228. After some digging I found ContinuumIO/anaconda-issues#1068. Seems like @continuumio broke something so waiting for them to fix it after which we should undo this pinning.


betatim:
Looks like this is working. Merge when the bots are happy?

glouppe:
+1

MechCoder:
Thanks!

PR #233, opened 2016-09-22
Updated .gitignore

mp4096:
These are very small changes:
* ignore `build` folder (after `python setup.py install`)
* ignore Jupyter Notebook checkpoints
* ignore the new egg folder


.gitignore
```diff
@@ -1,6 +1,9 @@
 *.py[co]
 *~
+examples/.ipynb_checkpoints/
 # created by installing scikit-learn from git
 src/*
+build/
 dist/
 skopt.egg-info/
+scikit_optimize.egg-info/
```
betatim: does this mean we don't need `skopt.egg-info/` anymore? I think so.

.gitignore
```diff
@@ -1,6 +1,9 @@
 *.py[co]
 *~
+examples/.ipynb_checkpoints/
```
betatim: Would something like `*/.ipynb_checkpoints/` match both checkpoint directories in the `example/` folder and the top level? That would be sweet, mainly because I am messy and start `jupyter` in either directory :-/

mp4096: Yep, makes sense.

.gitignore
mp4096: I took the liberty to ignore all `.egg-info`-folders. :smiley_cat:

betatim:
Thanks for this and congrats on your first contribution to scikit-optimize!

mp4096:
Thank you! A very compact one too. :smile:

But I'm going to use `skopt` more intensively now, so let's see what happens along the way.

PR #234, opened 2016-09-23
[MRG] Add a ask-and-tell interface

betatim:
This refactors a large part of the base_minimize function to
use the new Optimizer object. Paving the way to standalone
use of the Optimizer object for a ask-and-tell interface.

Below is the old message for this PR which is now totally out of date.

---

Setting out the API for an minimization interface that
gives control over the iterations and how to evaluate
the objective to the user.

What are your thoughts on the interface? How should it work?

Notes:
- implement `suggest()`ing several points later
- explicit `suggest() -> report() -> fit() -> suggest()` or `suggest() -> report()  -> suggest()` what should the user have to do? (the difference is that in the second case `fit()` is called for the user)
- reuse as much of the `*_minimize()` machinery as possible even if it isn't as computational efficient as it could be


betatim:
This PR is working on #68. I will start working on this again. Will try out the API @glouppe [sketched](https://github.com/scikit-optimize/scikit-optimize/issues/68#issuecomment-214146104).

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,141 @@
...
+from sklearn.utils import check_random_state
+
+from ..acquisition import _gaussian_acquisition
+from ..acquisition import gaussian_acquisition_1D
+from ..space import Space
+from ..utils import create_result
+
+
+class Optimizer(object):
+    """docstring for Optimizer."""
```
glouppe: :)

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,141 @@
...
+                 xi=0.01, kappa=1.96, n_jobs=1):
+        super(Optimizer, self).__init__()
+        self._check_arguments(func, dimensions, base_estimator,
+                              n_random_starts,
+                              acq_func, acq_optimizer,
+                              verbose,
+                              n_restarts_optimizer,
+                              n_jobs)
+        self.specs = {"args": copy.copy(inspect.currentframe().f_locals),
+                      "function": inspect.currentframe().f_code.co_name}
```
glouppe: This is not correct anymore

betatim:
Where should we do the initialisation of arguments? In `__init__` and `_check_arguments` or for some reason delay it like in `sklearn`?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,141 @@
...
+        if len(self.Xi) < self.n_random_starts:
+            return self.space.rvs(random_state=self.rng)[0]
+
+        else:
+            transformed_bounds = np.array(self.space.transformed_bounds)
+            est = clone(self.base_estimator)
+
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                est.fit(self.space.transform(self.Xi), self.yi)
```
glouppe: (We can consider the gp model as being part of the state of this object)

glouppe:
I like it! Tell me when you are ready for a second round of review.

glouppe:
Any update on this? Flagged as item for 0.3

betatim:
I will try and work on this over the weekend.

jnothman:
Is ask-and-tell more-or-less the same as Active Learning?

betatim:
😄 

No this is just boring software engineering. Right now `X_optimize()` controls the `for` loop. Which can be annoying if the function you are evaluating is "print out value, send someone to the lab to perform experiment, wait a few days, type result in, next iteration". Ask-and-tell gives the user the job of controlling the loop.

skopt/optimizer/base.py
```diff
@@ -188,136 +175,78 @@ def base_minimize(func, dimensions, base_estimator,
...
-
-    # Random function evaluations.
-    X_rand = space.rvs(n_samples=n_random_starts, random_state=rng)
-    Xi = x0 + X_rand
-    yi = y0
 
-    for i, x in enumerate(X_rand):
-        yi.append(func(x))
+        for i, (x, y) in enumerate(zip(x0, y0)):
+            fit_model = i == len(x0) - 1
```
betatim: Is there a better way to do something on the last iteration of a loop? Should we add a `fit()` to `Optimizer` to handle this case (not sure I like that idea, too many ways of calling ask, tell and fit in the wrong order)?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+                a = cand_acqs[best_ind]
+
+                if a < best:
+                    next_x, best = cand_xs[best_ind], a
+
+            # lbfgs should handle this but just in case there are
+            # precision errors.
+            next_x = np.clip(
+                next_x, transformed_bounds[:, 0], transformed_bounds[:, 1])
+            # XXX why the need for [0] at the end?
```
glouppe: `inverse_transform` operates at the level of an array of points, while here we only want to transform one point. That is why we reshape and then select element 0.

skopt/tests/test_common.py
```diff
@@ -81,22 +83,23 @@ def test_minimizer_api():
...
+        yield (check_minimizer_api, result, n_calls)
+        yield (check_minimizer_bounds, result, n_calls)
         assert_raise_message(ValueError,
                              "return a scalar",
                              dummy_minimize, lambda x: x, [[-5, 10]])
 
         n_calls = 7
         n_random_starts = 3
-        n_models = n_calls - n_random_starts
+        n_models = n_calls - n_random_starts + 1
```
glouppe: I dont understand the `+1`.

Say `n_random_starts=1` and `n_calls=1`. Why would we build 2 models?

betatim:
This needs some more unit testing in `Optimizer`

betatim:
Will punt on adding more tests for parameter options for now and create an issue instead. The longer this goes unmerged, the higher the chance of screwing something up when rebasing :-/

@MechCoder could you double check that I moved your changes in base.py correctly?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,242 @@
...
+        self.acq_func = acq_func
+        self.rng = check_random_state(random_state)
+        self.kappa = kappa
+        self.models = []
+        self.n_points = n_points
+        self.n_restarts_optimizer = n_restarts_optimizer
+        self.space = Space(dimensions)
+        self.Xi = []
+        self.xi = xi
+        self.yi = []
```
glouppe: Shall we group and document the computed attributes? (`models`, `Xi`, `yi`, `space`)?

skopt/tests/test_optimizer.py
```diff
@@ -0,0 +1,21 @@
...
+    # calling ask() multiple times without a tell() inbetween should
+    # be a "no op"
+    base_estimator = ExtraTreesRegressor(random_state=2)
+    opt = Optimizer([(-2.0, 2.0)], base_estimator, n_random_starts=1,
+                    acq_optimizer="sampling")
+
+    opt.run(bench1, n_iter=3)
+    # tell() computes the next point ready for the next call to ask()
+    # hence there are three after three iterations
+    assert_equal(len(opt.models), 3)
```
glouppe: While it does not make a difference here, shouldnt we instead evaluate the length of `opt.Xi`?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,242 @@
...
+
+    * `n_jobs` [int, default=1]
+        Number of cores to run in parallel while running the lbfgs
+        optimizations over the acquisition function. Valid only when
+        `acq_optimizer` is set to "lbfgs."
+        Defaults to 1 core. If `n_jobs=-1`, then number of jobs is set
+        to number of cores.
+    """
+    def __init__(self, dimensions, base_estimator,
+                 n_random_starts=10, acq_func="EI", acq_optimizer="auto",
```
glouppe: Wasnt `"auto"` removed?

glouppe: and `acq_optimizer` set to `"lbfgs"` by default. 

See #285

glouppe:
Besides my comments, this looks good to me. Thanks for working on this, this is indeed much cleaner now :)

Manoj should definitely review this properly though before we hit the merge button.

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,242 @@
...
+        self.Xi = []
+        self.xi = xi
+        self.yi = []
+
+        self._check_arguments(dimensions, base_estimator,
+                              n_random_starts,
+                              acq_func, acq_optimizer,
+                              n_jobs)
+
+        self.parallel = Parallel(n_jobs=n_jobs)
```
MechCoder: My opinion is that it might be better to do the first `n_random_starts` calls at init time (and in parallel). This would make the first call to `ask` intelligent . Thoughts?

MechCoder: This would clean up some code, for instance `ask` becomes just

```python
def ask():
    return self._next_x
```

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,242 @@
...
+        """Record an observation of the objective function.
+
+        Provide values of the objective function at points suggested by `ask()`
+        or other points. By default a new model will be fit to all
+        observations. The new model is used to suggest the next point at
+        which to evluate the objective. This point can be retrieved by calling
+        `ask()`.
+
+        To add several observations as a batch without fitting a new model
+        set `fit_model` to False.
```
betatim: The use case I had in mind was one where you know you don't want to fit yet, but don't have a batch (say the user is entering them by hand). Mostly a question of where you do the queuing I guess. You decide, I have no strong feelings.

skopt/optimizer/optimizer.py
betatim: Currently `Optimizer` does not have access to the objective function, so that the caller has control over the loop. If we wanted to compute the objective for the random points in batch we'd have to pass the objective func and the first few points would be treated 'specially' compared to the rest from the users perspective.

skopt/tests/test_optimizer.py
betatim: Done.

skopt/optimizer/optimizer.py
MechCoder: Okay is the `model` in `fit_model` redundant?

MechCoder: that is, rename `fit_model` to just fit?

skopt/optimizer/optimizer.py
glouppe: I would prefer keep things separated as Tim proposed in this PR. Evaluating the function within this class `Optimizer` would make the ask and tell interface inconsistent in my opinion (i.e., if the `Optimizer` object had access to the objective function, then there would be no point in having a `tell` method at all.)

skopt/optimizer/optimizer.py
glouppe: or `refit`, as inspired from `GridSearchCV`?

betatim: I like `fit`, changed.

skopt/optimizer/base.py
```diff
@@ -186,129 +172,81 @@ def base_minimize(func, dimensions, base_estimator,
...
-        Xi.append(next_x)
-        curr_res = create_result(Xi, yi, space, rng, specs, models)
-        for c in callbacks:
-            if callbacks:
-                c(curr_res)
+    for n in range(n_calls - n_init_func_calls):
+        # fit model after last random iteration
+        fit_model = n >= n_random_starts - 1
+        next_x = optimizer.ask()
+        optimizer.tell(next_x, func(next_x), fit_model=fit_model)
```
MechCoder: I think we can do away with `n_random_starts` in the `Optimizer`?

If we want to random start n times we can simply call `tell` `n` times with `fit_model` set to False?

skopt/optimizer/base.py
```diff
@@ -186,129 +172,81 @@ def base_minimize(func, dimensions, base_estimator,
...
-    else:
-        y0 = []
 
-    # Random function evaluations.
-    X_rand = space.rvs(n_samples=n_random_starts, random_state=rng)
-    Xi = x0 + X_rand
-    yi = y0
+        for i, (x, y) in enumerate(zip(x0, y0)):
+            fit_model = i == len(x0) - 1
+            optimizer.tell(x, y, fit_model=fit_model)
```
betatim: if you provide `x0`, `y0` and `n_random_starts` we add the observations you provide, and then do an additional `n_random_starts`. The case where just `x0` is provided (but not `y0` is handled on line 193.

skopt/optimizer/base.py
betatim: I was thinking that `Optimizer` should always fit the model when provided a new point. If you don't want it to do that, you need to tell it. Hence this design where it is the caller's job to tell `Optimizer` to not fit these points until we are on the last random point in which case we do want a fit.

If you remove `n_random_starts` then `ask()` won't know what to suggest to you no?

It is true that it is a bit weird to tell `Optimizer` that you want random points and then assume it doesn't know if you want it to fit or not while suggesting random points.

`fit_model = n >= n_random_starts - 1` looks a bit horrible because we need it to be False for all iterations except for the last one.

betatim:
* [ ] understand why `bench2` benchmark changes.

skopt/tests/test_forest_opt.py
```diff
@@ -54,7 +54,7 @@ def test_tree_based_minimize():
         # max(sampled_points) would give a constant value.
         X0 = [[-5.6], [-5.8], [5.8], [5.6]]
         yield (check_minimize, minimizer, bench2, -5,
-               [(-6.0, 6.0)], 0.1, 100, 10, X0)
+               [(-6.0, 6.0)], 0.4, 100, 10, X0)
```
betatim: TBH no idea really. I've added a todo for it.

skopt/optimizer/base.py
```diff
@@ -186,129 +172,81 @@ def base_minimize(func, dimensions, base_estimator,
...
-    if y0 is None and x0:
-        y0 = []
+    # User suggested points at which to evaluate the objective first
+    if x0 and y0 is None:
         for i, x in enumerate(x0):
-            y0.append(func(x))
-            curr_res = create_result(x0[:i + 1], y0, space, rng, specs)
+            fit_model = i == len(x0) - 1
+            y = func(x)
+            optimizer.tell(x, y, fit=fit_model)
```
MechCoder: It seems to me that when `x0`, and `n_random_starts` are present, the first model is fit after the `len(x0)` calls to the function but before it was `len(x0) + n_random_starts` no?

That is what I had meant by "behaviour change"

betatim: :zap: you are right!

betatim: Test added

MechCoder:
> I was thinking that Optimizer should always fit the model when provided a new point. If you don't want it to do that, you need to tell it. Hence this design where it is the caller's job to tell Optimizer to not fit these points until we are on the last random point in which case we do want a fit.

I agree, but it is very awkward that I need to supply both `n_random_starts` and that the `Optimizer` is unable to infer that fitting the first model should be delayed till the first `n_random_starts` points are provided, but I'll have to think a bit more to give a suitable suggestion.

> fit_model = n >= n_random_starts - 1 looks a bit horrible because we need it to be False for all iterations except for the last one.

Actually, you can set it to be True for all iterations as well (assuming there is no additional expense in fitting the base regressor), and `ask` will still return random points till it gets `n_random_starts` points. Was this the intention?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,248 @@
...
+        function.
+    * `space`
+        An instance of `skopt.space.Space`. Stores parameter search space used
+        to sample points, bounds, and type of parameters.
+
+    """
+    def __init__(self, dimensions, base_estimator,
+                 n_random_starts=10, acq_func="EI", acq_optimizer="lbfgs",
+                 random_state=None, n_points=10000, n_restarts_optimizer=5,
+                 xi=0.01, kappa=1.96, n_jobs=1):
```
glouppe: +1, that's a good idea

MechCoder:
I have to look one last time but the API looks great. The major nagging point is explicit supplying `fit=False` even when `n_random_starts` is provided. One way to handle this is to have a `fit="auto"` option that does not fit the model till `ask` is called `n_random_starts` times. WDYT?

betatim:
> Actually, you can set it to be True for all iterations as well (assuming there is no additional expense in fitting the base regressor), and ask will still return random points till it gets n_random_starts points. Was this the intention?

Yes, you could set `fit=True` always and it would not change the behaviour regarding random points. I think of setting it to `False` during the `n_random_starts` phase as a speed optimisation. The caller of `tell()` knows that they are doing stuff during which they don't want the estimator to be fitted.

We could make `Optimizer` never fit while returning random points/figure out by itself if it should fit or not.

> I agree, but it is very awkward that I need to supply both n_random_starts and that the Optimizer is unable to infer that fitting the first model should be delayed till the first n_random_starts points are provided, but I'll have to think a bit more to give a suitable suggestion.

The more I think about it, the more I am starting to think that `Optimizer` should decide when to start fitting. If the caller dislikes the behaviour they should just set `n_random_starts=0` and do their own thing. I would keep the `fit` argument though to allow the caller to set the behaviour of fitting or not (I think even if we added a way to provide points in a batch you might want to have several batches without fit in between).

-1 on an 'auto' value, somehow it is easier to make the logic explicit/give it to a human then try and write nice code that implements 'auto'.

skopt/optimizer/optimizer.py
betatim: New PR? (please)

glouppe: I personally dont mind delaying this for another PR.

MechCoder:
> We could make Optimizer never fit while returning random points/figure out by itself if it should fit or not.

I would be +1 on doing that but how do you suggest to do it?

skopt/tests/test_forest_opt.py
MechCoder: This works for me now with 0.1, locally.

betatim: Same here, maybe related to the bugs you pointed out before regarding change in behaviour

betatim:
> I would be +1 on doing that but how do you suggest to do it (in terms of API)?

I'd say if the `n_random_starts` aren't used up yet then it won't fit. Also provide the parameter `fit=True` so that people can explicitly prevent fitting from happening.

MechCoder:
> I'd say if the n_random_starts aren't used up yet then it won't fit.

I'm sorry but I do not understand. That is not the current behaviour right? I have to set `fit=False` explicitly as you have done in `base.py`

betatim:
Correct, it isn't the current behaviour. It was my suggestion for a better/new API.

MechCoder:
Ah okay, cool we are on the same page then :)

betatim:
This adds support for passing batches of values to `tell()`. Need this as `Optimizer` has no way of knowing whether you are passing in one of many precomputed points (in which case it should not fit) or if this is a single point. When `n_random_starts=0` it would fit after each call to `tell`.

This means the callbacks are not called after each point, but only after the batch has been added. Do we care?

betatim:
Also detecting a batch vs a objective function that returns something that isn't a scalar is getting a bit ugly.

skopt/optimizer/optimizer.py
betatim: -> #294

betatim:
Implemented @MechCoder's suggestions and did some clean up of the code. Will it be ready this time? :smile:

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,263 @@
...
+                          "removed in 0.4.")
+            acq_optimizer = "lbfgs"
+        self.acq_optimizer = acq_optimizer
+        if self.acq_optimizer not in ["lbfgs", "sampling"]:
+            raise ValueError(
+                "Expected acq_optimizer to be 'lbfgs' or 'sampling', "
+                "got %s" % acq_optimizer)
+
+    def ask(self):
+        """Suggest next point at which to evaluate the objective."""
```
MechCoder: I suppose this deserves an extra line of documentation, on the lines of "this returns a random point for the first `n_random_starts`, after which it suggests a point based on the underlying base estimator?"

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,263 @@
...
+
+        else:
+            if not self.models:
+                raise ValueError("Random evaluations exhausted and no "
+                                 "model has been fit.")
+            # return point computed from last call to tell()
+            return self._next_x
+
+    def tell(self, x, y, fit=True):
+        """Record an observation of the objective function.
```
MechCoder: This is no longer "an observation" and should we add something like "of the objective function to suggest the next candidate point"?

skopt/tests/test_common.py
MechCoder: Seems we can avoid this by doing `fit=False` at the last iteration (which is the current behaviour)?

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,263 @@
...
+        To add observations in a batch pass a list-of-lists for `x` and a list
+        of scalars for `y`.
+
+        * `x` [list or list-of-lists]:
+            Point at which objective was evaluated.
+        * `y` [scalar or list]:
+            Value of objective at `x`.
+        * `fit` [bool, default=True]
+            Fit a model to observed evaluations of the objective. A model will
+            only be fitted after `n_random_starts` points have been queried.
```
MechCoder: I would add "irrespective of the value given to fit"

skopt/optimizer/optimizer.py
betatim: I prefer this, where the fitting happens feels like an implementation detail :-/

skopt/optimizer/optimizer.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        # if y isn't a scalar it means we have been handed a batch of points
+        if (isinstance(y, Iterable) and all(isinstance(point, Iterable)
+                                            for point in x)):
+            self.Xi.extend(x)
+            self.yi.extend(y)
+
+        else:
```
betatim: Fixed

MechCoder:
LGTM. @glouppe would you like to have another look before merging?

glouppe:
Just had another look. Looks good to me too. Merging this beast! Great work Tim :) :beers:

betatim:
Woah! Thanks @MechCoder for all the patience and suggestions! :champagne:

Issue #230, opened 2016-09-19
Unpin PyQT version for CI builds


betatim:
Undo #229 once ContinuumIO/anaconda-issues#1068 has been fixed.

Once the bug in conda has been fixed we should stop pinning the version of pyqt that we use in the CI builds.

glouppe:
Fixed by #278.

PR #235, opened 2016-09-23
Yet more pinning of QT versions

betatim:
See ContinuumIO/anaconda-issues#1090

This PR will also need undoing as part of #230

### DIFF
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -30,7 +30,7 @@ popd
 # Configure the conda environment and put it in the path using the
 # provided versions
 conda create -n testenv --yes python pip nose \
-   numpy scipy cython matplotlib pyqt==4.11.4
+   numpy scipy cython matplotlib pyqt=4.11.4=3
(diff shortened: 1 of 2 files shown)


MechCoder:
We pip install from sklearn master, right? Any idea why it searches for `sklearn-0.18 dev`?

betatim:
Potentially because of the #sklearn-0.18dev in the URL (I think). I am a bit frustrated with conda atm. Looks like we need to pin to different builds for py27 and py3, so this is getting pretty tedious.

Was wondering if we could just use `pip` for everything instead? Seems like there are `manylinux` wheels for lots of packages these days.

MechCoder:
I restarted the build on master and it seemed to pass

betatim:
Looks like ContinuumIO/anaconda-issues#1090 is fixed which should make most of this obsolete and in ContinuumIO/anaconda-issues#1068 I see a lot of people unpinning QT as well. So maybe that is fixed as well. Closing this.

PR #236, opened 2016-09-25
Add benchmarks based on the SurrogateBenchmarks library

MechCoder:
This Pull Request is unlikely to get merged, but I thought I'll add it as a token of my efforts. The problem is that the pickled `SurrogateBenchmarks` model depends on sklearn 0.15 and using a newer vesion of `sklearn` breaks the pickled model because of some new attributes added to the tree module.

(And of course, we have to a have a newer version of sklearn because of GPR etc)

### DIFF
--- a/benchmarks/bench_surrogate_logreg_withcv.py
+++ b/benchmarks/bench_surrogate_logreg_withcv.py
@@ -1,0 +1,48 @@
+# We are a big fan of legacy python.
+
+import cPickle
+import os
+import tarfile
+import urllib
+
+import numpy as np
+import Surrogates
+
+from skopt import gp_minimize
+
+model_loc_path = "logreg/cv/models/ENCODED_logreg_cv_all_RandomForest"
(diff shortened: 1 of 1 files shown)


betatim:
Is it worth us importing the models with an old version, storing them as text, re-pickling with python3? Or are the models fundamentally incompatible because of the different parameters?

MechCoder:
> Or are the models fundamentally incompatible because of the different parameters?

Yes. Pickling in a newer version gives the following error because I suppose the `max_depth` parameter was added post 0.15

```python
/home/manoj/scikit-optimize/src/sklearn/sklearn/tree/_tree.pyx in sklearn.tree._tree.Tree.__setstate__ (sklearn/tree/_tree.c:7860)()
    631     def __setstate__(self, d):
    632         """Setstate re-implementation, for unpickling."""
--> 633         self.max_depth = d["max_depth"]
    634         self.node_count = d["node_count"]
    635 

KeyError: 'max_depth'
```

PR #237, opened 2016-09-27
[MRG] Transform GP-optimizer space to [0, 1]

MechCoder:
Also,

1. Created a Pipeline transformer
2. Added `HyperCube` transformer to the list of transformers.

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -39,7 +39,7 @@ def base_minimize(func, dimensions, base_estimator,
                   acq_func="EI", acq_optimizer="auto",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96):
+                  xi=0.01, kappa=1.96, space=None):
     """
     Parameters
     ----------
@@ -170,7 +170,9 @@ def base_minimize(func, dimensions, base_estimator,
 
     # Check params
     rng = check_random_state(random_state)
-    space = Space(dimensions)
+
+    if space is None:
+        space = Space(dimensions)
 
     # Initialize with provided points (x0 and y0) and/or random points
     if x0 is None:
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -169,8 +169,8 @@ def base_minimize(func, dimensions, base_estimator,
             "%s has to be a regressor." % base_estimator)
 
     # Check params
-    space = Space(dimensions)
     rng = check_random_state(random_state)
+    space = Space(dimensions)
 
     # Initialize with provided points (x0 and y0) and/or random points
     if x0 is None:
@@ -261,6 +261,7 @@ def base_minimize(func, dimensions, base_estimator,
     # Bayesian optimization loop
     models = []
     n_model_iter = n_calls - n_total_init_calls
+    transformed_bounds = np.array(space.transformed_bounds)
     for i in range(n_model_iter):
         gp = clone(base_estimator)
 
@@ -296,6 +297,9 @@ def base_minimize(func, dimensions, base_estimator,
                 if a < best:
                     next_x, best = x, a
 
+        # lbfg should handle this but just in case there are precision errors.
+        next_x = np.clip(
+            next_x, transformed_bounds[:, 0], transformed_bounds[:, 1])
         next_x = space.inverse_transform(next_x.reshape((1, -1)))[0]
         yi.append(func(next_x))
         Xi.append(next_x)
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -39,7 +39,7 @@ def base_minimize(func, dimensions, base_estimator,
                   acq_func="EI", acq_optimizer="auto",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96, space=None):
+                  xi=0.01, kappa=1.96):
     """
     Parameters
     ----------
@@ -168,11 +168,9 @@ def base_minimize(func, dimensions, base_estimator,
         raise ValueError(
             "%s has to be a regressor." % base_estimator)
 
+    space = Space(dimensions)
     # Check params
     rng = check_random_state(random_state)
-
-    if space is None:
-        space = Space(dimensions)
 
     # Initialize with provided points (x0 and y0) and/or random points
     if x0 is None:
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -168,8 +168,8 @@ def base_minimize(func, dimensions, base_estimator,
         raise ValueError(
             "%s has to be a regressor." % base_estimator)
 
-    space = Space(dimensions)
     # Check params
+    space = Space(dimensions)
     rng = check_random_state(random_state)
 
     # Initialize with provided points (x0 and y0) and/or random points
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -168,7 +168,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
     """
     # Check params
     rng = check_random_state(random_state)
-    space = Space(dimensions)
+    space = Space(dimensions, transform_space="hypercube")
 
     # Default GP
     if base_estimator is None:
@@ -187,4 +187,4 @@ def gp_minimize(func, dimensions, base_estimator=None,
         n_points=n_points, n_random_starts=n_random_starts,
         n_restarts_optimizer=n_restarts_optimizer,
         x0=x0, y0=y0, random_state=random_state, verbose=verbose,
-        callback=callback)
+        callback=callback, space=space)
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -3,15 +3,15 @@
 import numpy as np
 
 from sklearn.base import clone
-from skopt.learning import GaussianProcessRegressor
 from sklearn.gaussian_process.kernels import Matern
 from sklearn.gaussian_process.kernels import ConstantKernel
 from sklearn.gaussian_process.kernels import WhiteKernel
 from sklearn.utils import check_random_state
 
-from .base import base_minimize
-from ..space import Space
-
+from skopt.optimizer.base import base_minimize
+from skopt.learning import GaussianProcessRegressor
+from skopt.space import check_dimensions
+from skopt.space import Space
 
 def gp_minimize(func, dimensions, base_estimator=None,
                 n_calls=100, n_random_starts=10,
@@ -168,7 +168,9 @@ def gp_minimize(func, dimensions, base_estimator=None,
     """
     # Check params
     rng = check_random_state(random_state)
-    space = Space(dimensions, transform_space="hypercube")
+    # To make sure that GP operates in the [0, 1] space
+    dimensions = check_dimensions(dimensions, transform="hypercube")
+    space = Space(dimensions)
 
     # Default GP
     if base_estimator is None:
@@ -187,4 +189,4 @@ def gp_minimize(func, dimensions, base_estimator=None,
         n_points=n_points, n_random_starts=n_random_starts,
         n_restarts_optimizer=n_restarts_optimizer,
         x0=x0, y0=y0, random_state=random_state, verbose=verbose,
-        callback=callback, space=space)
+        callback=callback)
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -8,10 +8,10 @@ from sklearn.gaussian_process.kernels import ConstantKernel
 from sklearn.gaussian_process.kernels import WhiteKernel
 from sklearn.utils import check_random_state
 
-from skopt.optimizer.base import base_minimize
-from skopt.learning import GaussianProcessRegressor
-from skopt.space import check_dimensions
-from skopt.space import Space
+from .base import base_minimize
+from ..learning import GaussianProcessRegressor
+from ..space import check_dimension
+from ..space import Space
 
 def gp_minimize(func, dimensions, base_estimator=None,
                 n_calls=100, n_random_starts=10,
@@ -168,8 +168,10 @@ def gp_minimize(func, dimensions, base_estimator=None,
     """
     # Check params
     rng = check_random_state(random_state)
+
     # To make sure that GP operates in the [0, 1] space
-    dimensions = check_dimensions(dimensions, transform="hypercube")
+    dimensions = [
+        check_dimension(dim, transform="normalize") for dim in dimensions]
     space = Space(dimensions)
 
     # Default GP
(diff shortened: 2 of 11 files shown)


MechCoder:
@glouppe @betatim please review.

MechCoder:
Not sure, why the convergence tests pass in Python3 but fail in Python2.

betatim:
(restarted travis in the hopes that it is random ... lame solution I know)

betatim:
Not sure I like the new `transform_space` parameter. Especially as parameter for a `Dimension` I think it should just be `transform`.

I like having a chain of transformers.

The other thing that is ugly is passing `dimensions` and a `space` to `base_minimize`. If we really need the space, then we should exclusively pass that as it contains the dimensions.

Some questions I am asking myself (my answers in brackets):
* GP always wants to transform to `(0, 1)`? (yes)
* should the user be able to override that choice? (no)
* other minimisation methods don't want things to be transformed to `(0,1)` or want to apply their own transforms to some of the dimensions (correct)
* does the user ever need to be able to specify their own transform? (maybe, support this by allowing people to pass a fully configured `Space` instance to `dimensions=`)

-> Have a `transform` parameter for `Dimension` that GP sets to `Normalized()` for those that it wants to have normalised.

WDYT?

betatim:
Might be worth making [this code](https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/space.py#L350-L369) available as `check_dimension` or something like that so that `gp_minimize` (and friends) can use it to take user specified arguments and turn them into a `Dimension`.

MechCoder:
> Especially as parameter for a Dimension I think it should just be transform.

That was my idea as well, but then there is a `transform` method as well. So I cannot set `self.transform = transform`. We can underscore that particular attribute then.

I have to have a closer read, but I agree to all the other points.

MechCoder:
I've introduced `check_dimensions` that has an optional `transform` parameter because I preferred that to the idea of setters and getters on Dimension instances. For now I've assumed that if `transform` is provided all dimensions have to be transformed. If needed, later we can allow `transform` to be a list of strings as well.

I've also moved the transformers to a separate file.

betatim:
> That was my idea as well, but then there is a `transform` method.

Should we name the argument `transformation=`? Can already see myself forever trying to remember which of `sp.transform_` and `sp.transform` is the attribute and which the method :)

skopt/optimizer/gp.py
```diff
@@ -3,15 +3,15 @@
...
 from sklearn.gaussian_process.kernels import WhiteKernel
 from sklearn.utils import check_random_state
 
-from .base import base_minimize
-from ..space import Space
-
+from skopt.optimizer.base import base_minimize
+from skopt.learning import GaussianProcessRegressor
+from skopt.space import check_dimensions
+from skopt.space import Space
```
betatim: relative imports are out of fashion? I think we should stick with using them, or make a new PR where we switch the whole code base.

skopt/space.py
```diff
@@ -5,98 +5,70 @@
...
-        Returns
-        -------
-        * `X` [array-like, shape=(n_samples,)]:
-            The original categories.
-        """
-        Xt = np.asarray(Xt)
-        return [
-            self.inverse_mapping_[i] for i in self._lb.inverse_transform(Xt)
-        ]
+def check_dimensions(dimensions, transform=None):
```
betatim: Originally thought about having `check_dimension()` (no s) and the loop being performed by the caller. It would make it easier for the caller to decide what transform to they wanted for what dimension. Not sure if that is needed or over engineering. You decide.

skopt/space.py
```diff
@@ -5,98 +5,70 @@
...
+
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - as a list of categories (for `Categorical` dimensions), or
+        - an instance of a `Dimension` object (`Real`, `Integer` or
+          `Categorical`).
+
+    * `transform` ["hypercube", optional] :
```
betatim: Nitpicking ahead: I really dislike `'hypercube'` as name for this 😃 `'unithypercube'` maybe, but that is super long. Why not boring old `'normalize'`?

skopt/space.py
```diff
@@ -205,9 +198,10 @@ def bounds(self):
 
     @property
     def transformed_bounds(self):
-        if self.prior == "uniform":
+        if self.transform_ == "hypercube":
+            return (0.0, 1.0)
+        elif self.prior == "uniform":
```
betatim: Had to look at this a few times and still not sure if mixing `self.prior` and `self.transform_` in the `if/elif` does what we want. I'd prefer splitting these into two different `if` blocks one for priors, one for transforms.

skopt/tests/test_space.py
```diff
@@ -225,3 +227,38 @@ def test_space_api():
...
+    a = Real(2.0, 30.0, transform="hypercube")
+    for i in range(50):
+        yield (check_limits, a.rvs(random_state=i), 2, 30)
+
+    rng = np.random.RandomState(0)
+    X = rng.randn(100)
+    X_scaled = 28 * (X - X.min()) / (X.max() - X.min()) + 2
+    X_scaled = np.clip(X_scaled, 2.01, 30.01)
+
+    # Check transform
```
betatim: -> Check transformed values are all between zero and one

(because both the asserts are _less(...) it took me a while to parse it, hopefully the comment will help with that)

betatim: Do we want a transform that does `0<= x <= 1` or `0 < x < 1`? The test here is for the latter but my guess would be that we want out transform to be the first?

Maybe instead of random values or in addition to them we should test explicitly the edge cases.

skopt/transformers.py
```diff
@@ -0,0 +1,134 @@
...
+
+class HyperCube(Transformer):
+    """
+    Transform each dimension into [0, 1]
+    """
+    def __init__(self, low, high):
+        self.low = low
+        self.high = high
+
+    def transform(self, X):
```
betatim: Need some error handling for when the user wants to transform a value outside of the `low -> high` range.

skopt/space.py
```diff
@@ -5,98 +5,70 @@
...
 
-from sklearn.preprocessing import LabelBinarizer
 from sklearn.utils import check_random_state
 from sklearn.utils.fixes import sp_version
 
+from skopt.transformers import CategoricalEncoder
+from skopt.transformers import HyperCube
+from skopt.transformers import Identity
+from skopt.transformers import Log10
+from skopt.transformers import Pipeline
```
betatim: Same comment about relative vs absolute imports.

skopt/optimizer/gp.py
glouppe: Nope. We should keep relative imports.

glouppe:
Thanks Manoj! I need to have a closer look, I'll try to find some spare time for a review later today.

glouppe:
From a code organization point of view, what do you think of defining a transformation as a wrapper around a `Dimension` object (i.e. as function taking a dimension as input and producing a new dimension as output)? In this way, chaining transformations would be done in a functional way (e.g. `Hypercube(Real(0, 10))`). 

(not strong opinion yet, I am thinking aloud...)

skopt/optimizer/base.py
```diff
@@ -168,9 +168,9 @@ def base_minimize(func, dimensions, base_estimator,
         raise ValueError(
             "%s has to be a regressor." % base_estimator)
 
+    space = Space(dimensions)
     # Check params
```
glouppe: this comment should be moved above

skopt/space.py
```diff
@@ -5,98 +5,70 @@
...
 
 # helper class to be able to print [1, ..., 4] instead of [1, '...', 4]
 class _Ellipsis:
     def __repr__(self):
         return '...'
 
 
-class _Identity:
```
glouppe: Not sure I like moving everything to a top-level `transformers` module. Maybe create instead a `space/` submodule?

skopt/space.py
glouppe: +1 for nomalize

MechCoder:
> From a code organization point of view,

If we do that, then it might make sense to completely decouple the transformations from the dimension instances. Something like just defining the `init` logic in Dimensions and the `transform`, `rvs` attributes to the Transformer.

```python
class Real(Dimension):
    def __init__(self, low, high):
        self.low = low
        self.high = high

class Log10(Transformer):
    def __init__(self, dimension):
        self.dimension = dimension
        self._rvs = uniform(
            np.log10(self.low),
            np.log10(self.high) - np.log10(self.low))

    def rvs(self, n_samples):
        # Sample from rvs

    def transform(self, X):
        np.log10
```        

and then call `Log10(Real(10**-5, 10**5))` internally.

MechCoder:
What do we do about a chain of transformations?

```
Pipeline([transformers], Dimension)
```

would be a straightforward way.

MechCoder:
I like the idea, because it seems cleaner to me.

glouppe:
You could similarly do TransformB(TransformA(BaseDimension)) since
TransformA would share the Dimension API.

MechCoder:
I would need some time to think but is there a particular reason that you would prefer chained function transformers as opposed to a class accepting a list of transformers and a Dimension instance?

glouppe:
The way I see a `Dimension` object is basically as a set and a transformation as a function taking a set as input and producing another set as output. Or more formally, if we have a transformation `f` and a dimension `D`, then `f(D)` is the image of D through `f`, and is itself a set/dimension. This is why I think it would be natural to define transformation as a wrapper object taking as argument another Dimension object and then implementing the `Dimension` interface. Another argument, is that, in my opinion, it might make things cleaner in terms of composition if a `Dimension` does not have to care about how it is transformed.

glouppe:
(Of course this is all open for discussion, but we should really think twice on this important issue)

MechCoder:
The more I think about it, the more it seems to make sense that way.

skopt/space.py
MechCoder: I was hesitant because `normalize` can mean a number of things. Is there a standard way to define this scaling?

MechCoder:
@glouppe @betatim In your example, that is `Normalize(Real(0, 10))` or more generally `TransformB(TransformA(BaseDimension))` after sampling from the resultant dimensional space, there is no way to go back to the original dimensional space for a given sample, unless you have implemented an inverse mapping as well. 

Which is why I feel, the current interface is all right, i.e for a Dimension:

1. `rvs` samples it from the original space
2. `transform` maps it to a transformed space as given by the `transformer` argument.
3, `inverse_transform` maps it to from the transformed space to the original space.

Alternatively, we could do away with the `transform` and `inverse_transform` on the dimension objects and replace it with a `Transformer` object that takes in the original space and warped space as class arguments, but I would not prefer that.

glouppe:
Is this now superseded by #243 ?

glouppe:
> or more generally TransformB(TransformA(BaseDimension)) after sampling from the resultant dimensional space, there is no way to go back to the original dimensional space for a given sample, unless you have implemented an inverse mapping as well.

Why wouldnt we implement the inverse mapping as well?

MechCoder:
Let's say I implement a Transformer as a function that returns a new dimensional space and I apply a  chain of transformations to convert the original Dimensional space (provided by the user) to the new Dimensional space (supplied to the optimizer). I also implement the inverse mapping, i.e chain of transformations to go back to the original dimensional space. Also now I have a new point as computed by the optimizer.

How do you suggest to convert this new point to the original dimensional space? (The inverse mapping takes in a Dimension and returns a Dimension)

skopt/space/space.py
```diff
@@ -5,98 +5,65 @@
...
-        Returns
-        -------
-        * `X` [array-like, shape=(n_samples,)]:
-            The original categories.
-        """
-        Xt = np.asarray(Xt)
-        return [
-            self.inverse_mapping_[i] for i in self._lb.inverse_transform(Xt)
-        ]
+def check_dimension(dimension, transform=None):
```
glouppe: Can you also document about the checks that are done?

skopt/space/space.py
```diff
@@ -163,35 +130,53 @@ def __init__(self, low, high, prior="uniform"):
...
+        # The rvs on Dimension calls inverse_transform on the points sampled
+        # using _rvs
+        if self.transform_ == "normalize":
+            self._rvs = uniform(0, 1)
+            if self.prior == "uniform":
+                self.transformer = Pipeline(
+                    [Identity(), Normalize(low, high)])
+            else:
+                self.transformer = Pipeline(
+                    [Log10(), Normalize(np.log10(low), np.log10(high))]
```
glouppe: This is OK for now, but this is why the functional approach is better in my opinion. For example in this case, the second step in the pipeline need to know the bounds from the previous step, which is not that clean.

skopt/space/space.py
```diff
@@ -163,35 +130,53 @@ def __init__(self, low, high, prior="uniform"):
...
               `log10(lower)` and `log10(upper)`.`
+
+        * `transform` [None or "normalize", optional]:
+            If `transform=normalize`, calling `transform` on X scales X to
+            [0, 1]
         """
         self.low = low
         self.high = high
         self.prior = prior
+        self.transform_ = transform
```
glouppe: is there a reason to use a trailing underscore here?

glouppe: Oh right, we already have the `transform` method. Yuck.

skopt/tests/test_space.py
```diff
@@ -225,3 +227,35 @@ def test_space_api():
                       [(0.0, 1.0), (-5, 5), (0.0, 1.0), (0.0, 1.0), (0.0, 1.0),
                        (np.log10(1.0), np.log10(5.0)), (0.0, 1.0)]):
         assert_array_equal(b1, b2)
+
+
+def test_real_hypercube():
```
glouppe: This test should be renamed

skopt/tests/test_gp_opt.py
```diff
@@ -21,7 +21,7 @@ def check_minimize(func, y_opt, bounds, acq_optimizer, acq_func,
 
 
 def test_gp_minimize():
-    for search, acq in product(["sampling", "lbfgs"], ["LCB", "PI", "EI"]):
+    for search, acq in product(["sampling", "lbfgs"], ["LCB", "EI"]):
```
glouppe: what is the reason for removing PI?

glouppe:
Not super fan of the API, but that will do the job for now. +1 for merge once the other comments are addressed.

skopt/space/__init__.py
```diff
@@ -0,0 +1 @@
+from .space import *
```
glouppe: Can you add a top-level docstring for the module?

skopt/space/space.py
MechCoder: But how to address this? (https://github.com/scikit-optimize/scikit-optimize/pull/237#issuecomment-251415962)

skopt/tests/test_gp_opt.py
MechCoder: I removed it because tests were failing in Python2 by a small value and I did not want to increase the number of iterations for that. Moreover, I'm not sure PI is commonly used.

MechCoder:
@glouppe Addressed your comments, Anything else?

PR #243, opened 2016-10-04
Proof of Concept: Keep transformation and dimensions independent

MechCoder:
With the current API, the way we handle dimensions and transformations is confusing. The places where the idea of a transformed dimensional space are used currently is while:
1. Supplying a log-uniform transform to `Real`
2. Supplying a one-hot encoding to `Categorical`

Also, the current methods on every `Dimension` instance are:
1. Rvs -- Sampling from the transformed dimensional space and then transforming back to the original space.
2. Transform -- Transforming to the transformed dimensional space.
3. Inverse Transform -- Transforming back from the transformed dimensional space.

I feel we should only keep `rvs` and have a `Transformer` object that takes the original dimension as a constructor argument and the transformed space should be defined by the Transformer itself. In this way we can do

``` python
t = Log10(Real(10**2, 10**4))
t.transform(10**3)
t.inverse_transform(3)
```

The reason it is better to use it a class rather than a function that returns the transformed dimension is that it becomes convenient to do the inverse mapping as well.

Open Questions:
- [ ] Is this the best way?
- [ ] What would be the best way to handle a pipeline of transformations?
- [ ] What would be the best way to handle this in terms of the Space API?

### DIFF
--- a/skopt/space.py
+++ b/skopt/space.py
@@ -76,19 +76,6 @@ class Log10(Transformer):
         return 10.0 ** np.asarray(Xt, dtype=np.float)
 
 
-class _Log10:
-    """Base 10 logarithm transform."""
-
-    def fit(self, X):
-        return self
-
-    def transform(self, X):
-        return np.log10(np.asarray(X, dtype=np.float))
-
-    def inverse_transform(self, Xt):
-        return 10.0 ** np.asarray(Xt, dtype=np.float)
-
-
 class CategoricalEncoder(Transformer):
     """OneHotEncoder that can handle categorical variables."""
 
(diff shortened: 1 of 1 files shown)


glouppe:
> Rvs -- Sampling from the transformed dimensional space and then transforming back to the original space.

This is confusing in my opinion, because we will always wonder from what space we are actually sampling.

`rvs` should sample the dimension it is attached to, no more. The semantics is clear and simple.

glouppe:
I have not been following closely. Can you remind me what are the "confusing" parts we are trying to address here?

MechCoder:
> rvs should sample the dimension it is attached to, no more. The semantics is clear and simple.

Not so sure about that, (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/space.py#L119). For the "log-uniform case", we sample in the transformed log(dimensions) and then transform it back to the original space.

MechCoder:
Oh I agree, that it should be that way. But I'm afraid it is not like that right now.

MechCoder:
> Can you remind me what are the "confusing" parts we are trying to address here?

On how to elegantly supply a transform to a Dimension instance.

MechCoder:
Closing for now, can reopen if there is a necessity

Issue #190, opened 2016-08-10
Notebook showing how to persist/load a optimization result


betatim:
Persisting the result of an optimisation run is a useful tool to have. Either for plotting later on, creating a history of your research or sharing your results with others.

Having a notebook that gives a clear recommendation on how to do this would be nice. This avoids people doing "crazy" things.

Could be as simple as using `joblib.dumps(result)`

betatim:
Opinions on providing `skopt.load` and `skopt.dump` as shortcuts to `sklearn.externals.joblib.dump/load`?

glouppe:
No strong opinion, fine with me.

betatim:
How should we handle `res.specs['args']['func']` when serialising the result object? It might contain weird things that can't be pickled and you might want to share your results with others who don't have the code for it. Sharing pickles between different users/python versions/machines is probably not the smartest thing anyway so should we worry about this? It could be that `joblib` makes promises on top of `pickle`??

I'd be interested in a method that allows the user to share their results with others (including their future self which might be on a different python version). This is potentially much harder to solve...

MechCoder:
We can have an optional `return_specs`, which the user should be intelligent enough to set to be True if he wants to pickle or we can backport cloupickle, pyspark does it over here (https://github.com/apache/spark/blob/master/python/pyspark/cloudpickle.py)

MechCoder:
Umm. Isn't using `joblib.load and dump` over-kill for our-case? It says it is useful only for large numpy arrays (http://scikit-learn.org/stable/modules/model_persistence.html) Our largest data structure is a list of lists.

betatim:
Reason for suggesting `joblib` is that it is the recommended way to pickle scikit-learn classifiers, which can (in essence) be large numpy arrays. Not sure how applicable that is to our regressors.

I would strongly favour a solution where we don't have to maintain the pickling code ourselves (straight `pickle` or `joblib` or some other library).

I would make the "save specs or not" flag an argument of the `save()` function.

MechCoder:
Hmm.. but then how do we serialize lambdas :(

betatim:
Do we understand why they aren't picklable by default? The "hacks" I've seen to make them (and other common things that aren't picklable) are usually quite short, so it seems like there must be a good reason why the python devs don't just include that no? We should find out.

We should try and balance the (realistic) use cases with maintaining our own pickle substitute.

mp4096:
[...]

PR #245, opened 2016-10-10
Persistence with skopt.dump() and skopt.load()

mp4096:
See #190 

To do:
- [x] Write unit tests

~~Be more specific when catching the pickling exception, search for `func` in the error message~~

### DIFF
--- a/examples/store-and-load-results.ipynb
+++ b/examples/store-and-load-results.ipynb
@@ -1,0 +1,269 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Store and load `skopt` optimization results\n",
+    "\n",
+    "Mikhail Pak, October 2016."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "np.random.seed(777)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Problem statement\n",
+    "\n",
+    "We often want to store optimization results in a file. This can be useful, for example,\n",
+    "\n",
+    "* if you want to share your results with colleagues;\n",
+    "* if you want to archive and/or document your work;\n",
+    "* or if you want to postprocess your results in a different Python instance or on an another computer.\n",
+    "\n",
+    "The process of converting an object into a byte stream that can be stored in a file is called _serialization_.\n",
+    "Conversely, _deserialization_ means loading an object from a byte stream.\n",
+    "\n",
+    "**Warning:** Deserialization is not secure against malicious or erroneous code. Never load serialized data from untrusted or unauthenticated sources!"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Simple example\n",
+    "\n",
+    "We will use the same optimization problem as in the [`bayesian-optimization.ipynb`](https://github.com/scikit-optimize/scikit-optimize/blob/master/examples/bayesian-optimization.ipynb) notebook:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from skopt import gp_minimize\n",
+    "\n",
+    "noise_level = 0.1\n",
+    "\n",
+    "def obj_fun(x, noise_level=noise_level):\n",
+    "    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2)) + np.random.randn() * noise_level\n",
+    "\n",
+    "res = gp_minimize(obj_fun,            # the function to minimize\n",
+    "                  [(-2.0, 2.0)],      # the bounds on each dimension of x\n",
+    "                  x0=[0.],            # the starting point\n",
+    "                  acq_func=\"LCB\",     # the acquisition function (optional)\n",
+    "                  n_calls=15,         # the number of evaluations of f including at x0\n",
+    "                  n_random_starts=0,  # the number of random initialization points\n",
+    "                  random_state=777)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As long as your Python session is active, you can access all the optimization results via the `res` object.\n",
+    "\n",
+    "So how can you store this data in a file? `skopt` conveniently provides functions `skopt.dump()` and `skopt.load()` that handle this for you. These functions are essentially thin wrappers around the [`joblib`](http://pythonhosted.org/joblib) module's `dump()` and `load()`.\n",
+    "\n",
+    "We will now show how to use `skopt.dump()` and `skopt.load()` for storing and loading results."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Using `skopt.dump()` and `skopt.load()`\n",
+    "\n",
+    "For storing optimization results into a file, call the `skopt.dump()` function:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "from skopt import dump, load\n",
+    "\n",
+    "dump(res, 'result.pkl')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "And load from file using `skopt.load()`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "-0.17493386623950199"
+      ]
+     },
+     "execution_count": 4,
+     "metadata": {},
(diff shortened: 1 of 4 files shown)


betatim:
Rendered notebook: https://173-54340642-gh.circle-artifacts.com/0/tmp/circle-artifacts.jN1zNtS/doc/skopt/notebooks/store-and-load-results.html

betatim:
I am in favour of providing these two methods for users (as opposed to #246). Mainly to save on typing.

* Should we use some magic to try and import `dill` and if it works use that instead? I am -1
* Instead of removing `func` and retrying the pickling I would add an argument to, maybe `dump(res, without_objective=True)`, where the user can explicitly say that they want to store things without the objective

Thoughts @glouppe and @MechCoder?

mp4096:
I'm against trying to import `dill` and using `pickle` as a fallback due to the compatibility. The `load()` function cannot know which one was used; in worst case, one has to try both.

MechCoder:
Thanks for handling this. We should also have some sort of checkpointing machinery so that the information is dumped after every iteration incase something goes wrong after some iteration.

betatim:
Coming back to this, I think we want this approach and not the other one in #246. Things that need doing are tracked in the first post of this PR.

We decided against `dill` (and guessing). I would also prefer that we simplify things by asking the user to explicitly decide if they [want to store the objective or not](https://github.com/scikit-optimize/scikit-optimize/pull/245#issuecomment-252556914). This will make our code cleaner and no one will be surprised if there is no objective in the pickle.

I'd put checkpointing in a new PR once this is merged. Does that work for you @MechCoder?

Would be good to merge this before 0.2 is released.

mp4096:
Ok,, I'll add an explicit argument for storing the objective function!

skopt/utils.py
```diff
@@ -44,3 +52,58 @@ def create_result(Xi, yi, space=None, rng=None, specs=None, models=None):
...
+    * `res` [`OptimizeResult`, scipy object]:
+        Optimization result object to be stored.
+
+    * `filename` [string]:
+        The path of the file in which it is to be stored.
+    """
+
+    try:
+        with open(filename, 'wb') as f:
+            pickle.dump(res, f)
```
MechCoder: We should use joblib.dump no? (from `sklearn.externals.joblib`)

examples/store-and-load-results.ipynb
```diff
@@ -0,0 +1,387 @@
...
+   "metadata": {},
+   "source": [
+    "`skopt.dump()` and `skopt.load()` use the `pickle` module internally. Notice that if the objective function is not serializable, `skopt.dump()` will raise a warning and store a copy of the optimization results object without the objective function. \n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Custom alternatives\n",
```
MechCoder: Is this section necessary? Not sure we should be explaining about pickle or joblib here (especially since we are not doing anything more other than a thin wrapper over these)

MechCoder:
What do you guys think about adding an argument to the minimize functionalities that stores the optimisation results on to disk directly after the optimisation is over?

betatim:
A bit weary of adding yet more arguments to the `*_minimize()` function. At some point it will be able to configure the coffee machine to brew coffee while we are optimising :-/ I'd rather support having a ready made callback that stores the result every `n` iterations.

skopt/utils.py
betatim: Somewhere we discussed whether or not to always use `joblib`. I think it was about "is it necessary to use `joblib` or would `pickle` suffice?" Can't find that thread right now. I'd be happy simply using `joblib` always.

examples/store-and-load-results.ipynb
mp4096: Ok, I'll remove it.

skopt/utils.py
mp4096: It was discussed in #190, somewhere around [this comment](https://github.com/scikit-optimize/scikit-optimize/issues/190#issuecomment-249296787)

I'm ok with `joblib`.

mp4096:
Liked your comment about the coffee brewing. An email notification after each iteration would be nice too. :grin:

I think a checkpoint callback is an elegant solution.

betatim:
> Liked your comment about the coffee brewing. An email notification after each iteration would be nice too. :grin:

-> in a new PR please :)

mp4096:
* Now using `joblib`. Additional keyword arguments are simply propagated further to `joblib.dump()` / `joblib.load()` 
* I think `store_objective` is better than `without_objective` for it is positive/affirmative. Maybe `dump_objective` would be even better?

mp4096:
I wasn't sure whether the unit test for `dump()` and `load()` should go into `test_common.py` or in its own script. Is it ok this way?

skopt/tests/test_utils.py
```diff
@@ -0,0 +1,65 @@
...
+from skopt import dump
+
+
+def obj_fun(x):
+    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2))
+
+
+def check_optimization_results_equality(res_1, res_2):
+    # Check if the results objects have the same keys
+    assert_equal(res_1.keys(), res_2.keys())
```
MechCoder: Should sort before comparing?

skopt/tests/test_utils.py
```diff
@@ -0,0 +1,65 @@
...
+from sklearn.utils.testing import assert_array_equal
+from sklearn.utils.testing import assert_array_less
+from sklearn.utils.testing import assert_equal
+from sklearn.utils.testing import assert_not_equal
+from sklearn.utils.testing import assert_less_equal
+from sklearn.utils.testing import assert_greater
+from sklearn.utils.testing import assert_true
+
+from skopt import gp_minimize
+
```
MechCoder: extra space

skopt/tests/test_utils.py
```diff
@@ -0,0 +1,65 @@
...
+        dump(res, f)
+        f.seek(0)
+        res_loaded = load(f)
+    check_optimization_results_equality(res, res_loaded)
+    assert_true('func' in res_loaded.specs['args'])
+
+    # Test dumping without objective function
+    with tempfile.TemporaryFile() as f:
+        dump(res, f, store_objective=False)
+        f.seek(0)
```
MechCoder: Why is this line necessary?

skopt/tests/test_utils.py
```diff
@@ -0,0 +1,65 @@
...
+from sklearn.utils.testing import assert_greater
+from sklearn.utils.testing import assert_true
+
+from skopt import gp_minimize
+
+from skopt import load
+from skopt import dump
+
+
+def obj_fun(x):
```
MechCoder: You can import from `skopt.benchmarks` but meh...

skopt/utils.py
```diff
@@ -44,3 +51,70 @@ def create_result(Xi, yi, space=None, rng=None, specs=None, models=None):
...
+        method corresponding to one of the supported filename extensions ('.z',
+        '.gz', '.bz2', '.xz' or '.lzma') will be used automatically.
+
+    * `store_objective` [boolean, default=True]:
+        Whether the objective function should be stored. Set `store_objective`
+        to `False` if your objective function (`.specs['args']['func']`) is
+        unserializable (i.e. if an exception is raised when trying to serialize
+        the optimization result). Notice that in this case a deep copy of the
+        optimization result is created. This can potentially lead to performance
+        problems.
```
MechCoder: You can add the note that it is advisable to delete the objective function before dumping to avoid such issues

MechCoder:
Looks fine pending minor comments. Thanks @mp4096! @betatim please merge if happy.

skopt/tests/test_utils.py
mp4096: * A `tempfile.TemporaryFile` object is deleted right after it has been closed, in this case when we leave the `with` context.
* `joblib` writes into the file and leaves the seeker pointing to EOF. So if we `read` or `load` it, the file will appear to be empty.
* So we have to 'rewind' the file object to the beginning.

skopt/tests/test_utils.py
mp4096: Good point, thanks!

skopt/tests/test_utils.py
MechCoder: Are you sure about the second point? Atleast tests pass locally after I remove the `seek` line.

skopt/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+import tempfile
+
+from sklearn.utils.testing import assert_array_almost_equal
+from sklearn.utils.testing import assert_array_equal
+from sklearn.utils.testing import assert_array_less
+from sklearn.utils.testing import assert_equal
+from sklearn.utils.testing import assert_not_equal
+from sklearn.utils.testing import assert_less_equal
+from sklearn.utils.testing import assert_greater
+from sklearn.utils.testing import assert_true
```
betatim: could you remove the imports that aren't used please? Needed: assert_true, assert_equal and assert_array_equal

skopt/tests/test_utils.py
mp4096: You're absolutely right, sorry! I just checked that `dump` leaves the seeker at the EOF and haven't thought that `load` itself might do `seek(0)`, which it does.

skopt/tests/test_utils.py
mp4096: Thanks, missed this one.

mp4096:
Are you ok with the wording in the example notebook?
https://github.com/mp4096/scikit-optimize/blob/dev-save/examples/store-and-load-results.ipynb

betatim:
Merged! 🎉 🎈

mp4096:
Cool, thanks! :cake:

PR #246, opened 2016-10-10
Persistence: Notebook with examples only

mp4096:
See #190


betatim:
Rendered notebook: https://172-54340642-gh.circle-artifacts.com/0/tmp/circle-artifacts.y5UrNPN/doc/skopt/notebooks/store-and-load-results.html

PR #247, opened 2016-10-11
[MRG] Provide gradients to LCB

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/244

- [x] Compute gradients of all the kernels with respect to a new point
- [x] Compute gradient of the mean and std of the predictions of the GaussianProcessRegressor
- [x] Compute gradient of the Lower Confidence Bound function
- [x] Integrate with `gp_minimize`
- [x] Use in-place operations


MechCoder:
Any high level comments?

skopt/learning/tests/test_gp_kernel.py
```diff
@@ -0,0 +1,53 @@
...
+
+def numerical_gradient(X, Y, kernel):
+    grad = []
+    for y in Y:
+        num_grad = optimize.approx_fprime(X, kernel_X_Y, 1e-4, y, kernel)
+        grad.append(num_grad)
+    return np.asarray(grad)
+
+def test_gradient_correctness():
+    for kernel in KERNELS:
```
betatim: should we use a `yield check_gradient...` kind of construct to make it easier to spot which of the kernels it is that is failing the test?

skopt/learning/gp_kernels.py
```diff
@@ -0,0 +1,186 @@
...
+from sklearn.gaussian_process.kernels import ExpSineSquared as sk_ExpSineSquared
+from sklearn.gaussian_process.kernels import Matern as sk_Matern
+from sklearn.gaussian_process.kernels import Product as sk_Product
+from sklearn.gaussian_process.kernels import RationalQuadratic as sk_RationalQuadratic
+from sklearn.gaussian_process.kernels import RBF as sk_RBF
+from sklearn.gaussian_process.kernels import Sum as sk_Sum
+from sklearn.gaussian_process.kernels import WhiteKernel as sk_WhiteKernel
+
+
+class Kernel(sk_Kernel):
```
betatim: No one inherits from this? It is good to have it in one place to define the API, but can we make use of it somehow?

skopt/learning/gp_kernels.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Base class for skopt.gaussian_process kernels.
+    Supports computation of the gradient of the kernel with respect to X
+    """
+    def gradient_X(self, x, Y):
```
betatim: I find the use of `x`, `X` and `Y` a bit confusing. Coming from `sklearn` land where there is a clear meaning for `X` and `y` already you have to think a few times before adopting to this convention. I'll make a standalone comment for this.

betatim:
High level comments, that is just about what I am good for 😀 

The naming of parameters in the gradient computation is a bit confusing. I like `x` for "gradient at coordinate x". Not sure I like `X` and `Y`.

Can we say "gradient at x wrt each of the dimensions"?

Not a super helpful comment I know, just wanted to mention that I find the notation confusing :-/

betatim:
BTW, increasing test coverage by 2%! 🎉

skopt/learning/__init__.py
```diff
@@ -4,6 +4,17 @@
...
+from .gp_kernels import ConstantKernel
+from .gp_kernels import DotProduct
+from .gp_kernels import Exponentiation
+from .gp_kernels import ExpSineSquared
+from .gp_kernels import Matern
+from .gp_kernels import Product
+from .gp_kernels import RationalQuadratic
+from .gp_kernels import RBF
+from .gp_kernels import Sum
+from .gp_kernels import WhiteKernel
```
MechCoder: Are you referring to the `__all__`? That is used when you want to do `from skopt.learning import *`

skopt/learning/gp_kernels.py
MechCoder: I meant to just to avoid copying the documentation :/

skopt/learning/gp_kernels.py
MechCoder: I wrote it as small `x` to mean it as a 1-D array.

skopt/learning/gp_kernels.py
MechCoder: (which means I have to inherit from this which I forgot to do:P)

skopt/learning/__init__.py
betatim: I am confused why we import all the kernels but don't list them in `__all__`. Does this allow us to do `from skopt.learning import WhiteKernel`?

skopt/learning/gp_kernels.py
betatim: I like `x` for the "at this point"/1D array. Mostly confused by the use of `X` and `Y`.

skopt/learning/__init__.py
MechCoder: Yes! `__all__` is just used for the `*` wildcard while importing. I'll change it anyway

MechCoder:
Yay! Amazing speed improvements. One run of optimisation on `hart6` takes 321 seconds on master with `acq_func=LCB` but just 70 seconds on this branch. I did not check the performance but hopefully that will not change. More thorough benchmarks soon. :-)

MechCoder:
```
In master:
Hart6, 10 runs, LCB with lbfgs
Mean optimum: -3.23411021414
Std of optimal values0.0568717122
Best optima:- 3.32127871402
Mean time: 318.6931083202362

In this branch:
Hart6, 10 runs, LCB with lbfgs
Mean optimum: -3.197814482
Std of optimal values: 0.148118049358
Best optima: -3.3223090448
Mean time: 80.52304673194885

In master:
Branin, 3 runs, LCB with lbfgs
Mean optimum: 0.397887407408
Std of optimal values6.73478564188e-08
Best optima:0.39788735843
Mean time:  279.674

In this branch:
Branin, 3 runs, LCB with lbfgs
Mean optimum: 0.397887423718
Std of optimal values4.00767076148e-08
Best optima:0.397887375025
Mean time: 54.2164937655131

**Comment by glouppe** [comment]
Wow! This is fantastic :)

**Comment by betatim** [comment]
Bazinga! Seems like refreshing those linear algebra skills was worth it!

**Comment by MechCoder** [comment]
Updated from WIP to MRG

File: skopt/acquisition.py [hunk 5]
```diff
@@ -26,7 +26,7 @@ def _gaussian_acquisition(X, model, y_opt=None, method="LCB",
         raise ValueError("Acquisition function not implemented.")
 
 
-def gaussian_lcb(X, model, kappa=1.96):
+def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
```
glouppe: Can you document the new argument?

skopt/learning/gaussian_process/gpr.py
```diff
@@ -197,3 +204,112 @@ def fit(self, X, y):
...
+
+            if return_mean_grad:
+                grad = self.kernel_.gradient_X(X[0], self.X_train_)
+                grad_wrt_mean = np.dot(grad.T   , self.alpha_)
+
+                if return_std_grad:
+                    # XXX: Cache np.dot(K_trans, K_inv) from above
+                    grad_wrt_std = -np.dot(K_trans, np.dot(K_inv, grad))[0] / y_std
+                    return y_mean, y_std, grad_wrt_mean, grad_wrt_std
+                return y_mean, y_std, grad_wrt_mean
```
glouppe: Why do you return `y_std`? This avalanche of flags in the return values (which are not necessarily compatible as far I understand) makes the code very difficult to follow :/

skopt/learning/gaussian_process/kernels.py
```diff
@@ -0,0 +1,241 @@
...
+        -------
+        gradient_X: array-like, shape=(n_samples, n_features)
+            Gradient of K(x, X_train) with respect to x.
+        """
+        raise NotImplementedError
+
+
+class RBF(Kernel, sk_RBF):
+    def gradient_X(self, x, X_train):
+        # TODO: Do in-place operations
```
glouppe: left-over?

skopt/optimizer/base.py
```diff
@@ -282,6 +283,10 @@ def base_minimize(func, dimensions, base_estimator,
         elif acq_optimizer == "lbfgs":
             best = np.inf
 
+            if acq_func == "LCB":
+                approx_grad = False
+            else:
+                approx_grad = True
```
glouppe: `approx_grad = acq_func != "LCB"`?

MechCoder: Embarassing? Yes.

glouppe:
By the way, do you have an intuition about the speedup? Are numerical approximations so bad?

MechCoder:
I speculate that lbfgs has to call the expensive `predict` func `(n_dims * 2)` times for every gradient computation before, but now there is just one call to compute the gradient.

MechCoder:
I've addressed your comments. Would you be having any thoughts on making the gradient computation expression more readable?

glouppe:
>  Would you be having any thoughts on making the gradient computation expression more readable?

Not sure it can be made much clearer, but what I always find helpful is some inline comments specifying the mathematical formulas implemented.

MechCoder:
I added some inline comments. Would be great if you could give another pass.

skopt/learning/gaussian_process/gpr.py
```diff
@@ -197,3 +204,116 @@ def fit(self, X, y):
...
                         **{white_param: WhiteKernel(noise_level=0.0)})
         return self
+
+    def predict(self, X, return_std=False, return_cov=False,
+                return_mean_grad=False, return_std_grad=False):
+        """
+        In addition to the mean of the predictive distribution, also its
+        standard deviation (return_std=True) or covariance (return_cov=True),
+        the gradient of the mean and the gradient of the std can be optionally
+        provided.
```
glouppe: In the context of a GP, we should be more precise and specify with respect to what the gradients of mu and of std are computed (i.e. with respect to X, not with the hyper-params of the kernel)

skopt/learning/gaussian_process/tests/test_gpr.py
```diff
@@ -0,0 +1,92 @@
...
+    mean1, std1 = gpr1.predict(X, return_std=True)
+    mean2, std2 = gpr2.predict(X, return_std=True)
+    assert_array_equal(mean1, mean2)
+    assert_false(np.any(std1 == std2))
+
+
+def test_mean_gradient():
+    X = rng.randn(3, 3)
+    y = rng.randn(3)
+    X_new = rng.randn(3)
```
glouppe: Could we increase the size of the training and test data to 10?

skopt/learning/gaussian_process/tests/test_gpr.py
```diff
@@ -0,0 +1,92 @@
...
+        return_std=True, return_cov=False, return_mean_grad=True)
+    num_grad = optimize.approx_fprime(
+        X_new, lambda x: predict_wrapper(x, gpr)[0], 1e-4)
+    assert_array_almost_equal(mean_grad, num_grad, decimal=3)
+
+
+def test_std_gradient():
+    X = rng.randn(3, 3)
+    y = rng.randn(3)
+    X_new = rng.randn(3)
```
glouppe: same here

glouppe:
Some small comments, but otherwise LGTM once they are addressed! Fantastic work :) Really curious to see if we will get the same improvements for EI.

MechCoder:
Thanks! Will merge when tests pass.

PR #251, opened 2016-10-23
Ignore build_tools for the language bar

mp4096:
I'm sorry, this PR is _really_ nitpicking.

Currently, the language bar shows that `skopt` has 15.8% Mako and 2% Shell, although these are only for the build tools.

Wouldn't it be better to exclude them and indicate that `skopt` is 100% Python?

We could also include the Jupyter notebooks in the `examples` folder.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+build_tools/*  linguist-documentation=true


glouppe:
> We could also include the Jupyter notebooks in the examples folder.

Agreed! Could you also add that?

mp4096:
Sure!

...

Oh. Looks not so good: https://github.com/mp4096/scikit-optimize

betatim:
Oh wow! This is a whole new level of github-fu!

glouppe:
Hmm, why do we still see 93% for jupyter?

mp4096:
@glouppe Sorry, I think I got you wrong. The original PR excluded `build_utils` and the language bar was 100% Python.

Then I thought maybe it's nice to force _include_ notebooks to the language bar, but they're obviously much larger than the Python sources (probably due to images).

Ok, I'll now update the PR to show Python only.

MechCoder:
So where to see this language bar? I can't seem to find it anywhere?

mp4096:
Current:
current

Proposed:
proposed

MechCoder:
Whoa! I never noticed that bar till now.

betatim:
The only PR where less 🌈 is better! Thanks for this.

Issue #252, opened 2016-10-25
Bayesian optimisation example "broken"


betatim:
Something isn't quite right with one of the examples:

https://scikit-optimize.github.io/notebooks/bayesian-optimization.html

final plot shows that the model predicts a "constant". There was a discussion about this somewhere but I can't find it right now. I think we suspected it was related to a change in adding noise to the kernel??

MechCoder:
having a look

MechCoder:
All right, so I've been playing around with this for a while. Here is my code to reproduce with sklearn's GaussianProcessRegressor. I've also assumed that the standard deviation of the noise is known apriori, that is I've set `alpha` to be 0.1 which is never the case.

```python
import numpy as np
import matplotlib.pyplot as plt

from sklearn.gaussian_process import GaussianProcessRegressor
from skopt.learning.gaussian_process.kernels import ConstantKernel
from skopt.learning.gaussian_process.kernels import Matern

def f(x, noise_level=0.1):
    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2)) + np.random.randn() * noise_level

x_predict = np.linspace(-2, 2, 400).reshape(-1, 1)
x_fit = np.linspace(-2, 2, 20).reshape(-1, 1)
y = np.array([f(x) for x in x_fit])
y_predict = np.array([f(x) for x in x_predict])

cov_amplitude = ConstantKernel(1.0, (0.01, 1000.0))
matern = Matern(nu=2.5)
kernel = cov_amplitude * matern

skgpr = GaussianProcessRegressor(
    kernel=kernel, alpha=0.1, normalize_y=True, random_state=0)
skgpr.fit(x_fit, y)
y_pred_opt, y_std_opt = skgpr.predict(x_predict, return_std=True)
plt.plot(x_fit.ravel(), y, "ro")
plt.plot(x_predict.ravel(), y_predict, label="y_true")
plt.plot(x_predict.ravel(), y_pred_opt, label="y_pred, skopt GP")
plt.xlabel("X")
plt.ylabel("Predictions")
plt.legend()
plt.show()
```

If you run the code a number of times, you'll get plots ranging from 
![noise_1](https://cloud.githubusercontent.com/assets/1867024/19708997/f9002ab8-9af0-11e6-9341-d936a021623d.png)

to

![noise_2](https://cloud.githubusercontent.com/assets/1867024/19709014/1c537a6a-9af1-11e6-8061-36327838e697.png)

[...]

PR #256, opened 2016-10-26
[MRG] Fix constant prediction in bayesian optimisation

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/252 partially


MechCoder:
Should we stop promising the optimisation of noisy functions until we are clear about what we are doing?

glouppe:
Maybe we should also have more than 10 iterations, since the last plot is not very convincing at the moment (it is far from having converged).

MechCoder:
I tried that but it gives a flat prediction after that.

skopt/optimizer/gp.py
```diff
@@ -149,6 +150,14 @@ def gp_minimize(func, dimensions, base_estimator=None,
         Controls how much improvement one wants over the previous best
         values. Used when the acquisition is either `"EI"` or `"PI"`.
 
+    * `noise` [float, default="gaussian"]:
+        - If `noise="gaussian"` the function is noisy and returns
+          being `f(x_i) + e(i)` where e(i) is zero centered and has a fixed
+          variance.
+        - If the variance is known before-hand, this can be set directly
+          to the variance of the noise.
+        - Set this to zero if the function is non-noisy.
```
betatim: Set this to zero is the objective is noise free? (non-noisy seems odd)

betatim:
(Nitpick: could you remove `plt.rcParams["figure.figsize"] = (10, 6)` at the top of the notebook so we get smaller figures?)

MechCoder:
I'll do making the plots smaller just before the release for all examples together. I've addressed everything else.

betatim:
Thanks a lot!

glouppe:
:beers:

Issue #253, opened 2016-10-25
Smaller figures in examples


betatim:
The figures in our example notebooks are very large. Making them a bit smaller will make things more readable.

PR #258, opened 2016-10-27
Cosmetic fixes to examples

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/253


betatim:
Can't see the diff in the browser but it looks like all you changed is the figure size. If the answer is yes, feel free to merge.

MechCoder:
Thanks but however the output from `plot_evaluations` and `plot_objective` still does not scale to a single page because it seems that the functions return an array of axes?

betatim:
They set the [figure size "by hand"](https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/plots.py#L299). The idea was that you want a minimum size for each square in the grid otherwise it becomes impossible to look at.

Issue #193, opened 2016-08-10
Pseudo-random sampling for random start points


glouppe:
See discussion in #187

glouppe:
Note that this can be used in different places:
- in `dummy_minimize`
- as an alternative to `search="sampling"`
- to provide initial points to the solver when `search="lbfgs"`

betatim:
Another strategy we should consider (for tree based methods?) is creating samples at the edges/boundaries of the search space as the initialisation points. If your minimum is close to an edge this will help, in other cases you are wasting samples.

MechCoder:
Having a look right now.

betatim:
Probably related to #250

glouppe:
Pseudo random sampling did not prove to be worth the trouble over simpler uniform random sampling (see #262).

Closing for now.

PR #262, opened 2016-11-02
[WIP] Pseudo-random sampling

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/193

### DIFF
--- a/benchmarks/bench_branin.py
+++ b/benchmarks/bench_branin.py
@@ -1,3 +1,6 @@
+from time import time
+from functools import partial
+
 import numpy as np
 
 from skopt.benchmarks import branin
@@ -17,16 +20,19 @@ for name, optimizer in optimizers:
 
     for random_state in range(10):
         print(random_state)
+        t = time()
         if name == "gp_minimize":
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls,
-                noise=1e-10)
+                noise=1e-10, rvs="pseudo_uniform", acq_optimizer="spearmint")
         else:
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls)
+        time_ += time() - t
         results.append(res)
         min_func_calls.append(np.argmin(res.func_vals) + 1)
 
+    print(time_ / 10)
     optimal_values = [result.fun for result in results]
     mean_optimum = np.mean(optimal_values)
     std = np.std(optimal_values)
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -31,8 +31,8 @@ def base_minimize(func, dimensions, base_estimator,
                   n_calls=100, n_random_starts=10,
                   acq_func="EI", acq_optimizer="auto",
                   x0=None, y0=None, random_state=None, verbose=False,
-                  callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96, n_jobs=1):
+                  callback=None, n_points=10000, n_restarts_optimizer=20,
+                  xi=0.01, kappa=1.96, n_spray_points=10, n_jobs=1):
     """
     Parameters
     ----------
@@ -257,7 +257,7 @@ def base_minimize(func, dimensions, base_estimator,
             acq_optimizer = "lbfgs"
         else:
             acq_optimizer = "sampling"
-    elif acq_optimizer not in ["lbfgs", "sampling"]:
+    elif acq_optimizer not in ["lbfgs", "sampling", "spearmint"]:
         raise ValueError(
             "Expected acq_optimizer to be 'lbfgs', 'sampling' or 'auto', "
             "got %s" % acq_optimizer)
@@ -277,7 +277,7 @@ def base_minimize(func, dimensions, base_estimator,
 
         models.append(gp)
 
-        if acq_optimizer == "sampling":
+        if acq_optimizer == "sampling" or acq_optimizer == "spearmint":
             X = space.transform(space.rvs(n_samples=n_points,
                                           random_state=rng))
             values = _gaussian_acquisition(
@@ -285,10 +285,31 @@ def base_minimize(func, dimensions, base_estimator,
                 xi=xi, kappa=kappa)
             next_x = X[np.argmin(values)]
 
-        elif acq_optimizer == "lbfgs":
+        if acq_optimizer == "lbfgs" or acq_optimizer == "spearmint":
             best = np.inf
-            x0 = space.transform(space.rvs(n_samples=n_restarts_optimizer,
-                                           random_state=rng))
+
+            if acq_optimizer == "lbfgs":
+                x0 = space.transform(space.rvs(n_samples=n_restarts_optimizer,
+                                               random_state=rng))
+            else:
+                std = 1e-3
+
+                best_X_trans = space.transform([Xi[np.argmin(yi)]])
+
+                # Sample points close to the current minimum.
+                X_spray = best_X_trans + std * rng.randn(
+                    n_spray_points, space.transformed_n_dims)
+                bounds = np.array(space.transformed_bounds)
+
+                # Clip X_spray to be in-between bounds.
+                X_spray = np.minimum(
+                    np.maximum(X_spray, bounds[:, 0]), bounds[:, 1])
+                spray_vals = _gaussian_acquisition(
+                    X=X_spray, model=gp,  y_opt=np.min(yi), acq_func=acq_func,
+                    xi=xi, kappa=kappa)
+                X = np.vstack((X, X_spray))
+                values = np.concatenate((values, spray_vals))
+                x0 = X[np.argsort(values)[:n_restarts_optimizer]]
 
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
@@ -308,7 +329,6 @@ def base_minimize(func, dimensions, base_estimator,
             if a < best:
                 next_x, best = cand_xs[best_ind], a
 
-        # lbfgs should handle this but just in case there are precision errors.
         next_x = np.clip(
             next_x, transformed_bounds[:, 0], transformed_bounds[:, 1])
         next_x = space.inverse_transform(next_x.reshape((1, -1)))[0]
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -15,8 +15,8 @@ def gp_minimize(func, dimensions, base_estimator=None,
                 n_calls=100, n_random_starts=10,
                 acq_func="EI", acq_optimizer="auto", x0=None, y0=None,
                 random_state=None, verbose=False, callback=None,
-                n_points=10000, n_restarts_optimizer=5, xi=0.01, kappa=1.96,
-                noise="gaussian", n_jobs=1):
+                n_points=10000, n_restarts_optimizer=20, xi=0.0, kappa=1.96,
+                noise="gaussian", rvs="uniform", n_spray_points=10, n_jobs=1):
     """Bayesian optimization using Gaussian Processes.
 
     If every function evaluation is expensive, for instance
@@ -189,8 +189,8 @@ def gp_minimize(func, dimensions, base_estimator=None,
     rng = check_random_state(random_state)
 
     # To make sure that GP operates in the [0, 1] space
-    dimensions = [check_dimension(d,
-                                  transform="normalize") for d in dimensions]
+    dimensions = [
+        check_dimension(dim, transform="normalize", rvs=rvs) for dim in dimensions]
     space = Space(dimensions)
 
     # Default GP
@@ -203,7 +203,8 @@ def gp_minimize(func, dimensions, base_estimator=None,
 
         base_estimator = GaussianProcessRegressor(
             kernel=cov_amplitude * matern,
-            normalize_y=True, random_state=rng, alpha=0.0, noise=noise)
+            normalize_y=True, random_state=rng, alpha=0.0, noise=noise,
+            n_restarts_optimizer=2)
 
     return base_minimize(
         func, dimensions, base_estimator=base_estimator,
@@ -212,4 +213,4 @@ def gp_minimize(func, dimensions, base_estimator=None,
         n_points=n_points, n_random_starts=n_random_starts,
         n_restarts_optimizer=n_restarts_optimizer,
         x0=x0, y0=y0, random_state=random_state, verbose=verbose,
-        callback=callback, n_jobs=n_jobs)
+        callback=callback, n_spray_points=n_spray_points, n_jobs=n_jobs)
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -13,6 +13,7 @@ from .transformers import Normalize
 from .transformers import Identity
 from .transformers import Log10
 from .transformers import Pipeline
+from ..utils import pseudo_uniform
 
 # helper class to be able to print [1, ..., 4] instead of [1, '...', 4]
 class _Ellipsis:
@@ -20,7 +21,7 @@ class _Ellipsis:
         return '...'
 
 
-def check_dimension(dimension, transform=None):
+def check_dimension(dimension, transform=None, rvs="uniform"):
     """
     Checks that the provided dimension falls into one of the
     supported types. For a list of supported types, look at
@@ -56,7 +57,7 @@ def check_dimension(dimension, transform=None):
     if (len(dimension) == 3 and
         isinstance(dimension[0], numbers.Real) and
         isinstance(dimension[2], str)):
-          return Real(*dimension, transform=transform)
+          return Real(*dimension, transform=transform, rvs=rvs)
 
     if len(dimension) > 2 or isinstance(dimension[0], str):
         return Categorical(dimension)
@@ -65,7 +66,7 @@ def check_dimension(dimension, transform=None):
         return Integer(*dimension, transform=transform)
 
     if isinstance(dimension[0], numbers.Real):
-        return Real(*dimension, transform=transform)
+        return Real(*dimension, transform=transform, rvs=rvs)
     raise ValueError("Invalid dimension %s. Read the documentation for "
                      "supported types." % dim)
 
@@ -117,7 +118,7 @@ class Dimension(object):
 
 
 class Real(Dimension):
-    def __init__(self, low, high, prior="uniform", transform=None):
+    def __init__(self, low, high, prior="uniform", transform=None, rvs="uniform"):
         """Search space dimension that can take on any real value.
 
         Parameters
@@ -143,6 +144,10 @@ class Real(Dimension):
         self.high = high
         self.prior = prior
         self.transform_ = transform
+        if rvs == "uniform":
+            _rvs = uniform
+        else:
+            _rvs = pseudo_uniform
 
         if self.transform_ and self.transform_ != "normalize":
             raise ValueError(
@@ -153,7 +158,7 @@ class Real(Dimension):
         # The rvs on Dimension calls inverse_transform on the points sampled
         # using _rvs
         if self.transform_ == "normalize":
-            self._rvs = uniform(0, 1)
+            self._rvs = _rvs(0, 1)
             if self.prior == "uniform":
                 self.transformer = Pipeline(
                     [Identity(), Normalize(low, high)])
@@ -163,10 +168,10 @@ class Real(Dimension):
                 )
         else:
             if self.prior == "uniform":
-                self._rvs = uniform(self.low, self.high - self.low)
+                self._rvs = _rvs(self.low, self.high - self.low)
                 self.transformer = Identity()
             else:
-                self._rvs = uniform(
+                self._rvs = _rvs(
                     np.log10(self.low),
                     np.log10(self.high) - np.log10(self.low))
                 self.transformer = Log10()
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -1,8 +1,10 @@
 from copy import deepcopy
 import numpy as np
 from scipy.optimize import OptimizeResult
+
 from sklearn.externals.joblib import dump as dump_
 from sklearn.externals.joblib import load as load_
+from sklearn.utils import check_random_state
 
 __all__ = (
     "load",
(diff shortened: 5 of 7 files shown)


skopt/utils.py
```diff
@@ -122,3 +124,383 @@ def load(filename, **kwargs):
...
+
+    def rvs(self, size=1, random_state=None):
+        rng = check_random_state(random_state)
+
+        # Generate random seed
+        seed = rng.randint(1000)
+        return self.loc + self.scale * i4_sobol_generate(1, size, seed).ravel()
+
+
+# Copied shamelessly from http://people.sc.fsu.edu/~jburkardt/py_src/sobol/sobol.html
```
glouppe: Is this compatible with their license?

skopt/utils.py
```diff
@@ -122,3 +124,383 @@ def load(filename, **kwargs):
...
+    global dim_max
+    global dim_num_save
+    global initialized
+    global lastq
+    global log_max
+    global maxcol
+    global poly
+    global recipd
+    global seed_save
+    global v
```
glouppe: yuck

skopt/utils.py
MechCoder: It is "MIT licensed". I think it is compatible with BSD. (http://softwareengineering.stackexchange.com/q/121998)

skopt/utils.py
betatim: If we decide to keep it we should clean it up a bit :)

glouppe:
I was looking at Spearmint implementation again. They are actually using this strategy when optimizing the aquisition function:
- Initialize a grid of points with a sobol sequence (https://github.com/HIPS/Spearmint/blob/e719c56cee9eefda5df6af40f834c9fa310e23a5/spearmint/choosers/default_chooser.py#L259)
- Extend this list with random points close from the current best point (https://github.com/HIPS/Spearmint/blob/e719c56cee9eefda5df6af40f834c9fa310e23a5/spearmint/choosers/default_chooser.py#L339)
- For all these points (i.e, many of them!!), evaluate the acquisition function EI  (https://github.com/HIPS/Spearmint/blob/e719c56cee9eefda5df6af40f834c9fa310e23a5/spearmint/choosers/default_chooser.py#L343)
- Run the actual optimization of the acquisition using the 20 best points as initial points (https://github.com/HIPS/Spearmint/blob/e719c56cee9eefda5df6af40f834c9fa310e23a5/spearmint/choosers/default_chooser.py#L375)

We should certainly do the same. I presume this coarse brute-force like optimization is helping a lot to locate a good enough location from which the optimizer should be started.

glouppe:
Also instead of adding and maintaining code for sobol sequences, we could maybe consider using instead a Python module that already implements it. I found https://github.com/naught101/sobol_seq, which is available in pip as `sobol_seq`.

betatim:
Do we need a Sobol sequence? My understanding of them is that they solve the problem of "real" random numbers clumping together. They are a low discrepancy sequence. You can draw a new sample from them and it will be "as far away" as possible from the existing ones. So people use them for MC integration were it is not known upfront how many samples will be drawn and you want to cover the space evenly. If you decide/know up front how many samples you will draw, you can make a regular grid and be done with it.

Unless we really need it I'd be -1 on adding the code or a new dependency. There might be something I am missing about the spearmint strategy or we might want to adopt a different strategy so please point that out to me :)

glouppe:
Yes, for the coarse scan, a regular grid should be enough, though this sobol sequence does that nicely for us.

glouppe:
(in particular, defining that regular grid for n points in R^d is not that immediate)

MechCoder:
I'm also -1 for adding an additional dependency, the code from https://github.com/naught101/sobol_seq is also a copy.

glouppe:
I am -1 for copy pasting code we dont own and dont understand. The less code we have, the less we have to maintain.

MechCoder:
OK, let us debate the merits and demerits of doing that, after we convince ourselves that it is useful through experiments.

skopt/optimizer/base.py
```diff
@@ -263,20 +263,33 @@ def base_minimize(func, dimensions, base_estimator,
...
+                n_spray_points = 20
+                std_ = 1e-3
+                best_X = Xi[np.argmin(yi)]
+
+                X_spray = best_X + rng.randn(n_spray_points, space.transformed_n_dims)
+
+                bounds = np.array(space.transformed_bounds)
+                candidates = np.minimum(np.maximum(X_spray, bounds[:, 0]), bounds[:, 1])
+                X = np.vstack((X, candidates))
+                x0_all = X[np.argsort(values)[:n_restarts_optimizer]]
```
glouppe: One thing that will certainly be important. I think `n_spray_points` should be dimensioned such that `n_spray_points < n_restarts_optimizer`, to be sure we run the optimization outside of that spray area as well. In spearmint, `n_spray_points=10` and `n_restarts_optimizer=20`.

MechCoder: sure, thanks!

MechCoder: I'm setting `n_restarts_optimizer` to 10 and `n_spray_points` to 5 for now considering how expensive it is.

MechCoder:
Here are my benchmarks. Doesn't look too promising:

On Branin:

|Acquisition Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min | Time | 
------------------|------------|-----------|---------------------|--------------------|-----------------------|-----|
lbfgs | 0.401 +/- 0.004 | 0.398 | 60.7 | 40.591 | 24 | 28.95
pseudosampling + lbfgs |  0.401 +/- 0.004 | 0.398 | 60.6 | 43.107 | 32 | 29
spearmint (w/o pseduosampling) | 0.398 +/ 0.000 | 0.398 | 71.2 | 22.203 | 32 | 188.98
spearmint (with pseduosampling) | 0.398 +/ 0.000 | 0.398 | 118.7 | 23.199 | 88 | 190

On Hart6:

|Acquisition Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min | Time | 
------------------|------------|-----------|---------------------|--------------------|-----------------------|-----|
lbfgs | -3.095 +/- 0.222 | -3.311 | 73.3| 50.465 |31 | 322
pseudosampling + lbfgs | -3.146 +/- 0.110 | -3.298 | 99.2 | 42.107 | 34 | 379.21
spearmint (w/o pseduosampling) | -3.230 +/ 0.057 | -3.321 | 84.5 | 54.57 | 31 | 1145.35
spearmint (with pseduosampling) | -3.112 +/ 0.050 | -3.192 | 90.9 | 46.580 | 23 | 2362.57

skopt/optimizer/base.py
```diff
@@ -263,20 +263,32 @@ def base_minimize(func, dimensions, base_estimator,
...
-                x0 = space.transform(space.rvs(n_samples=1,
-                                               random_state=rng))[0]
+            if acq_optimizer == "lbfgs":
+                x0_all = space.transform(space.rvs(n_samples=n_restarts_optimizer,
+                                         random_state=rng))
+            else:
+                std_ = 1e-3
+                best_X = Xi[np.argmin(yi)]
+
+                X_spray = best_X + rng.randn(n_spray_points, space.transformed_n_dims)
```
betatim: We should scale the std used by `randn` to match the bounds of each dimension or `space.transform` the samples before adding them to `best_X`

betatim:
Could you write a little interpretation of what the results mean to you?

I'm looking at the spread (for hart6) on the minimum we find and comparing it to spearmint -> our spread is 0.222 compared to their 0.05 -> we seem to sometimes get it right and sometimes we are miles off -> we are just getting lucky?

pseudosampling + lbsgs has a smaller spread -> we are relying less on luck/find the minimum more consistently -> so even though we don't find the minimum in less steps it is overall better?

glouppe:
Do you use the same value of `n_restarts_optimizers` in all cases? (we should if we want to make a fair comparison and determine whether this strategy is better)

glouppe:
Could you also add a comparison for the case where the optimizer is `"sampling"` with a value of `n_points` identical to what you use in the case of spearmint? Unless there is a bug somewhere we should at least see comparable runtimes.

skopt/optimizer/base.py
glouppe: I guess this is what `std_` above was meant to be used for, but it seems this variable is left unused.

benchmarks/bench_branin.py
```diff
@@ -17,13 +18,15 @@
...
+        t = time()
         if name == "gp_minimize":
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls,
-                noise=1e-10)
+                noise=1e-10, rvs="pseudo_uniform", acq_optimizer="spearmint")
         else:
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls)
+        time_ += time() - t
```
glouppe: this is never used, is it?

skopt/optimizer/base.py
```diff
@@ -263,20 +263,32 @@ def base_minimize(func, dimensions, base_estimator,
...
+
+                bounds = np.array(space.transformed_bounds)
+                candidates = np.minimum(np.maximum(X_spray, bounds[:, 0]), bounds[:, 1])
+                X = np.vstack((X, candidates))
+                x0_all = X[np.argsort(values)[:n_restarts_optimizer]]
```
glouppe: Am I missing something or you never evaluated the acquisition function at `candidates`? In particular, `values` only contains values from the random sampling...

glouppe:
Could you also share your benchmark script please?

betatim:
> Could you also share your benchmark script please?

Should be in `benchmarks/` no?

glouppe:
One other thing, it seems spearmint sets `xi=0.0` when evaluating the EI.

glouppe:
Do you have an idea why the spearmint strategy appears so much slower than lbfgs? For both, LBFGS is run the same number of times. Then the additional computations come from generating the sobol grid and evaluating EI through the GP. I guess the slow down comes from this step, as it basically require computing a cholesky decomposition for a large matrix (because of `n_points=10000`).

glouppe:
Just had a look to `gpr.py`. Why dont we precompute at fit these values? (L291+)

```
                L_inv = solve_triangular(self.L_.T, np.eye(self.L_.shape[0]))
                K_inv = L_inv.dot(L_inv.T)
```

As far as I understand, they dont depend on the test data, and therefore could be compruted. But that is maybe not the meat of the computation here...

glouppe:
One more thing. I just did some debugging and looked at `values`. In many cases, it seems we have again this issue of completely flat GP. Setting `n_restarts_optimizer > 0` in `GPRegressor` partially solves this issue. 

Even if this slows things down, I would really be in favour of increasing that value. Given that in real cases, evaluating the function should takes minutes-hours, I dont mind spending some more second in properly fitting the GP.

skopt/optimizer/base.py
```diff
@@ -30,8 +30,8 @@ def base_minimize(func, dimensions, base_estimator,
                   n_calls=100, n_random_starts=10,
                   acq_func="EI", acq_optimizer="auto",
                   x0=None, y0=None, random_state=None, verbose=False,
-                  callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96):
+                  callback=None, n_points=10000, n_restarts_optimizer=10,
```
glouppe: `n_restarts_optimizer` should also be updated in `gp_minimize`!!!! otherwise

MechCoder:
I'll address your comments soon. I was completely convinced that it will not work but seems like my obvious mistakes does not help :P

glouppe:
Any progress on this Manoj? Very curious to see the benchmarks with all these little things updated :)

MechCoder:
> I guess the slow down comes from this step, as it basically requires computing a cholesky decomposition for a large matrix (because of n_points=10000)

Yes, I agree. I have noticed in tests that most of the times just the "sampling" strategy is slower than the "lbfgs"!

MechCoder:
(Average of branin across 5 runs)

|Acquisition Function| Minimum | Best minimum | Mean calls to min | Fastest calls to min | Time |
------------------|------------|-----------|---------------------|--------------------|-----------------------|-----|
lbfgs | 0.398 +/- 0.000 | 0.398 | 81.2 +/ 24.72| 42 | 76.61
sampling | 0.398 +/- 0.000 | 0.398 | 56.2 +/  23.02 | 38 | 167.55
pseudosampling | 0.398 +/- 0.000 | 0.398 | 82.2 +/  34.30 | 45 | 172.19
spearmint (lbfgs + sampling) | 0.398 +/ 0.000 | 0.398 | 53.8 +/ 4.72 | 47 | 191.18
spearmint (lbfgs + pseudosampling) | 0.398 +/ 0.000 | 0.398 | 114.4 +/ 47.99 | 58 | 203.

(I reran the benchmarks for lbfgs + sampling again just to be sure)

MechCoder:
The benchmarks can be reproduced by setting `rvs` as `uniform` / `pseudo_uniform` and `acq_optimzer` as `sampling`, `lbfgs`, `spearmint`.

MechCoder:
Future steps involve seeing if the superior performance of `lbfgs + sampling` is because of the better restarts or the addition of spray points.

glouppe:
When you do spearmint (lbfgs + sampling), the only difference with doing instead pseudosampling is that you draw the points uniformly at random instead of using the sobol sequence, right? 

Also, could you run your benchmark on a more difficult problem?

glouppe:
From the results, spearmint (lbfgs + sampling) looks very good in my opinion. It yields the lowest mean number of calls *and* the lowest standard deviation (by quite a margin). This is very promising, but would be good to evaluate on a more difficult problem to see if that holds.

MechCoder:
> When you do spearmint (lbfgs + sampling), the only difference with doing instead pseudosampling is that you draw the points uniformly at random instead of using the sobol sequence, right? 

Yes.

MechCoder:
Benchmarks on brain (average across 5 runs). I removed the time since I was doing a lot of data-intensive operations in the side.

|Acquisition Function| Minimum | Best minimum | Mean calls to min | Fastest calls to min | Time |
------------------|------------|-----------|---------------------|--------------------|-----------------------|-----|
lbfgs | -3.207 +/ 0.075 | -3.281 | 90.8 +/ 64.22 | 30 | 
sampling | -3.207 +/ 0.047 | -3.295 | 127.4 +/ 61.242 | 25 |
pseudo_sampling | -2.998 +/ 0.088 | -3.158 | 146.2 +/ 30.55 | 96 |
spearmint (lbfgs + uniform) | -3.251 +/ 0.058 | -3.224 | 148.0 +/ 51.712 | 73 |
spearmint (lbfgs + pseudo_uniform) | -3.299 +/ 0.047 | -3.224 | 134.0 +/ 30.47 | 100 | 

Not sure how to interpret these numbers except that the spearmint strategy returns a better minimum. Maybe it makes sense to measure the (mean number of iterations / time) to reach a fixed number since the minimum clearly varies? Also I"m searching for other good benchmark problems.

glouppe:
For branin, the minima are fixed and dont vary. We should in priority measure the distance to these known minima. I would say here that spearmint (lbfgs + pseudo_uniform) wins, followed by spearmint (lbfgs + uniform).

Also I am not quite sure the "mean calls to min" is really a good metric. We should rather see the problem as having a budget of number of calls and then evaluate what is the minimum attained within that budget. It is not that important whether it is reached soon or late. 

What do you think?

glouppe:
Also I think you are evaluating on hart6, not branin :)

glouppe:
Hmm, something I dont quite get. What is "best minimum"? and how this value could be worse than "minimum" in the cases of spearmint?

MechCoder:
Yes, I agree that in case we should fix the budget as the number of calls and measure the minimum obtained within this budget. But if we see that the minimum is reached earlier we can always reduce the budget right? (i.e change the default number of `n_calls`, nothing comes free)

Also should we have more hyper-parameter sensitive benchmarks and not rely just on these two benchmarks for making decisions from now on?

MechCoder:
There are a number of benchmarks in the original paper (https://arxiv.org/pdf/1206.2944.pdf) under section 4. (Empirical Analyses). I'll try to have a look over the next few days.

betatim:
Training a clf on `make_classification`? Might be harder in the sense that you have mixed dimensions but also takes a while to evaluate.

Right now we do measure how close to the minimum each optimizer gets no? Just that the budget is very generous compared to the hardness of the problem. We should measure the spread on the minimum found to help us judge how much the algorithm relies on luck or maybe how "consistent" it is.

Can we add a purely random strategy to the comparison as well? This would give us a benchmark for the benchmarks: how much better than random is the strategy?

MechCoder:
I was thinking of hacking this to make it work (https://github.com/automl/HPOlib-logreg). It seems to also be a standard benchmark for Bayesian-optimisation problems.

I agree that `200` calls is a bit too many, I stuck with that because that was the one described in the original paper. Maybe we should cut it down to say 50 and allow 1 random start?

glouppe:
I wouldnt change `n_random_starts`. But I am fine with reducing the budget to 50 or 100 iterations.

MechCoder:
Here are the benchmarks when branin and hart6 with `n_calls=50`, averaged across 10 runs.

Branin:

|Acquisition Function| Minimum | Best minimum | Time |
|------------------|------------|-----------|---|
| lbfgs | 0.403 +/ 0.005 | 0.398 | 33.885
| uniform + sampling | 0.400 +/ 0.003 | 0.398 | 49.747
| pseudo_uniform + sampling |  0.402 +/ 0.009 | 0.399 | 57.640 |
| uniform + spearmint | 0.400 +/ 0.004 | 0.398 | 47.894
| pseudo_uniform + spearmint | 0.400 +/ 0.004 | 0.398 | 63.014

Hart6

|Acquisition Function| Minimum | Best minimum | Time |
|------------------|------------|-----------|---|
| lbfgs | -3.048 +/ 0.177 | -3.297 | 15.227
| uniform + sampling | -3.138 +/ 0.171 | -3.322 | 16.210
| pseudo_uniform + sampling |  -3.012 +/ 0.325 | -3.322 | 63.307 |
| uniform + spearmint | -3.090 +/ 0.215 | -3.322 | 23.503
| pseudo_uniform + spearmint | -3.138 +/ 0.194 | -3.322 | 77.941

MechCoder:
From just these two benchmarks, maybe we can infer that just doing a pseudo-uniform sampling and lbfgs are not great strategies?

glouppe:
> From just these two benchmarks, maybe we can infer that just doing a pseudo-uniform sampling and lbfgs are not great strategies?

Agree. Also we should expect that just doing sampling cannot be the best strategy. Therefore, I would advocate we switch to a spearmint-like procedure for optimizing the acquisition function. From hart6, it seems pseudo-random numbers work better, which is also consistent with the actual spearmint implementation. What do you think?

MechCoder:
The only reason that I'm slightly reluctant is that I would like to test with at least a couple of more benchmarks to make sure.

The logistic regression benchmark that I've described above uses the number of epochs and batch size as hyperparameters as well. Since lbfgs acts on the one-hot encoded transformed space, it might not be fair to make a comparison for that right now.

We should fix https://github.com/scikit-optimize/scikit-optimize/pull/62 first?

glouppe:
+1 for trying on a third benchmark problem. Not sure going all the way up to properly support categorical variables is necessary here (that is, for the purpose of determining whether this optimization of the acquisition is indeed better). I would suggest trying on a problem with non-categorical variables only.

MechCoder:
I wrote a wrapper around 2-layer MLP in tensorflow, the script can be seen here (https://github.com/MechCoder/Notebooks/blob/master/MLP_tensorflow_skopt_on_mnist.ipynb). I am running it on the clusters and will update once I get the results.

I just tune the regularization, initialisation scale and the learning rate. Are there any other interesting continuous parameters?

glouppe:
> I just tune the regularization, initialisation scale and the learning rate. Are there any other interesting continuous parameters?

Sounds good

MechCoder:
My job stopped just before it was about to end because of insufficient memory :/ Restarted it after requesting much more memory.

glouppe:
When calling `lbfgs`, shall we remove `maxiter=20`? (it is not set in spearmint). 

Shall we also set `n_restarts_optimizer=20` and `n_spray=10` to have the same settings as they have?

MechCoder:
MNIST seems to be a really easy problem. All of them reach a minimum of 1.0 within 50 calls. I shall try with a different dataset (CIFAR 10)

I don't think we should increase the `maxiter` parameter. It seems to be an overkill to me especially since we have set `n_restarts_optimizer` as well.

glouppe:
Fair enough. Can we still switch to 20 for n_restarts_optimizers and 10 for n_spray? We shoud really make everything necessary to not waste calls to the objective.

MechCoder:
I ran a benchmark for the CIFAR dataset using keras before I realized that (https://github.com/scikit-optimize/scikit-optimize/pull/279) might be creating a non-trivial impact because of the different scales of the Integer and Real dimensions.

Also could someone do a sanity check of the code before I run any more benchmarks?

benchmarks/bench_branin.py
```diff
@@ -1,3 +1,6 @@
+from time import time
+from functools import partial
```
glouppe: this import is never used

benchmarks/bench_branin.py
```diff
@@ -17,16 +20,19 @@
...
+        t = time()
         if name == "gp_minimize":
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls,
-                noise=1e-10)
+                noise=1e-10, rvs="pseudo_uniform", acq_optimizer="spearmint")
         else:
             res = optimizer(
                 branin, bounds, random_state=random_state, n_calls=n_calls)
+        time_ += time() - t
```
glouppe: `time_` is not initialized

benchmarks/bench_branin.py
```diff
@@ -17,16 +20,19 @@
...
         results.append(res)
         min_func_calls.append(np.argmin(res.func_vals) + 1)
 
+    print(time_ / 10)
```
glouppe: Here you report the average time, but below in `bench_hart6.py` you report the total time.

skopt/optimizer/base.py
```diff
@@ -274,18 +274,39 @@ def base_minimize(func, dimensions, base_estimator,
...
+        if acq_optimizer == "lbfgs" or acq_optimizer == "spearmint":
             best = np.inf
-            x0 = space.transform(space.rvs(n_samples=n_restarts_optimizer,
-                                           random_state=rng))
+
+            if acq_optimizer == "lbfgs":
+                x0 = space.transform(space.rvs(n_samples=n_restarts_optimizer,
+                                               random_state=rng))
+            else:
+                std = 1e-3
```
glouppe: Shall we make this a parameter of the method and rename it to `spray_std` or something better?

skopt/space/space.py
```diff
@@ -143,6 +144,10 @@ def __init__(self, low, high, prior="uniform", transform=None):
         self.high = high
         self.prior = prior
         self.transform_ = transform
+        if rvs == "uniform":
+            _rvs = uniform
+        else:
```
glouppe: can we make this explicit and switch to pseudo uniform only if `rvs == "pseudo-uniform"`?

skopt/utils.py
glouppe: We still need to decide about this.

skopt/optimizer/gp.py
```diff
@@ -15,8 +15,8 @@ def gp_minimize(func, dimensions, base_estimator=None,
                 n_calls=100, n_random_starts=10,
                 acq_func="EI", acq_optimizer="auto", x0=None, y0=None,
                 random_state=None, verbose=False, callback=None,
-                n_points=10000, n_restarts_optimizer=5, xi=0.01, kappa=1.96,
-                noise="gaussian", n_jobs=1):
+                n_points=10000, n_restarts_optimizer=20, xi=0.0, kappa=1.96,
+                noise="gaussian", rvs="uniform", n_spray_points=10, n_jobs=1):
```
glouppe: We should discuss on the default values of `acq_optimizer` and `rvs`. Not sure about exposing `rvs` either. Opinions?

MechCoder:
Sorry, I cannot run any GPU based benchmarks till the start of next week since there seems to be a limit for all users and I require it for my projects. After Monday, I can run the benchmarks in parallel for all the (acq_optimizer and rvs) combinations to decide the default values.

Also there seems to be a bug in the keras cnn script here (https://github.com/fchollet/keras/issues/4587) which makes the default SGD optimizer perform bad. I can replace it with the Adam Optimizer in the next set of benchmarks which is likely to give better results along with the bug-fix in the normalization of Integer dimensions.

MechCoder:
I am rerunning the benchmarks for "pseudorandom" and "spearmint" after rebasing on master and using the adam optimizer. Hopefully if all is well, will post the results in 24 hours and rerun the other combinations in parallel.

MechCoder:
The results don't look great. I used the cross-entropy on the validation dataset as the minimization objective. You can have a look at the script here (https://github.com/MechCoder/Notebooks/blob/master/bench_keras_cifar10.py) which is mainly a modification of the original keras cnn-model here (https://github.com/fchollet/keras/blob/master/examples/cifar10_cnn.py)

Here (https://gist.github.com/MechCoder/bd8c1cf5b2681b06569cfe4773265fb2) are the objectives at each of the 50th call., the first ten are at random. The minimum is obtained at the 7th call, which is at random.

MechCoder:
Here are plots of some of the integer hyperparameters against the function calls.

batch_size
epochs
filters1

It seems that all the optimum points for these Integer hyperparameters are either at the upper bound and lower bound. Anything obviously wrong with how we handle integer dimensions?

glouppe:
Hmmm, could be many things... 
- an issue due to hyper-parameters that are mostly irrelevant, hence making the optimization more difficult.
- an issue with the gp
- an issue due to scaling integers: for this can you try instead using only `Real` bounds and then casting back to `int` within the objective function? or can you evaluate your benchmarks only on the continuous hyper-parameters? (dropout rates, learning rate, and you can add epsilon in Adam)

In any case, I am not sure what you observe here is due to the sampling strategy of the starting points... I would be glad if we could settle this issue once and for all and move on.

MechCoder:
Here are the results for varying the three dropouts, learning rate and epsilon. (5 hyperparameters) I shall update the remaining two after the runs are finished. (~12 hours from now). The value here is the mean validation cross-entropy loss on the cross-validation data. For the benchmarks, it seems sampling is the best strategy ?!

| Acquisition Function | Minimum | Time |
| --- | --- | --- |
| lbfgs |  |  |  |
| uniform + sampling | 0.589 |  38402.7 |
| pseudo_uniform + sampling |  | |  |
| uniform + spearmint | 0.7359 | 40740.49 |
| pseudo_uniform + spearmint | 0.639 | 43089.46

MechCoder:
Seems to concur with my benchmarks here (https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-260835534)? Also should we change the default from lbfgs to sampling? I'll wait till the other benchmarks finish though.

MechCoder:
(updated)

MechCoder:
The benchmarks seem to suggest that doing anything fancier than random sampling does not perform better (at least on these two problems).
I would like https://github.com/scikit-optimize/scikit-optimize/pull/281 to be merged and this to be kept open for people to try it out if possible. I would also like to move on given the amount of time spent on this pull request and the not so obvious benefits.
WDYT?

glouppe:
Geh. I really dont understand. How could spearmint be worse than random sampling? 

What you call "uniform + sampling" is simply brute forcing the acquisition function right? Why would it be worse if you then run lbfgs from these top points? I should only improve. 

For how many runs do you perform your last benchmarks? Reading the discussion again, it seems that for branin and hart6, combining pseudo-uniform sampling of the starting points + lbfgs was doing quite well...

glouppe:
> The benchmarks seem to suggest that doing anything fancier than random sampling does not perform better (at least on these two problems).
I would like #281 to be merged and this to be kept open for people to try it out if possible. I would also like to move on given the amount of time spent on this pull request and the not so obvious benefits.
WDYT?

I disagree. Your results from Nov 14 show otherwise.

glouppe:
An opinion on this @betatim ? Would be good to decide something and move. 

I am personnally in favour of something close to uniform sampling of random initial points (many), evaluating the acquisition functions there, then starting lbfgs at the top 20 of them.

MechCoder:
Which results are you referring to? This (https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-260252061)? The best minimun obtained is the same no?

glouppe:
Yes, but the mean number of calls is smaller, and variance is much smaller
too. It is indeed a very good thing to not have to rely on luck to hope
getting the best results.

(sent from my phone)

MechCoder:
But please have a look at the benchmarks on hart6 here (https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-260835534), both the mean and standard deviation of the minimum obtained.

MechCoder:
> I cant believe running lbfgs on the top points can make things worse.

I don't know as well but it could be a number of things, such as non-smoothness of the acquistion function or the acquisition function not modeling the original objective function.

MechCoder:
Btw, I thought we agreed here (https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-260445579) that the mean calls to the minimum is not a great metric since we fix the budget beforehand.

glouppe:
Running lbfgs can only help us find a better minimum of the acquisition
function. And we do want to optimize the acquisition function correctly. If
we don't trust it and instead rely on luck, then there is no point in
building it in the first place. It could be that the acquisition function
is bad, but this is separate issue from optimizing it. What do you think ?

(sent from my phone)

MechCoder:
Yes, I'll verify on how the acquisition function is optimized and I'll reserve my comments for after that.

glouppe:
Superseded by #284.

Issue #264, opened 2016-11-03
Improved benchmarks


MechCoder:
We need additional benchmarks that measure number of iterations taken to achieve the minimum.

PR #265, opened 2016-11-04
Add benchmarks for function calls taken to reach minimum

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/264

### DIFF
--- a/benchmarks/bench_branin.py
+++ b/benchmarks/bench_branin.py
@@ -1,5 +1,4 @@
 from functools import partial
-
 import numpy as np
 
 from skopt.benchmarks import branin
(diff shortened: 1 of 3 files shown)


MechCoder:
@glouppe @betatim Comments?

benchmarks/bench_branin.py
```diff
@@ -1,5 +1,4 @@
 from functools import partial
-
```
betatim: I liked the empty line

betatim:
Uncontroversial I think. One thing that would be nice would be to not use `argmin()` but to find the first point after which the result does not significantly improve anymore. Not sure how to do that and might be overkill.

Issue #266, opened 2016-11-04
Benchmark crashes for gp_minimize with NaN gradients


betatim:
Tried running the benchmarks and got:

```
$ python benchmarks/bench_branin.py                                     (skopt)
gp_minimize
0
Traceback (most recent call last):
  File "benchmarks/bench_branin.py", line 23, in <module>
    noise=1e-10)
  File "/Users/thead/git/scikit-optimize/skopt/optimizer/gp.py", line 207, in gp_minimize
    callback=callback)
  File "/Users/thead/git/scikit-optimize/skopt/optimizer/base.py", line 288, in base_minimize
    maxiter=20)
  File "/Users/thead/anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/lbfgsb.py", line 193, in fmin_l_bfgs_b
    **opts)
  File "/Users/thead/anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/lbfgsb.py", line 330, in _minimize_lbfgsb
    f, g = func_and_grad(x)
  File "/Users/thead/anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/lbfgsb.py", line 278, in func_and_grad
    f = fun(x, *args)
  File "/Users/thead/anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/optimize.py", line 289, in function_wrapper
    return function(*(wrapper_args + args))
  File "/Users/thead/anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/optimize.py", line 63, in __call__
    fg = self.fun(x, *args)
  File "/Users/thead/git/scikit-optimize/skopt/acquisition.py", line 16, in gaussian_acquisition_1D
    X, model, y_opt, acq_func, xi, kappa, return_grad=True)
  File "/Users/thead/git/scikit-optimize/skopt/acquisition.py", line 35, in _gaussian_acquisition
    func_and_grad = gaussian_ei(X, model, y_opt, xi, return_grad)
  File "/Users/thead/git/scikit-optimize/skopt/acquisition.py", line 209, in gaussian_ei
    return_std_grad=True)
  File "/Users/thead/git/scikit-optimize/skopt/learning/gaussian_process/gpr.py", line 264, in predict
    X = check_array(X)
  File "/Users/thead/git/scikit-optimize/src/sklearn/sklearn/utils/validation.py", line 406, in check_array
    _assert_all_finite(array)
[...]

PR #268, opened 2016-11-07
[MRG] Fix infinite gradients 

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/266


MechCoder:
@betatim @glouppe reviews? I'm no longer able to reproduce the bug reported.

MechCoder:
Updated benchmarks:

```
Branin:
Mean optimum: 0.401670639097
Std of optimal values0.00555080380794
Best optima:0.3981513906
Mean func_calls to reach min: 61.6
Std func_calls to reach min: 25.7689735923
Fastest no of func_calls to reach min: 30

Hart6:
Mean optimum: -3.03880773988
Std of optimal values0.135679431367
Best optima:-3.29936142128
Mean func_calls to reach min: 76.6
Std func_calls to reach min: 35.2510992736
Fastest no of func_calls to reach min: 31
```

skopt/learning/gaussian_process/tests/test_kernels.py
```diff
@@ -44,11 +41,27 @@ def numerical_gradient(X, Y, kernel):
...
 
 def test_gradient_correctness():
+    rng = np.random.RandomState(0)
+    X = rng.randn(5)
+    Y = rng.randn(10, 5)
     for kernel in KERNELS:
-        yield check_gradient_correctness, kernel
+        yield check_gradient_correctness, kernel, X, Y
+
+def test_gradient_finiteness():
```
betatim: + 💯 you are a superstar :) I assume this test used to fail but not anymore.

betatim: https://github.com/scikit-optimize/scikit-optimize/issues/266

skopt/learning/gaussian_process/gpr.py
```diff
@@ -309,7 +309,10 @@ def predict(self, X, return_std=False, return_cov=False,
 
                 if return_std_grad:
                     # XXX: Cache np.dot(K_trans, K_inv) from above
-                    grad_std = -np.dot(K_trans, np.dot(K_inv, grad))[0] / y_std
+                    if y_std == 0.0:
```
betatim: Should we use something like `np.allclose` instead of `==`?

skopt/learning/gaussian_process/kernels.py
```diff
@@ -120,7 +130,11 @@ def gradient_x(self, x, X_train):
             sqrt_3_dist = sqrt(3) * dist
             f = np.expand_dims(1 + sqrt_3_dist, axis=1)
 
-            dist_expand = np.expand_dims(sqrt(3) / dist, axis=1)
+            # Gradients when dist equal to zero is zero.
```
betatim: Wut? Maybe we can find a better way of formulating this (I don't get it at the moment)

skopt/learning/gaussian_process/tests/test_kernels.py
MechCoder: It fails because it used to return infinite gradients.

betatim: Nods, I was thinking of: can we setup a small problem that leads to the old calculation producing infinite gradients but the new one doesn't do that. Does that make sense? (Not sure I am doing a good job explaining it)

MechCoder: I don't understand. This test does exactly that no?

betatim: This is what I thought at first, but then became unsure. Seems like I managed to spread confusion :-/ -> we all agree and I am back to calling you a superstar 😄

betatim:
Other than the small comments above 👍. My only complaint is that you are way to fast for me I had thought I might be able to fix this, turned my back and by the time I come back it was done.

skopt/learning/gaussian_process/kernels.py
MechCoder: OK I am confused now. Some fresh eyes might be helpful.

https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/learning/gaussian_process/kernels.py#L117

The value of `f` when `X = Y` equals 1.

From here (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/learning/gaussian_process/kernels.py#L132)

this reduces to `f_grad * g * ( 1- f )`, so it should be zero right when X equals Y?

MechCoder: or maybe not, just a second.

MechCoder: I'm getting zero whichever way I think about it.

MechCoder:
@betatim Thanks for spotting that! The corner-case is not as straightforward as I had thought initially. I have added a detailed explanation. Could you have a look?

skopt/acquisition.py
```diff
@@ -221,7 +221,7 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
     values[mask] = exploit + explore
 
     if return_grad:
-        if not np.any(mask):
+        if not np.all(mask):
```
glouppe: same here

skopt/acquisition.py
```diff
@@ -150,7 +150,7 @@ def gaussian_pi(X, model, y_opt=0.0, xi=0.01, return_grad=False):
     values[mask] = norm.cdf(scaled)
 
     if return_grad:
-        if not np.any(mask):
+        if not np.all(mask):
```
MechCoder: So if mask has more than one value, we don't support it. It raises an error before this line is reached when `predict` is called.

Anyway I can make that clearer?

glouppe: Sorry but I dont get it. Why returning zeros for all samples in X as soon as one of them has `std==0`?

MechCoder: Sorry about the confusion. `X.shape[0] > 1` raises an error before this (gradient-support is only for single samples). So this line of code is reached when there is only one sample. (Hence mask is a 1-D array of length 1 always)

betatim: we should add a comment about single sample etc (maybe just copy your comment)

glouppe: oooh, okay

MechCoder:
Documented. I forgot to copy-paste from the other acquisition function.

Anything else?

glouppe:
Merging this, thanks!

Issue #242, opened 2016-09-29
Speed up tests


betatim:
https://github.com/scikit-optimize/scikit-optimize/pull/241#issuecomment-250444363 and previously we discussed that travis is far too slow.

Which tests are taking a lot of time? [These guys](https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/163683638) (scroll to the bottom and you get a top ten)

I would like to keep some tests that try and minimize an objective and make sure we get below some "threshold" to detect changes we make that break the minimizers. How few iterations can we use for that? Not sure 10 is enough, maybe 50 or 80? You'd need to run it for long enough to detect the difference between "it is just guessing" and "better than guessing".

Another option is that we add a new entry to the build matrix that runs the heavy tests and the normal tests skip them.

PR #276, opened 2016-11-30
[MRG] Remove tests on branin and hart6 

glouppe:
This PR removes tests on `branin` and `hart6`, that should rather be considered as benchmarks.

Fix for #242.


glouppe:
Goes down from 27 to 13 minutes.

betatim:
🚄

PR #278, opened 2016-11-30
[MRG] Remove pyqt dependency

glouppe:
Let's see if the issue raised in #230 is now fixed.


glouppe:
Travis is happy, merging.

PR #279, opened 2016-12-06
[MRG+1] Normalize integer dimensions

MechCoder:
We should normalize integer dimensions as well because the isotropic GP hyper-parameters might be highly dominated by these if they are on a completely different scale from the others such as from [1, 47]

### DIFF
--- a/skopt/space/transformers.py
+++ b/skopt/space/transformers.py
@@ -102,9 +102,10 @@ class Normalize(Transformer):
     * `high` [float]:
         Higher bound.
     """
-    def __init__(self, low, high):
+    def __init__(self, low, high, is_int=False):
         self.low = low
         self.high = high
+        self.is_int = is_int
 
     def transform(self, X):
         X = np.asarray(X)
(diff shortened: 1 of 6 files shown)


skopt/space/transformers.py
```diff
@@ -120,7 +121,10 @@ def inverse_transform(self, X):
             raise ValueError("All values should be less than 1.0")
         if np.any(X < 0.0):
             raise ValueError("All values should be greater than 0.0")
-        return X * (self.high - self.low) + self.low
+        X_norm = X * (self.high - self.low) + self.low
```
glouppe: Isnt that the unnormalized `X`?

glouppe:
Except this small comment, +1

skopt/space/transformers.py
```diff
@@ -120,7 +121,10 @@ def inverse_transform(self, X):
...
+        if self.is_int:
+            return np.round(X_norm)
```
betatim: Should we convert the dtype as well? `np.round` returns an array of floats :-/

MechCoder:
done

betatim:
🎉

Issue #274, opened 2016-11-30
Remove option `acq_optimizer="auto"`


glouppe:
In order to simplify  slightly the API, I would vote for removing the `"auto"` value of `acq_optimizer`. 

I am not convinced we should default to random sampling as soon as one of the dimensions is not `Real`. 

I would rather default to `"lbfgs"` and have, maybe, a warning if one the dimensions isnt `Real`. (Which by the way does not mean that the optimization will fail.)

MechCoder:
Sorry, but I would humbly disagree. It is all-right if it performs slow but no so-much if it provides a wrong output that cannot be relied on as the next candidate point. It is wrong because the output of the `fit_transform` for a `LabelBinarizer` is unreliable when the input is not in {0, 1}.

See: https://github.com/scikit-learn/scikit-learn/issues/7247

glouppe:
My point is that there is nothing wrong with running `lbfgs` on integer or binary dimensions, as long as the solution returned is cast back correctly.

MechCoder:
Ah yes, sorry you are right. I misread the documentation of the output of `inverse_transform` of `LabelBinarizer`

glouppe:
Fixed.

PR #281, opened 2016-12-29
[MRG] Remove acq_optimizer='auto' and set default to 'sampling'

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/274

See benchmarks:

1. https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-269493170
2. https://github.com/scikit-optimize/scikit-optimize/pull/262#issuecomment-260835534

### DIFF
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -13,7 +13,7 @@ from ..space import Space
 
 def gp_minimize(func, dimensions, base_estimator=None,
                 n_calls=100, n_random_starts=10,
-                acq_func="EI", acq_optimizer="auto", x0=None, y0=None,
+                acq_func="EI", acq_optimizer="sampling", x0=None, y0=None,
                 random_state=None, verbose=False, callback=None,
                 n_points=10000, n_restarts_optimizer=5, xi=0.01, kappa=1.96,
                 noise="gaussian", n_jobs=1):
@@ -85,7 +85,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
         - `"EI"` for negative expected improvement.
         - `"PI"` for negative probability of improvement.
 
-    * `acq_optimizer` [string, `"auto"`, `"sampling"` or `"lbfgs"`, default=`"auto"`]:
+    * `acq_optimizer` [string, `"sampling"` or `"lbfgs"`, default=`"sampling"`]:
         Method to minimize the acquistion function. The fit model
         is updated with the optimal value obtained by optimizing `acq_func`
         with `acq_optimizer`.
@@ -97,9 +97,6 @@ def gp_minimize(func, dimensions, base_estimator=None,
               - `"lbfgs"` is run for 20 iterations with these points as initial
                 points to find local minima.
               - The optimal of these local minima is used to update the prior.
-        - If set to `"auto"`, then it is set to `"lbfgs"`` if
-          all the search dimensions are Real (continuous). It defaults to
-          `"sampling"` for all other cases.
 
     * `x0` [list, list of lists or `None`]:
         Initial input points.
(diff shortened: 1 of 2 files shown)


glouppe:
What shall we do with this now?

MechCoder:
There is definitely some issues with how we handle Integer dimensions. I'll get back to this as soon as I have some solid reasoning behind that.

MechCoder:
(In the context of switching the default from "auto" to "lbfgs" according to the original issue)

MechCoder:
Running it on https://scikit-optimize.github.io/notebooks/hyperparameter-optimization.html there does not seem to be a huge difference. I think we can remove it.

Without K-best:

| Acquisition Function | Mean Minimum | Std |
| --- | --- | --- |
| lbfgs | 2.884 | 0.037 |
| sampling | 2.844   | 0.024 |

With K-best

| Acquisition Function | Mean Minimum | Std |
| --- | --- | --- |
| lbfgs | 2.925 | 0.032 |
| sampling | 2.917   | 0.035 |

glouppe:
What do you call "K-best"?

MechCoder:
I had meant with and without the preprocessing using SelectKBest

MechCoder:
Another observation, I have to increase `n_restarts_optimizer` further to give a non-constant predictions for estimating piecewise constant functions.

```python
import numpy as np

from skopt.learning.gaussian_process.kernels import ConstantKernel
from skopt.learning.gaussian_process.kernels import Matern

def f(x):
    return int(x)**2 - 1

x = np.linspace(0, 20, 61)[:60]
y = [f(xi) for xi in x]

for startptr, endptr in zip(range(0, 61, 3), range(3, 62, 3)):
    plt.plot(x[startptr: endptr], y[startptr: endptr], "black")

cov_amplitude = ConstantKernel(1.0, (0.01, 1000.0))
matern = Matern(length_scale=1.0, length_scale_bounds=[0.01, 100], nu=2.5)

gp = GaussianProcessRegressor(
    kernel=cov_amplitude * matern, normalize_y=True, random_state=0,
    alpha=0.0, n_restarts_optimizer=5)
gp.fit(np.reshape(x, (-1, 1)), y)

x_unseen = np.linspace(0, 20, 201)[:200]
y_pred = gp.predict(np.reshape(x_unseen, (-1, 1)))

for startptr, endptr in zip(range(0, 201, 20), range(20, 221, 20)):
    plt.plot(x_unseen[startptr: endptr], y_pred[startptr: endptr], "red")
plt.show()
```

With `n_restarts_optimizer=5`, the graph looks like



With `n_restarts_optimizer=2`, the graph looks like

glouppe:
Interesting! This is also what we observe in the BO notebook, everything is flat except some spikes, just like your second plot. Maybe there is something wrong in the training procedure of the GP that we should mention upstream in scikit-learn? In the meantime, shall we increase `n_restarts_optimizer` further?

skopt/optimizer/gp.py
```diff
@@ -85,7 +85,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
         - `"EI"` for negative expected improvement.
         - `"PI"` for negative probability of improvement.
 
-    * `acq_optimizer` [string, `"auto"`, `"sampling"` or `"lbfgs"`, default=`"auto"`]:
+    * `acq_optimizer` [string, `"sampling"` or `"lbfgs"`, default=`"sampling"`]:
```
glouppe: Dont you think we should set instead LBFGS as default?

MechCoder:
> Maybe there is something wrong in the training procedure of the GP that we should mention upstream in scikit-learn? 

When I had a look sometime before, the implementation seemed to be correct. I just think it gets stuck in a local minima. I can have a closer look though.

MechCoder:
Btw, I closed the pull request because I messed up some git stuff

PR #284, opened 2017-01-03
[MRG] Add new acquisition function that chooses the start point of lbfgs wisely

MechCoder:


MechCoder:
New benchmarks:

On branin: 50 calls and averaged across 10 runs


| Acquisition Function | Mean Minimum | Std |
| --- | --- | --- |
| this branch | 0.3992 |  0.0011 |
| lbfgs | 0.4104 | 0.0146 |
| sampling | 0.3995   | 0.0014 |

MechCoder:
On hart6:

| Acquisition Function | Mean Minimum | Std |
| --- | --- | --- |
| this branch | -3.2273 |  0.0617 |
| lbfgs | -2.9502 | 0.3273 |
| sampling | -3.1416   | 0.0646 |

skopt/optimizer/base.py
```diff
@@ -277,18 +278,25 @@ def base_minimize(func, dimensions, base_estimator,
 
         models.append(gp)
 
-        if acq_optimizer == "sampling":
+        if acq_optimizer == "sampling" or acq_optimizer == "lbfgs_optim":
```
MechCoder: It might be better to implement this in "lbfgs" itself rather than create a separate acquisition function.

glouppe: +1, let's simply things by having only two options for `acq_optimizer`.

glouppe:
Shall we harmonize `n_points` and `n_restarts_optimizer` to have the same default values for all our optimizers?

glouppe:
Other than that, results look good in my opinion! This would be a nice compromise for closing #262. What do you think? (the added-value of spray points being left out for further study)

MechCoder:
Good to go from my side. Let us begin the new year on a good note :P

glouppe:
Travis is happy! Merging :)

PR #288, opened 2017-01-06
[MRG] Fix pep8 compliance

betatim:
Clean up some pep8 warnings and additional tests.

Added an additional test and logic in `check_dimension` to get coverage
for raising the exception.

### DIFF
--- a/skopt/learning/gaussian_process/kernels.py
+++ b/skopt/learning/gaussian_process/kernels.py
@@ -287,12 +287,13 @@ class Sum(Kernel, sk_Sum):
             self.k2.gradient_x(x, X_train)
         )
 
+
 class Product(Kernel, sk_Product):
 
     def gradient_x(self, x, X_train):
         x = np.asarray(x)
+        x = np.expand_dims(x, axis=0)
         X_train = np.asarray(X_train)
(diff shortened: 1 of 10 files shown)


glouppe:
:+1:

skopt/learning/gaussian_process/kernels.py
```diff
@@ -287,12 +287,13 @@ def gradient_x(self, x, X_train):
...
         )
 
+
 class Product(Kernel, sk_Product):
 
     def gradient_x(self, x, X_train):
         x = np.asarray(x)
+        x = np.expand_dims(x, axis=0)
         X_train = np.asarray(X_train)
-        X = np.expand_dims(x, axis=0)
```
betatim: Was this a typo? Capital X instead of x? I assumed that it was and fixed it accordingly.

glouppe: indeed, seems like a good catch

betatim: is there a test we can add that would have detected this? To stop the :beetle: :ghost: from coming back

glouppe:
Merging this, nothing controversial to be discussed.

PR #289, opened 2017-01-08
Enable python 3.6 on travis

betatim:
Let's see how we do on the next version of python


betatim:
The conda package for matplotlib doesn't seem to work with python3.6 :-/

glouppe:
Might be a little early for that. Is 3.6 already fully supported by Anaconda?

betatim:
I have a local environment with python3.6, scipy, numpy and scikit-learn but as on travis you can't install matplotlib (matplotlib is in conflict with python3.6). Not researched why. We can either close this or keep it open and every so often kick travis to see if it works now.

glouppe:
https://www.scivision.co/switch-anaconda-python-36/

A workaround seems to install matplotlib through pip, until conda install works again.

PR #291, opened 2017-01-09
[MRG] Issue warning when points are revaluated

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/pull/163
Supersedes https://github.com/scikit-optimize/scikit-optimize/pull/170


skopt/optimizer/base.py
```diff
@@ -303,6 +312,23 @@ def base_minimize(func, dimensions, base_estimator,
...
+            next_x_arr = np.array(next_x)
+            for x in Xi:
+                x_arr = np.array(x)
+                cat_eq = np.all(x_arr[cat_inds] == next_x_arr[cat_inds])
+                non_cat_eq = np.allclose(
+                    np.array(x_arr[non_cat_inds], dtype=np.float32),
+                    np.array(next_x_arr[non_cat_inds], dtype=np.float32))
+                if cat_eq and non_cat_eq:
+                    warnings.warn(
+                        "candidate point at the end of %d model fit, already chosen" % i)
```
glouppe: Hmm, so this does produce warnings, but provides no workaround for the problem, right?  So, while it is still beter than master, I am sure this is yet a solution for #163. What do you think?

glouppe: Okay, never mind, just read the discussion in #170 again, where we assumed that by default the function should be considered stochastic, hence reevaluating it might be worth something.

glouppe:
Looks good to me.

(Might be nice to have #234 merged to avoid constantly rebasing it upon new changes such as this one.)

betatim:
Rebase and merge.

(Tried out the "resolve conflicts" button on the github webpage but not sure I really know what I am doing. Looks useful though)

MechCoder:
Not sure what happened? I rebased and pushed and it closed automatically! ?

glouppe:
All commits are gone :o

betatim:
:ghost: ? :-/

PR #297, opened 2017-01-10
[MRG] Issue warning when points are reevaluated: Take 3

MechCoder:


skopt/optimizer/optimizer.py
```diff
@@ -260,6 +268,24 @@ def tell(self, x, y, fit=True):
             self._next_x = self.space.inverse_transform(
                 next_x.reshape((1, -1)))[0]
 
+            cat_inds = self._cat_inds
+            non_cat_inds = self._non_cat_inds
+            if len(cat_inds) == 0:
+                if np.any(np.apply_along_axis(lambda x: np.allclose(x, next_x), 1, self.Xi)):
+                    warnings.warn("candidate point already chosen")
```
betatim: -> "Objective has previously been evaluated at this point"?

betatim:
Should we move the check to `ask`? As a user you might think that it is the point you just passed to `tell` that it is complaining about.

MechCoder:
Done, thanks!

glouppe:
> As a user you might think that it is the point you just passed to tell that it is complaining about.

Isnt that precisely what we should check? I.e. warn the user when they `tell` the same point twice or more.

MechCoder:
Isn't warning while asking a better option? You have the flexibility to change the point (or) the objective value after it warns while at `tell` time it will fit a new model anyway.

glouppe:
I honestly dont understand how you can warn the user about a point that was not `told` yet. (The user has the freedom the tell a point `x` that is different from the one returned by `ask`.)

MechCoder:
This fix deals with the original issue of suggesting the same point multiple times in the `*_opt` functions, where there is no option of "telling" a point in between and not about warning the use if he provides the same point multiple times (which may be a separate issue).

We can add a warning at both `ask` and `tell` time.

MechCoder:
I was about to add a warning at both `ask` and `tell` time but the told point already has the function evaluated at that point. So why even warn?

betatim:
I was thinking that we'd want to inform the user that we are suggesting to them the point we think is optimal, but that we have suggested it to them before. Hence in `ask`. In some sense as the optimizer I don't care if you repeatedly tell me the same point again and again. (Maybe we should make it a mistake to add the same point with the same objective value repeatedly?)

glouppe:
Okay makes sense.

glouppe:
(We might want to improve the warning message then)

MechCoder:
I've reworded it. Let me know if you want anything more specific.

betatim:
How about: "The objective has been evaluated at this point before." ? Not sure it is better, but I didn't like the 'told' so I tried to come up with something different :-/ Otherwise merge?!

MechCoder:
That was the initial warning message :P

betatim:
:man_shrugging: let's let @glouppe decide :)

glouppe:
I side with Tim on English issues :)

betatim:
:clipboard: :pen: :white_check_mark:

Issue #294, opened 2017-01-09
Group acquisition function and optimizer arguments


betatim:
The number of arguments that need passing around is increasing, sooner or later we will start having collisions, and it is getting unwieldy.

[Suggestion](https://github.com/scikit-optimize/scikit-optimize/pull/234/files#r95050797) is to replace them by `acq_func_kwargs` and `acq_optimizer_kwargs` instead of (`n_points`, `n_restarts_optimizer`, `xi`, `kappa`).

Maybe start this before #115

PR #298, opened 2017-01-10
[MRG] Group acquisition function and acquisition optimizer arguments

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/294

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -4,20 +4,20 @@ import warnings
 from scipy.stats import norm
 
 
-def gaussian_acquisition_1D(X, model, y_opt=None, acq_func="LCB", xi=0.01,
-                            kappa=1.96):
+def gaussian_acquisition_1D(X, model, y_opt=None, acq_func="LCB",
+                            acq_func_kwargs=None):
     """
     A wrapper around the acquisition function that is called by fmin_l_bfgs_b.
 
     This is because lbfgs allows only 1-D input.
     """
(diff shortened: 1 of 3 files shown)


skopt/acquisition.py
```diff
@@ -4,20 +4,20 @@
...
+def gaussian_acquisition_1D(X, model, y_opt=None, acq_func="LCB",
+                            acq_func_kwargs=None):
     """
     A wrapper around the acquisition function that is called by fmin_l_bfgs_b.
 
     This is because lbfgs allows only 1-D input.
     """
     return _gaussian_acquisition(np.expand_dims(X, axis=0),
-                                 model, y_opt, acq_func, xi, kappa,
+                                 model, y_opt, acq_func_kwargs=acq_func_kwargs,
```
betatim: It would be nice to keep the explicit arguments here and call the acq function with `func(**kwargs)`. Was thinking we could then send users here from the docs of `Optimizer` with "Additional kwargs, see your acq func's documentation for what is available"

MechCoder: Sorry, not clear. This function is private no, so why does it matter how I call it over here?

We can still say look at the acquisition functions such as `gaussian_lcb`, `gaussian_ei` ec, to see what is available.

betatim: You are right! More :coffee: for me next time :-/

MechCoder:
@glouppe Merge?

glouppe:
Looks good, merging.

PR #304, opened 2017-01-25
[MRG] Implement entropy search acquisition function

MechCoder:


MechCoder:
@jmetzen Would you be able to have a look for sanity? Thanks!

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        rep_curr = rep_cands[startptr: startptr+n_trial_points]
+        argmin = np.argmin(quad_func(rep_curr, model))
+        rep_points[i] = rep_curr[argmin]
+
+    # # Step 2: Calculate p_min for each of these represnter points using
+    # # Monte-Carlo approximation.
+    # sampled_f = model.sample_y(
+    #     X_candidates, n_samples=n_samples, random_state=rng)
+    # p_min = (np.bincount(np.argmax(sampled_f, axis=0)) + 1) / (n_candidates + float(n_samples))
+    # base_entropy = entropy(p_min)
```
MechCoder: The computation of this seems unnnecessary since the base entropy is the same for all the candidate points.

glouppe:
You rock! I'll try to review tomorrow

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+    for cand_ind in range(n_rep_points, n_rep_points + X.shape[0]):
+
+        # delta(cov(X_rep)) = \Sigma(X_rep, X_cand) \Sigma^-1(X_cand, X_cand) \Sigma(X_cand, X_rep)
+        # Note that \Sigma(X_cand, X_cand) is a scalar since we only require the next
+        # best evaluation.
+        cov_rep_X = X_all_cov[cand_ind][: n_rep_points]
+        cov_rep_X_row = np.reshape(cov_rep_X, (-1, 1))
+        cov_rep_X_col = np.reshape(cov_rep_X, (1, -1))
+        cov_cand_inv = 1.0 / X_all_cov[cand_ind, cand_ind]
+        cov_delta = cov_cand_inv * np.dot(cov_rep_X_row, cov_rep_X_col)
```
MechCoder: @jmetzen In your implementation here (https://github.com/jmetzen/bayesian_optimization/blob/master/bayesian_optimization/acquisition_functions.py#L311) there is a negative sign. Could you let me know where this comes from?

MechCoder:
Thanks! A verification of the code would be great, before benchmarks etc

jmetzen:
Thanks for adding this. I will try to check the code on the weekend!

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
         return values, grad
 
     return values
+
+
+def gaussian_entropy_search(X, model, space, n_rep_points=50,
+                            n_trial_points=200, measure="EI", n_samples=500,
+                            random_state=None):
+    """
+    Uses entropy search to optimize the acquisition function.
```
betatim: Uses -> Use

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+
+    1. It approximates the search space by a set of representer points sampled from
+       a suitable measure such as the "probability of improvement" or
+       "expected improvement".
+
+    2. p_min at a representer point X_i is calculated by the Integral
+       p(f(x)) \prod_{j=1}^N_rep \Theta(X_j - X_i) \\
```
betatim: opinions on using unicode characters instead of "latex in your head" for the greek symbols?

betatim:
Needs tests :)

skopt/acquisition.py
MechCoder: sure, but might be better to all at once.

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
         return values, grad
 
     return values
+
+
+def gaussian_entropy_search(X, model, space, n_rep_points=50,
```
glouppe: Just for my own knowledge, what is specifically "gaussian" about entropy search?

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        The fit estimator that approximates the function through the
+        method ``predict``.
+        It should have a ``return_std`` parameter that returns the standard
+        deviation.
+
+    * `space` [Space instance]:
+        This is required to sample representer points.
+
+    * `n_rep_points` [int, default 50]:
+        Number of represter points.
```
glouppe: representer

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+
+    * `n_trial_points` [int, default 200]:
+        Number of trial points sampled uniformly for each representer point
+        candidate.
```
glouppe: In your formula above, what is a trial point? the X_j?

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+
+    * `n_samples` [int, default 500]:
+        Number of samples to sample from the GP posterior to approximate p_min
```
glouppe: In your formula above, what is a sample? the X_j?

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+
+    * `random_state` [int, RandomState instance, or None (default)]:
+        For reproducible results.
+    """
+    rng = np.random.RandomState(random_state)
+    n_features = X.shape[1]
+
+    # Step 1: Sample `n_representer_points` from the given nonmeasure.
```
glouppe: measure

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+    # XXX: Use MCMC sampling to sample from the quadrature measure.
+    n_total = n_rep_points*n_trial_points
+    rep_cands = space.transform(space.rvs(n_samples=n_total, random_state=rng))
+    rep_points = np.zeros((n_rep_points, n_features))
+
+    if measure == "LCB":
+        quad_func = gaussian_lcb
```
glouppe: why the name `quad_func`?

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+
+    elif measure in ["EI", "PI"]:
+        if measure == "EI":
+            quad_func = lambda X, model: -gaussian_ei(X, model)
+        else:
+            quad_func = lambda X, model: -gaussian_pi(X, model)
+
+    for i, startptr in enumerate(range(0, X.shape[0], n_total)):
```
glouppe: do you need the loop? you can evaluate `quad_func` at all `rep_cands`, then reshape by `(n_rep_points, n_trial_points)` and take the argmin by specifying the axis.

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        rep_curr = rep_cands[startptr: startptr+n_trial_points]
+        argmin = np.argmin(quad_func(rep_curr, model))
+        rep_points[i] = rep_curr[argmin]
```
glouppe: Also, am I missing something or `rep_points[i]` will eventually have the same values (as `n_trial_points` gets larger), no matter the considered representer point `X_i`? If yes, couldnt we simply minimize `quad_func` with lbfgs?

skopt/acquisition.py
glouppe: also not sure to understand the upper bound on `X.shape[0]` and the `n_total` step?

skopt/acquisition.py
glouppe: Okay, I think this is wrong. The optimization should be wrt a function sampled from the GP posterior, shouldnit?

See https://github.com/jmetzen/bayesian_optimization/blob/master/bayesian_optimization/acquisition_functions.py#L368:L369

CC @jmetzen

skopt/acquisition.py
jmetzen: It doesn't affect the maximizer of the information gain since - as you say - it is the same for all candidate points. However, if you want to interpret the value of the acquisition function as information gain, you need this "base" entropy. It is also relatively cheap as it has effectively only the cost of evaluating one additional query point. But you don't need it strictly. But then the returned quantity should not be named "inf_gain".

skopt/acquisition.py
```diff
@@ -264,3 +267,119 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        Number of samples to sample from the GP posterior to approximate p_min
+
+    * `random_state` [int, RandomState instance, or None (default)]:
+        For reproducible results.
+    """
+    rng = np.random.RandomState(random_state)
+    n_features = X.shape[1]
+
+    # Step 1: Sample `n_representer_points` from the given nonmeasure.
+    # XXX: Use MCMC sampling to sample from the quadrature measure.
```
jmetzen: If I understand it correctly, you resample the representer points for every call of `gaussian_entropy_search`. And `fmin_l_bfgs_b` calls it several times during the search for the next query point. This causes two issues:
 1) There is unnecessary computation   
 2) Since the representer points are different in each call, the values returned by the acquisition function are not compatible (e.g., evaluating the same point twice might return different values). This issue could be fixed by setting the same random seed before generating the representer points.

skopt/acquisition.py
jmetzen: This is a good question. According to the paper, it should be a positive sign. I believe I checked it and came to the conclusion that it should be a negative sign and this was missing in the paper, but I cannot say for sure. I should have added a comment, sorry. It seems that RoBO has also a negative sign (https://github.com/automl/RoBO/blob/master/robo/acquisition_functions/information_gain.py#L263). But please double check (if you factor this part out, you could write a unit test for it).

MechCoder: Yes, you are right. There should be a negative sign. Do you know the approach to derive it analytically? Here is a numeric approach.

```python
gpr = GaussianProcessRegressor(n_restarts_optimizer=5)
X = [[1.0], [2.0], [3.0], [4.0], [5.0]]
y = [1.0, 4.0, 9.0, 16.0, 25.0]
gpr.fit(X, y)

eval_X = np.linspace(1.0, 5.0, 30.0).reshape(-1, 1)
_, init_cov = gpr.predict(eval_X, return_cov=True)

new_X = [[1.5], [2.5], [3.5]]
new_y = y + [t**2 for x in new_X for t in x]

all_X = np.vstack((eval_X, new_X))
_, all_cov = gpr.predict(all_X, return_cov=True)
cross_cov = all_cov[-3:, :-3]
delta_cov = np.dot(cross_cov.T, np.dot(np.linalg.inv(all_cov[-3:, -3:]), cross_cov))

X_after = np.vstack((X, new_X))
gpr.fit(X_after, new_y)
_, new_cov = gpr.predict(eval_X, return_cov=True)
act_delta = new_cov - init_cov

act_delta[-2]
array([ 0.        ,  0.00118236,  0.00206068,  0.00254834,  0.00259283,
        0.00218859,  0.00138669,  0.00029841, -0.00091064, -0.00203781,
       -0.00287196, -0.00323024, -0.00299673, -0.00215471, -0.00080388,
        0.00084434,  0.00249582,  0.00382406,  0.00453331,  0.00442178,
        0.00343106,  0.00167007, -0.00059326, -0.00297201, -0.00502493,
       -0.00633262, -0.00657042, -0.0055635 , -0.00331457,  0.        ])
delta_cov[-2]
array([  0.00000000e+00,  -1.18235670e-03,  -2.06067792e-03,
        -2.54833878e-03,  -2.59283482e-03,  -2.18858735e-03,
        -1.38668602e-03,  -2.98406054e-04,   9.10643899e-04,
         2.03780813e-03,   2.87196229e-03,   3.23024283e-03,
         2.99672751e-03,   2.15471226e-03,   8.03883150e-04,
        -8.44338624e-04,  -2.49581620e-03,  -3.82405586e-03,
        -4.53330821e-03,  -4.42177769e-03,  -3.43106330e-03,
        -1.67006787e-03,   5.93257256e-04,   2.97201330e-03,
         5.02492645e-03,   6.33262299e-03,   6.57041884e-03,
         5.56350449e-03,   3.31456744e-03,   6.71512219e-17])

```

skopt/acquisition.py
MechCoder: We use a Gaussian Process posterior to approximate `p_min` at each of the representer points, i.e this implementation works only if the posterior is multi-variate gaussian. Sec 2.3 in the entropy search paper (http://jmlr.csail.mit.edu/papers/volume13/hennig12a/hennig12a.pdf)

MechCoder: (which means this will work only for the surrogate model GP case)

skopt/acquisition.py
MechCoder: yes ofc, sorry.

skopt/acquisition.py
MechCoder: Good point, sorry that I missed it.

skopt/acquisition.py
MechCoder: This is a blunder. It should have been `(0, n_total, n_trial_points)`

MechCoder:
In case the diff gets collapsed in the future.

```python
gpr = GaussianProcessRegressor(n_restarts_optimizer=5)
X = [[1.0], [2.0], [3.0], [4.0], [5.0]]
y = [1.0, 4.0, 9.0, 16.0, 25.0]
gpr.fit(X, y)

eval_X = np.linspace(1.0, 5.0, 30.0).reshape(-1, 1)
_, init_cov = gpr.predict(eval_X, return_cov=True)

new_X = [[1.5], [2.5], [3.5]]
new_y = y + [t**2 for x in new_X for t in x]

all_X = np.vstack((eval_X, new_X))
_, all_cov = gpr.predict(all_X, return_cov=True)
cross_cov = all_cov[-3:, :-3]
delta_cov = np.dot(cross_cov.T, np.dot(np.linalg.inv(all_cov[-3:, -3:]), cross_cov))

X_after = np.vstack((X, new_X))
gpr.fit(X_after, new_y)
_, new_cov = gpr.predict(eval_X, return_cov=True)
act_delta = new_cov - init_cov

act_delta[-2]
array([ 0.        ,  0.00118236,  0.00206068,  0.00254834,  0.00259283,
        0.00218859,  0.00138669,  0.00029841, -0.00091064, -0.00203781,
       -0.00287196, -0.00323024, -0.00299673, -0.00215471, -0.00080388,
        0.00084434,  0.00249582,  0.00382406,  0.00453331,  0.00442178,
        0.00343106,  0.00167007, -0.00059326, -0.00297201, -0.00502493,
       -0.00633262, -0.00657042, -0.0055635 , -0.00331457,  0.        ])
delta_cov[-2]
array([  0.00000000e+00,  -1.18235670e-03,  -2.06067792e-03,
        -2.54833878e-03,  -2.59283482e-03,  -2.18858735e-03,
        -1.38668602e-03,  -2.98406054e-04,   9.10643899e-04,
         2.03780813e-03,   2.87196229e-03,   3.23024283e-03,
         2.99672751e-03,   2.15471226e-03,   8.03883150e-04,
        -8.44338624e-04,  -2.49581620e-03,  -3.82405586e-03,
        -4.53330821e-03,  -4.42177769e-03,  -3.43106330e-03,
        -1.67006787e-03,   5.93257256e-04,   2.97201330e-03,
         5.02492645e-03,   6.33262299e-03,   6.57041884e-03,
         5.56350449e-03,   3.31456744e-03,   6.71512219e-17])
```

skopt/acquisition.py
MechCoder: (not any more)

MechCoder:
> Okay, I think this is wrong. The optimization should be wrt a function sampled from the GP posterior, shouldnit?

No, it is not wrong. Choosing the representer points can be done using any function. Look at the last paragraph in section 2.2 of http://jmlr.csail.mit.edu/papers/volume13/hennig12a/hennig12a.pdf

Sampling from a uniform grid and choosing the point that gives the best acquisition function value is a naive alternative to using advanced techniques to sample directly from the acquisition function.

jmetzen:
>> Okay,
>>I think this is wrong. The optimization should be wrt a function sampled from the GP posterior, shouldnit?

> No, it is not wrong. Choosing the representer points can be done using any function. Look at the last paragraph in section 2.2 of http://jmlr.csail.mit.edu/papers/volume13/hennig12a/hennig12a.pdf

> Sampling from a uniform grid and choosing the point that gives the best acquisition function value is a naive alternative to using advanced techniques to sample directly from the acquisition function.

I used effectively Thompson sampling as method for sampling representer points instead of EI and PI. In principle, using any of these is legitimate. One disadvantage of EI and PI is that the only place where stochasticity enters the selection of the representer points is in selecting the trial-points. In contrast, Thompson sampling also adds stochasticity in selecting the representer points from the trial points. This can lead to a more diverse set of representer points (just think what happens if n_trial_points becomes very large - you would select effectively always the same representer point, namely the maximizer of EI/PI).  Thus, I would propose to add Thompson sampling as an alternative to EI/PI here. What do you think?

glouppe:
>> Okay, I think this is wrong. The optimization should be wrt a function sampled from the GP posterior, > shouldnit?
> No, it is not wrong. Choosing the representer points can be done using any function. Look at the last paragraph in section 2.2 of http://jmlr.csail.mit.edu/papers/volume13/hennig12a/hennig12a.pdf

> Sampling from a uniform grid and choosing the point that gives the best acquisition function value is a naive alternative to using advanced techniques to sample directly from the acquisition function.

What is still not clear to me is that the representer will eventually _all_ converge towards the same value as `n_trial_points` tends to infinity.

Maybe I am wrong, but I thought the representer points were instead draw from the distribution of minima, with respect to the functions sampled from the GP posterior. There the distribution would not collapse to a same representer value. 

What am I missing?

skopt/acquisition.py
```diff
@@ -264,3 +266,109 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+    elif measure in ["EI", "PI"]:
+        if measure == "EI":
+            quad_measure = lambda X : -gaussian_ei(X, model)
+        else:
+            quad_measure = lambda X : -gaussian_pi(X, model)
+
+    n_total = n_rep_points * n_trial_points
+    all_values = np.reshape(quad_measure(rep_cands), (n_rep_points, n_trial_points))
+    rep_inds = np.arange(0, n_total, n_trial_points) + np.argmin(all_values, axis=1)
+    rep_points = rep_cands[rep_inds]
```
glouppe: How does that work when `rep_cands` is `None` by default?

MechCoder:
I agree that the current approach is deterministic and incorrect and the right approach would be to sample from the acquisition function directly. But it it is not necessary that the representer points should be sampled from the minima, it can be sampled from any function that has a strong effect on the entropy change such as EI. In fact it is better to use EI since the representer points sampled from EI take care of the exploitation-exploration tradeoff. I'll change it to follow @jmetzen 's approach and mark MCMC sampling from such functions as TODO for later.

However, I have a question with respect to Thompson sampling. Am I right in thinking that for picking each candidate the game is played from scratch. (i.e the prior updating does not happen?)

skopt/acquisition.py
MechCoder: It doesn't. Should I make it a positional argument?

MechCoder: It should be set outside the loop, since this should be constant for every acquisition function call.

skopt/acquisition.py
```diff
@@ -264,3 +266,86 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        points.
+
+    * `random_state` [int, RandomState instance, or None (default)]:
+        For reproducible results.
+    """
+    rng = np.random.RandomState(random_state)
+    n_rep_points = len(rep_points)
+    n_features = X.shape[1]
+    X_all = np.vstack((rep_points, X))
+    X_all_mean, X_all_cov = model.predict(X_all, return_cov=True)
```
MechCoder: The prediction of the mean and covariance of the representer points can be moved outside the function `gaussian_entropy_search` for lbfgs but I prefer to do that later.

MechCoder:
I have implemented Thompson sampling as a method to give a distribution of the representer points or the minima of the black-box function. I am running benchmarks to check the state as of now.

Issue #292, opened 2017-01-09
Add tests for mixed dimensions


MechCoder:

betatim:
What exactly did you have in mind for mixed dimensions? A space made of categorical and real?

MechCoder:
Yes. We don't test those no?

glouppe:
Fixed in #311

PR #311, opened 2017-02-03
[MRG] Fix handling of mixed spaces in gp_minimize

MechCoder:


MechCoder:
Also fixes https://github.com/scikit-optimize/scikit-optimize/issues/292

PR #317, opened 2017-02-12
[MRG] Supervised Learning benchmark

iaroslav-ai:
Firstly, thanks for the great repo!

For quite some time I had a draft of benchmark for the black box optimization algorithms, applied to a task of optimizing parameters of ML algorithms for the task of supervised learning. The code of this benchmark is in this pull request, adapted to use your optimizers. 

The code requires numpy, pandas, scipy, sklearn, keras, tensorflow or theano, xlrd and of course skopt packages to be installed. You can run the benchmark by running “bench_ml.py”.

Currently the code implements benchmark on 10 datasets where parameters for 6 classes of supervised models are tuned to optimize performance on datasets. Regression learning task is solved on 5 datasets, and classification on the rest of datasets. 3 model classes are regression models, and rest are classification models.

Among models used, the most notable ones are deep neural network models for classification or regression (see _sk_keras.py). These models both have 10 parameters that need tuning, (2 categorical, 1 integer, rest real values), and thus are hard to enumerate a la brute force. Other models are SVM and decision trees, which have <= 3 parameters. 

The code collects all the data about experiments that optimizers perform, and can save it as a json. This data can be used in 2 ways:
* For algorithm comparison, eg for optimization of decision tree model parameters: 
figure_1

* In order to reduce the benchmark runtime. Let f: P -> R denote expensive black box objective. The experiments data contains pairs of (p, f(p)) \in P x R. One could use this data to train SVR to estimate f(p) given p, and use it instead of f. SVR however cannot model noise in f(p), so one could use VAE or GAN generative models instead.

The code can run experiments in parallel, but this does not work well with either theano or tensorflow. Running in parallel would be beneficial, as for small amount of data both cannot make use of parallelism efficiently as it seems. Maybe something like chainer would be better for this.

It is easy to extend the code with new datasets, especially if the data is available somewhere online as a csv, xsl, arff or space delimited table - then it takes ~5 lines of code. Similarly, if the models follow the sklearn standards, then it is sufficient to specify python class implementing the model and parameters that will be tuned. 

The code is in working state, however some improvements are necessary (eg add command line arguments). Let me know what you think about all this.


benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+        :param optim_type: Type of optimizer to use, 'rmsprop' or 'adam'
+        :param opt_par_1: Learning rate
+        :param opt_par_2: Rho for rmsprop, Beta1 for adam
+        :param opt_par_3: Beta2 for adam
+        :param batch_size: Size of data subset that is used to update the model
+        :param dropout: dropout regularization value
+        :param l2_regularization: regularization imposed on weights of neural network
+
+        """
+        self.config = n_neurons, n_layers, neuron_type, opt_par_1, opt_par_2, opt_par_3, optim_type, batch_size, dropout, l2_regularization
```
MechCoder: If this is a true sklearn-like interface, this should be

```python
self.n_neurons = n_neurons
self.n_layers = n_layers
```
etc

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+        if self.classification:
+            self.model.compile(optimizer=opt, loss='sparse_categorical_crossentropy')
+        else:
+            self.model.compile(optimizer=opt, loss='mse')
+
+        # split data into training and validation to use for stopping criterion
+        Tr = int(len(X) * 0.7) # use 70% for training, rest for valiation
+
+        X, Xv = X[:Tr], X[Tr:]
+        Y, Yv = Y[:Tr], Y[Tr:]
```
MechCoder: This should be done outside of fit.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+
+        max_patience = 16
+        patience = max_patience
+        best_perf = -np.inf
+        best_weights = None
+
+        for epoch in range(64): # limit the number of epochs
```
MechCoder: Ideally `num_epochs` should be another hyperparameter to tune and the hyper-parameter optimizer should be able to figure out when to stop as per the target score.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+                 classification=False,
+                 n_neurons=128,
+                 n_layers=2,
+                 neuron_type='relu',
+                 opt_par_1=0.001,
+                 opt_par_2=0.9,
+                 opt_par_3=0.99,
+                 optim_type='adam',
+                 batch_size=128,
+                 dropout=0.0,
```
MechCoder: It seems like you do not use dropout.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+            Y = Y.astype('int32')
+        else:
+            Y = Y.astype('float32')
+
+        n_neurons, n_layers, neuron_type, opt_par_1, opt_par_2, opt_par_3, optim_type, batch_size, dropout, l2_regularization = self.config
+        n_features = X.shape[1]
+
+        ip = Input(shape=(n_features,), dtype='float32')
+
+        x = ip
```
MechCoder: You can remove this.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+            #x = Dropout(p=dropout)(x)
+
+        if self.classification:
+            y = Dense(len(np.unique(Y)), activation='softmax', W_regularizer=l2(l2_regularization))(x)
+        else:
+            y = Dense(1, W_regularizer=l2(l2_regularization))(x)
+
+        self.model = Model(input=ip, output=y)
+
+        opt = None # variable for optimizer
```
MechCoder: You can remove this,

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+class _KerasModel():
+    def __init__(self,
+                 classification=False,
+                 n_neurons=128,
+                 n_layers=2,
+                 neuron_type='relu',
+                 opt_par_1=0.001,
+                 opt_par_2=0.9,
+                 opt_par_3=0.99,
+                 optim_type='adam',
```
MechCoder: `optim_type` -> `optimizer`

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+        :param opt_par_3: Beta2 for adam
+        :param batch_size: Size of data subset that is used to update the model
+        :param dropout: dropout regularization value
+        :param l2_regularization: regularization imposed on weights of neural network
+
+        """
+        self.config = n_neurons, n_layers, neuron_type, opt_par_1, opt_par_2, opt_par_3, optim_type, batch_size, dropout, l2_regularization
+        self.classification = classification # remember if the model should do classification or regression
+
+    def fit(self, X, Y):
```
MechCoder: `Y` should be lower-case.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
+import numpy as np
+import keras
+from keras.models import Model
+from keras.layers import Input, Dense, Dropout
+from keras.optimizers import Adam, RMSprop
+from sklearn.metrics import r2_score
+from keras.regularizers import l2
+
+
+class _KerasModel():
```
MechCoder: I would remove a generic name such as `_KerasModel` and rename it to `FullyConnectedNetwork` or something similar.

benchmarks/_sk_keras.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        self.config = n_neurons, n_layers, neuron_type, opt_par_1, opt_par_2, opt_par_3, optim_type, batch_size, dropout, l2_regularization
+        self.classification = classification # remember if the model should do classification or regression
+
+    def fit(self, X, Y):
+        """
+        Fits regressor nn to input data.
+
+        :param X: A matrix with every row being example input
+        :param Y: A vector with every entry being example output
```
MechCoder: `Y` is a matrix when it is a classifier.

MechCoder:
Thanks! Just had a very quick look. So, how should be approach this? My idea was to have a set of benchmark problems and if a new proposal performs well in a few of these, we can integrate it with the codebase. For that reason, it might be best to cache the benchmarks here (https://github.com/scikit-optimize/scikit-optimize/blob/master/benchmarks/README.md) for a set of problems.

Are we still on the same page?

betatim:
First up: thanks! Realistic benchmarks is exactly what we need to help tune the default parameters of the optimizers!

Some formalities:
* how do we tell users that the benchmarks have much heavier dependencies than `skopt` itself?
* with non trivial benchmarks, how do we run them automatically to check they are still working?

---

Maintaining an overview like we have in the README right now is a good idea. I like it. We should also try and have as much as possible a similar interface to run each benchmark and have it print out results so that they can be copy and pasted into the README.

If individual benchmarks also produce other output (plots, more statistics, etc) that is fine, as long as we can agree on a minimum of information in a common format :smile:

iaroslav-ai:
Thanks everyone for you informative responses! :) 

@MechCoder: I will make another pass over the code and fix where you commented, and beyond, so that it conforms better to the standards of the repo. I will also add the code so that the benchmark outputs metrics same to the ones found in benchmark README.MD so that  they can be cached.

@betatim: For the dependencies, I propose the following: have the necessary dependencies imported in try … except block somewhere in the beginning of benchmark script, and on ImportError display to user to run ‘pip install -r bench_ml.txt’ , where bench_ml.txt contains list of necessary dependencies.
Wrt your second point, do you mean if the benchmark itself is working properly? This is indeed something to look for, I will I add a few tests for that.

betatim:
I'm wondering how many of the benchmarks can we keep if we switch to `MLPClassifier` in `sklearn` and restrict ourselves to datasets that `fetch_ml()` knows how to obtain?

The rational is that adding several dependencies and lots of custom code (for example wrapping keras classifiers to look like sklearn ones) that isn't directly related to the benchmark (it is all infrastructure code) adds a lot of maintenance burden. Because we are such a small team I think it makes sense to use as much code as possible that "someone else" has to maintain. It frees us up to spend time on the bits that are unique to `skopt`.

So I would switch the classifiers to use one from `sklearn` ([`MLPClassifier`](http://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html) for example) and use datasets we can easily obtain through [`sklearn.datasets.fetch_mldata`](http://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_mldata.html)

Also, this PR is pretty big (~800lines) which means it will take a lot of time to get reviewed properly, if we can use more code from `sklearn` that should speed up the review.

What do you think?

iaroslav-ai:
@betatim I think this is a good idea - the less code needs to be maintained the better. I will do as you suggest, and look for maybe what else I could reuse. 
I think that eventually however it would be nice to have some benchmarks with convolutional neural networks / generative neural networks / deep q networks / and other AI field stuff, where parameter tuning makes a large difference. These would normally require some extra dependencies. But it is good to start simple.

betatim:
Starting to look good. If you are ready for review add "[MRG]" to the start of the title of this PR. Or feel free to ask questions in the mean time ;)

iaroslav-ai:
@betatim: Yup :) Will add MRG to the title now.

iaroslav-ai:
I switched to use sklearn only, and because of that many of the datasets used previously are not used now. These can be returned later. Currently there are 4 datasets, 2 for regression and 2 for classification. 
Now the code outputs in console the data in same format as in README.MD, which one should be able to copy paste without too much hassle. Here is example benchmark when *only decision trees are used* and for 50 evaluations budget:

GaussianProcessRegressor
|Blackbox Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min
------------------|------------|-----------|---------------------|--------------------|-----------------------
|ML_hypp_tune|-0.654 +/- 0.002|-0.656|32.600|10.892|22.000

ExtraTreesRegressor
|Blackbox Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min
------------------|------------|-----------|---------------------|--------------------|-----------------------
|ML_hypp_tune|-0.653 +/- 0.003|-0.656|32.400|5.571|28.000

RandomForestRegressor
|Blackbox Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min
------------------|------------|-----------|---------------------|--------------------|-----------------------
|ML_hypp_tune|-0.651 +/- 0.002|-0.655|31.200|12.496|10.000

This is the format I found 
Currently what benchmark does is for particular estimator class (and particular repeat) it goes over all feasible combinations of modes / datasets, and calculates average objective over all such combinations at 1st step of optimization, 2nd optimization step etc. One could however split the benchmark eg into sub- benchmark with decision trees only, nn only, etc. Would be curious to know your input on this and let me know if I am not clear on this. 

Looking forward to your review! :)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,445 @@
...
+            return ("%.3f" % float_value)
+
+        output = "|".join([fmt(min_mean) + " +/- " + fmt(min_stdd)] + [fmt(v) for v in [ min_best, f_mean, f_stdd, f_best]])
+        result = table_template + output
+        print("")
+        print(Algorithm)
+        print(result)
+
+
+def evaluate_optimizer((base_estimator, model, dataset, max_iter, seed_value)):
```
MechCoder: The double bracket raises a Syntax Error.

iaroslav-ai: Yup, can confirm python 3.x does not like double bracket. Will fix now.

iaroslav-ai: I made changes so that code runs with python 3.4. Let me know if this still raises errors.

betatim:
Super sorry for the lack of reviewing after you put in the work. It seems like all three have to focus on non-skopt things for the moment. So this isn't a comment on your work :) We will get to this, promised. If someone else has time and wants to step up, please do!

iaroslav-ai:
Hi, no problem :) It is not too urgent anyway. Thanks for letting me know! :)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+
+        try:
+            model.fit(X, Y)
+            r = model.score(Xv, Yv)
+        except BaseException as ex:
+            r = 0.0 # on error: return assumed smallest value of objective function
+
+        # while negative values could be informative, they could be very large also,
+        # which could mess up the optimization procedure. Suggestions are welcome.
+        r = max(r, 0.0)
```
MechCoder: As an alternative, we can use the mean squared error instead of the r2_score. Also it might be better to use the `cross-entropy` loss since accuracy is not continuous.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+    # OUTPUT_TYPE:  FMT_NUMBER / FMT_CATEGORY
+    # MODEL_PARAMETERS: dictionary whose keys correspond to parameters that are passed to __init__ of python class
+    #                   that implements model
+    # MODEL_BACKEND: python class that implements model
+    models = {
+        #
+        #   regressors go here
+        #
+        MLPRegressor.__name__:{
+            OUTPUT_TYPE: FMT_NUMBER,
```
iaroslav-ai: FMT is a shortened ForMaT

benchmarks/bench_ml.py
iaroslav-ai: I considered using MSE, but this is why I did not. The larger the values of outputs, the larger the  magnitude of the values of the MSE. So it might happen that surrogate A is good for one specific task with large values of output (and thus large MSE), which makes a large contribution to  average performance over multiple tasks, compared to surrogate B which is better for other tasks, but other tasks do not make a large contribution due to values of output being relatively small. Thus it might seem that surrogate A performs on average better than B, even though if you look at case by case performance it is not. One could solve this by normalizing values of MSE, but this is what r2_score already does. Additionally, range of values of  r2_score is in most cases in range of [0..1.0], which is also the range for accuracy.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+    Main function used to run the experiments.
+
+    :param n_calls: evaluation budget
+    :param n_runs: how many times to repeat the optimization in order to average out noise
+    :param save_traces: whether to save data collected during optimization
+    :param run_parallel: whether to run different repeats of optimization in parallel
+    :return: None
+    """
+
+    selected_regressors = [GaussianProcessRegressor, RandomForestRegressor, ExtraTreesRegressor]
```
iaroslav-ai: Yeah, surrogates actually sounds better :)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+    selected_models = TesterClass.models.keys()
+    selected_datasets = TesterClass.datasets.keys()
+
+    pool = Pool()
```
iaroslav-ai: Looks interesting, will take a look!

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+
+            if not Supports:
+                continue
+
+            all_data[Model][Dataset] = {}
+
+            for Regressor in selected_regressors:
+                print(Regressor.__name__, Model, Dataset)
+
+                params = [(Regressor, Model, Dataset, n_calls, np.random.randint(2 ** 30)) for rep in range(n_runs)]
```
iaroslav-ai: @MechCoder thanks for comments! Will update the code accordingly.

iaroslav-ai:
@MechCoder thanks for comments! Will update the code accordingly.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,451 @@
...
+            dim_range = v[VARIABLE_RANGE]
+            dim = Categorical(dim_range)
+        else:
+            raise BaseException("Unknown type of variable!")
+
+        dimensions.append((dim, k)) # need to remember names of dimensions for evaluation code
+
+    # initialization
+    estimator = base_estimator(random_state=seed_value)
+    solver = Optimizer([d[0] for d in dimensions], estimator, acq_optimizer="sampling", random_state=seed_value)
```
iaroslav-ai: Currently default for acq_optimizer is "lbfgs", which for example does not work with GaussianRegressor class when I do not specify kernel for this class - it throws AttributeError: 'Product' object has no attribute 'gradient_x' . I think you have already opened issue for this. This does not work also with RandomForestRegressor (TypeError: predict() got an unexpected keyword argument 'return_mean_grad') and same with ExtraTreesRegressor.

MechCoder: The first one should be fixed now, you need to rebase to get the changes. For `forest_minimize` the default acquisition function is "sampling".

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+        #
+        "Leukemia":{
+            OUTPUT_TYPE: OUTPUT_CATEGORY,
+            DATASET_LOADER: lambda : load_data_target(skd.fetch_mldata('leukemia'))
+        },
+        "Climate Model Crashes":{
+            OUTPUT_TYPE: OUTPUT_CATEGORY,
+            DATASET_LOADER: lambda : load_data_target(skd.fetch_mldata('climate-model-simulation-crashes', target_name=-1))
+        },
+    }
```
MechCoder: Why are these variables models and datasets bound to the object? I suppose they can be safely moved outside?

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+            return ("%.3f" % float_value)
+
+        output = "|".join([fmt(min_mean) + " +/- " + fmt(min_stdd)] + [fmt(v) for v in [ min_best, f_mean, f_stdd, f_best]])
+        result = table_template + output
+        print("")
+        print(Algorithm)
+        print(result)
+
+
+def evaluate_optimizer(base_estimator, model, dataset, n_calls, seed_value):
```
MechCoder: `base_estimator` -> `surrogate`

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+
+    # below seed is necessary for processes which fork at the same time
+    # so that random numbers generated in processes are different
+    np.random.seed(seed_value)
+
+    problem = TesterClass(model, dataset)
+    space = problem.model_class[MODEL_PARAMETERS]
+
+    # initialization
+    estimator = base_estimator(random_state=seed_value)
```
MechCoder: It seems to me that `base_estimator` is an instance. In that case how does this work?

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+    def __init__(self, model, dataset):
+
+        supports, reason_if_not = TesterClass.supports(model, dataset)
+        if not supports:
+            raise BaseException(reason_if_not)
+
+        X, Y = self.datasets[dataset][DATASET_LOADER]()
+
+        self.dataset = split_normalize(X, Y)
+        self.model_class = TesterClass.models[model]
```
MechCoder: `space` might be a better name for this attribute.

MechCoder: `space` might be a better name for this attribute.

benchmarks/bench_ml.py
MechCoder: `base_estimator` -> `surrogate`

benchmarks/bench_ml.py
MechCoder: Why are these variables bound to the object? Seems like they can be safely moved outside.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+    Main function used to run the experiments.
+
+    :param n_calls: evaluation budget
+    :param n_runs: how many times to repeat the optimization in order to average out noise
+    :param save_traces: whether to save data collected during optimization
+    :param parallel_jobs: whether to run different repeats of optimization in parallel
+    :return: None
+    """
+
+    surrogates = [get_gaussian_surrogate]
```
MechCoder: Just FYI, you can supply the instance directly and use the `set_params` interface to set the random_state since that seems to be your intention. (http://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.set_params)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+
+    trace = []
+    best_y = np.inf
+    best_x = None
+
+    # optimization loop
+    for i in range(n_calls):
+        point_list = solver.ask()
+
+        point_dct = {k: v for k, v in zip(space.keys(), point_list)}  # convert list of dimension values to dictionary
```
MechCoder: I think we should sort the keys of the space configuration both here and while specifying the dimensions to guarantee they are entered and received in the same order.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+
+        # apply transformation to model parameters, for example exp transformation
+        point_mapped = {}
+
+        for param in point.keys():
+            v = point[param]
+            v = self.model_class[MODEL_PARAMETERS][param][1](v)
+            point_mapped[param] = v
+
+        model = cls(**point_mapped)
```
MechCoder: You should make `self.model_class[MODEL_BACKEND]` to be an instance instead of an object, and all the above block of code can be rewritten as `self.model_class[MODEL_BACKEND].set_params(**point)`, (assuming point is a dict of parameter names and parameter values)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+
+        try:
+            model.fit(X, Y)
+            r = model.score(Xv, Yv)
+        except BaseException as ex:
```
MechCoder: You should not catch using `BaseException` since this will be unable to catch any specific error raised such as a `ValueError`.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+            },
+            MODEL_BACKEND: SVR,
+        },
+        DecisionTreeRegressor.__name__:{
+            OUTPUT_TYPE: OUTPUT_CONTINUOUS,
+            MODEL_PARAMETERS: {
+                'max_depth': (Real(1.0, 4.0), pow2intmap),
+                'min_samples_split': (Real(1.0, 8.0), pow2intmap),
+            },
+            MODEL_BACKEND: DecisionTreeRegressor,
```
MechCoder: If you make this an instance instead of a class, you can infer `OUTPUT_TYPE` using `isinstance(model, RegressorMixin)` or `isinstance(model, ClassifierMixin)` where the mixins are inherited from `sklearn.base`

MechCoder:
It is also deserves a doc-string at the top of the file to summarize what the benchmarks are about.

iaroslav-ai:
I just realized that I did not rebase the repo to the most recent version, so I tried to do that. I think this is why there appear to be so many commits :)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+    Splits data into training and validation parts.
+    Test data is assumed to be used after optimization.
+
+    :param X: matrix where every row is a training input
+    :param y: vector where every entry is correpsonding output
+    :return: split of data into training and validation sets.
+            70% of data is used for training, rest for validation.
+    """
+
+    X, Xv, y, yv = train_test_split(X, y, train_size=0.7)
```
iaroslav-ai: Yup added

benchmarks/bench_ml.py
iaroslav-ai: I think this is a good idea, I extended it even further - I used `issubclass`, this way the code is reduced even more

benchmarks/bench_ml.py
iaroslav-ai: Currently those are static variables for the class. Do you think it would be better if I would move them outside?

benchmarks/bench_ml.py
iaroslav-ai: Renamed

benchmarks/bench_ml.py
iaroslav-ai: The above block of code applies transformation to the values of the point being evaluated, so for example to the value of C = -1.0 the exponent is applied so that it becomes 0.1 (there are some other transformations like rounding and integer exponentiation). This is what the loop above the cls(**point_mapped) is for. With how things are right now I can just write `self.model_class[MODEL_BACKEND].(**point_mapped)`, where the point_mapped contains transformed values.

benchmarks/bench_ml.py
iaroslav-ai: The way it is right now is so that if in any way the parameters provided by the optimizer which break the fitting or evaluation (and thus cause any kind of exception) of the model yield the objective of 0, so that solver learns which parameters are infeasible.

iaroslav-ai: Thus I just catch the BaseaException, and on this exception I set the value of objective to 0.0. I think there has been a discussion about constrained black box optimization, where you could explicitly specify which configurations of parameters are feasible or not, but I think you or some other member suggested that this is encoded into the objective being optimized. I cannot find atm where you had this discussion though.

benchmarks/bench_ml.py
iaroslav-ai: renamed

benchmarks/bench_ml.py
iaroslav-ai: Yup, I added `dimensions_names = space.keys()` variable in the beginning of evaluate_optimizer, which is used to ensure the right order of keys.

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+        solver.tell(point_list, objective_at_point)
+
+    return trace
+
+# the function below is necessary as a workaround for the pool.map method requiring map function to take only a single argument
+def _evaluate_optimizer(params):
+    return evaluate_optimizer(*params)
+
+def get_gaussian_surrogate(random_state):
+    return GaussianProcessRegressor(kernel=Matern(), random_state=random_state)
```
iaroslav-ai: When I tested it last time, I did not rebase the repo. Now it works fine.

iaroslav-ai: With default regressor I mean

benchmarks/bench_ml.py
iaroslav-ai: I see. Currently it seems that supplying the class as such instead of instance keeps code a bit shorter (but I am open to suggestions of course)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+
+        for dataset in selected_datasets:
+            supports, _ = TesterClass.supports(model, dataset)
+
+            if not supports:
+                continue
+
+            all_data[model][dataset] = {}
+
+            for surrogate in surrogates:
```
iaroslav-ai: The reason I did this is so that it is a bit easier later to do visualizations if you want to do comparison of different surrogates on one particular problem, then you just take `all_data[model][dataset]` for some model and dataset, and visualize the results that are stored in there. It might be that some particular dataset / model or subset of those is more representative of some particular optimization task for which you want to choose which surrogate to use, so you might just consider those particular results instead of all results. It might be good also if at some point we decide to split this one large benchmark into smaller ones (just of top of my head)

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+                print(surrogate.__name__, model, dataset)
+
+                raw_trace = Parallel(n_jobs=1)(
+                    delayed(evaluate_optimizer)(
+                        surrogate, model, dataset, n_calls, np.random.randint(2 ** 30)
```
iaroslav-ai: Oh, fixed. Is even evaluation of arguments (that is `np.random.randint(2 ** 30)`) that are supplied to `evaluate_optimizer` are done in separate processes? I thought that it is still done in the main process, so that different seeds are fed into different processes

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+    calculate_performance(all_data)
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        '--n_calls', nargs="?", default=64, type=int,
+        help="Maximum number of allowed function calls.")
```
iaroslav-ai: renamed

benchmarks/bench_ml.py
```diff
@@ -0,0 +1,422 @@
...
+    parser.add_argument(
+        '--n_runs', nargs="?", default=5, type=int,
+        help="Number of re-runs of single algorithm on single instance of a problem.")
+    parser.add_argument(
+        '--save_traces', nargs="?", default=False, type=bool,
+        help="Whether to save pairs (point, objective) obtained during experiments in a json file.")
+    parser.add_argument(
+        '--parallel_processes', nargs="?", default=-1, type=bool,
```
iaroslav-ai: renamed

iaroslav-ai:
@MechCoder I made a pass over comments - please take a look. I will extend a bit more the docstring soon.

benchmarks/bench_ml.py
iaroslav-ai: One could add some warning on error though

MechCoder:
I think there are some spurious commits in the diff. Would you be able to fix those?

MechCoder:
`git reflog` is your friend to figure out where you missed things up :-)

iaroslav-ai:
Hi, sorry for the silence - I was away. Will take a look over the next few days.

PR #322, opened 2017-02-16
[MRG] Fix some pep8 violations

betatim:
Minor tweaks and one bug fix. Should also increase coverage.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -48,12 +48,12 @@ def check_dimension(dimension, transform=None):
           - "onehot" (default) one-hot transformation of the original space.
           - "identity" same as the original space.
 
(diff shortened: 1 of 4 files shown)


skopt/space/space.py
```diff
@@ -254,7 +256,7 @@ def __init__(self, low, high, transform=None):
         if transform not in ["normalize", "identity"]:
             raise ValueError(
                 "transform should be 'normalize' or 'identity' got %s" %
-                transform_)
+                self.transform_)
```
betatim: this is the only code change

betatim:
I am sure we could do more but maybe worth just merging this quickly.

glouppe:
Looking good, merging.

PR #323, opened 2017-02-17
[MRG] Move creation of result object to `Optimizer` class

### LINKED ISSUE #315: Optimizer does not create `OptimizeResult` objects
Ask and tell interface does not create a `OptimizeResult` object like `XXX_minimize` does which makes it harder to use the plotting tools in `skopt`.

`Optimizer.tell()` could return a result object that is compatible with our plotting tools and for API uniformity reasons.

---

This was pointed out in #305

betatim:
Fixes #315

Move the call to `create_result` to `Optimizer` so that we can use the plot tools we have with ask-and-tell.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -214,6 +214,9 @@ class Real(Dimension):
     def bounds(self):
         return (self.low, self.high)
 
+    def contains(self, point):
+        return self.low <= point <= self.high
+
     @property
     def transformed_bounds(self):
         if self.transform_ == "normalize":
@@ -283,6 +286,9 @@ class Integer(Dimension):
     @property
     def bounds(self):
         return (self.low, self.high)
+
+    def contains(self, point):
+        return self.low <= point <= self.high
 
     @property
     def transformed_bounds(self):
@@ -375,6 +381,9 @@ class Categorical(Dimension):
     @property
     def bounds(self):
         return self.categories
+
+    def contains(self, point):
+        return point in self.categories
 
     @property
     def transformed_bounds(self):
@@ -568,6 +577,13 @@ class Space:
                 b.extend(dim.bounds)
 
         return b
+
+    def contains(self, point):
+        """Check that `point` is within the bounds of the space."""
+        for component, dim in zip(point, self.dimensions):
+            if not dim.contains(component):
+                return False
+        return True
 
     @property
     def transformed_bounds(self):
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -129,6 +129,13 @@ class Dimension(object):
         raise NotImplementedError
 
 
+def _uniform_inclusive(loc=0.0, scale=1.0):
+    # like scipy.stats.distributions but inclusive of `high`
+    # XXX scale + 1. might not actually be a float after scale if
+    # XXX scale is very large.
+    return uniform(loc=loc, scale=np.nextafter(scale, scale + 1.))
+
+
 class Real(Dimension):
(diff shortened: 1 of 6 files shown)


skopt/space/space.py
```diff
@@ -214,6 +214,9 @@ def inverse_transform(self, Xt):
     def bounds(self):
         return (self.low, self.high)
 
+    def __contains__(self, point):
+        return self.low <= point <= self.high
```
betatim: are these the right comparison operators? (We should document this in the doc string)

skopt/space/space.py
```diff
@@ -284,6 +287,9 @@ def inverse_transform(self, Xt):
...
```
betatim: are these the right comparison operators? (We should document this in the doc string)

betatim:
As well as moving the creation of the results object I also added some bounds checking in this PR. Ready for reviewing.

skopt/space/space.py
glouppe: We have documented the upper bound of a `Real` dimension to be exclusive.

glouppe:
Looks good to me, except the comment for the upper bound.

skopt/space/space.py
betatim: Sorry, didn't spot it in the doc string :(

betatim:
Fixed the operators in the bounds checking and modified the tests a bit. They were using identical comparison operators for both `Integer` and `Real`. However one is inclusive and the other exclusive at the upper boundary.

betatim:
Mistake in my implementation somewhere or is our docstring wrong? https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/203851498

betatim:
I think it is because [`fmin_l_bfgs_b()`](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.optimize.fmin_l_bfgs_b.html) interprets the bounds differently (both ends inclusive) to us.

Ahh yiis: https://github.com/scipy/scipy/blob/master/scipy/optimize/lbfgsb/README

skopt/space/space.py
```diff
@@ -214,15 +214,18 @@ def inverse_transform(self, Xt):
...
         return (self.low, self.high)
 
+    def __contains__(self, point):
+        return self.low <= point < self.high
+
     @property
     def transformed_bounds(self):
         if self.transform_ == "normalize":
-            return 0.0, 1.0
+            return 0.0, 1.0 - np.finfo(float).eps
```
betatim: Is this the best we can do? Or should we modify the upper edge just before the call to BFGS?

skopt/space/space.py
```diff
@@ -214,15 +214,19 @@ def inverse_transform(self, Xt):
...
 
+    def __contains__(self, point):
+        return self.low <= point < self.high
+
     @property
     def transformed_bounds(self):
+        eps = np.finfo(float).eps
         if self.transform_ == "normalize":
-            return 0.0, 1.0
+            return 0.0, 1.0 - eps
```
betatim: Is this the best we can do? Or should we modify the upper edge just before the call to BFGS?

betatim:
Decided to change the behaviour of `Real` to be inclusive of upper edge to be consistent with BFGS behaviour. Not sure if this adaptation of `scipy.stats.distributions.uniform` does what I think it does.

betatim:
`np.all(p in self.space for p in x)` can you spot the bug in this?

Time to merge this.

glouppe:
Looks good to me! Didnt know about `nextafter`, learned something :)

PR #327, opened 2017-03-02
[WIP] Fix passing on the verbose flag in GBRT

### LINKED ISSUE #326: gbrt_minimize, verbose=True not working
Hi,

I am trying to use gbrt_minimize with verbose=True, but it's not printing anything to the screen. forest_minimize works fine.

Thanks,

Mario

betatim:
Fixes #326 

What would be a good way to test for output when the verbose flag is set in a test (without having to resort to replacing `sys.stdout` myself)?

### DIFF
--- a/skopt/optimizer/gbrt.py
+++ b/skopt/optimizer/gbrt.py
@@ -145,4 +145,5 @@ def gbrt_minimize(func, dimensions, base_estimator=None,
                          n_random_starts=n_random_starts,
                          x0=x0, y0=y0, random_state=random_state, xi=xi,
                          kappa=kappa, acq_func=acq_func,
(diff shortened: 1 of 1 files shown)


betatim:
From some googling it seems that in nosetest there is no other way than to replace `sys.stdout` with a `BytesIO`. Which is simple enough but starts to get annoying if you want to run the test in pdb.

MechCoder:
lgmt

betatim:
I'd like to add a test for this before we merge (unless there is a need to merge quickly in which case we can punt the test to a new issue).

glouppe:
I am happy with merging this now.

betatim:
Done.

Issue #325, opened 2017-02-21
Knowledge transfer for black box optimization


iaroslav-ai:
Consider that I want to come up with an algorithm that optimizes treatment parameters  for patients (eg amount & type of medication to take).

Currently if I use algorithms from skopt, I would need to start optimization for particular patient from scratch (correct me if I am wrong). However, I will likely have a lot of data from previous patients, and one can expect it to be useful to converge faster to better solutions. Additionally, for different patients I could provide some features like age etc., which could help even further.

Do you have any plans to include a possibility of knowledge transfer between optimization problems? Maybe something like this: http://papers.nips.cc/paper/5086-multi-task-bayesian-optimization.pdf  .

iaroslav-ai:
One reason why I am interested is that I run many experiments with datasets which are “similar”, to optimize neural network parameters on these datasets. It would be great to reuse data from previous experiments to boost convergence to good neural network architectures on new ones.

iaroslav-ai:
I have been experimenting with this recently. I came up with a code [here](https://gist.github.com/iaroslav-ai/ded34b4e492f1b5dc8e9ddbe3c5cb166), where I have a simplistic   implementation of a modified sequential optimization with GP that can take advantage of similarity of a set of optimization tasks.

Every such “similar” task corresponds to tuning of 3 parameters of SVM on a different synthetic dataset. 

Here is what I get if I do not provide any features describing a task (but data for prior tasks is available):

![figure_1](https://cloud.githubusercontent.com/assets/14038787/23600614/b0ddbaa2-0247-11e7-8f00-ac20d6b33813.png)

Here is what happens if I provide features that tell something abouta a task being solved:

![figure_1-1](https://cloud.githubusercontent.com/assets/14038787/23600623/bd2b03a0-0247-11e7-952d-c547d8a518a3.png)

In both cases I get less of a cold start problem, which is exactly what I am looking for. The data shown in graph is an average for 8 different tasks, repeated 5 times. Knowledge transfer was done for 8 consecutive tasks. 

Here is what I do: I consider a sequence of “similar” tasks as one large optimization task. For every point p and value y that I collect, I transform it into a larger vector by concatenating: 

[f][h][p] = p’

[...]

PR #332, opened 2017-03-11
Knowledge Transfer

iaroslav-ai:
I implemented a knowledge transfer strategy as was discussed in #325 .

Currently the code can work with all acquisition functions and surrogates, and should work with new ones if added. I tried to make the changes to the code minimal. 

I have a toy problem to test the code in tmp/example.py.. There, 5 optimization tasks which share some structure are solved. Every such task corresponds to parameter tuning of SVM on regression learning task. Some features are available to the optimizer that provide extra information about instance of optimization problem.

Here are the results, when following strategy outlined in #325 :

figure_1

In the subplot title, "Task number" means that modified knowledge transfer optimizer was only provided with the index of similar task being solved. "Task features and number" means that optimizer was provided with both index of similar task and some features describing the task. 

 Clearly, initially the solver which utilizes prior knowledge experiences much less of a cold start problem. However, with more iterations the solver which starts from scratch (no prior knowledge) achieves better objective values. 

My current hypothesis is that solver with more data (prior knowledge) becomes overconfident about objective values in unexplored regions, which prevents the solver from getting to the good values (quickly).

One strategy that I intend to test  to circumvent this is to take the best of two worlds - start with solver with prior knowledge, and after some number of iterations to switch to solver with no prior knowledge. I expect this to yield the minimum of two curves.  

This might be a bug in my code of course, but it seems like it is not. I will also test this on instances of optimization with larger number of parameters to tune. 

Consider this PR a more of an "experiment in progress" to keep you updated about this. As I will have more data I will post it here.


iaroslav-ai:
It seems that I have added changes to the older version of skopt. Will take a look on how to move this to the newer one.

iaroslav-ai:
@betatim sorry, I know you asked for the minimal pull request - but it seems that overall integrating this does not take much change, most of the code in the commit is actually code of the example problem. Plus it is always a possibility that there is something completely different that is better than the current approach, so I decided to close the loop.

iaroslav-ai:
Hmmm I will close this one for now, and as I will put the updates into newest version of skopt and as I will fix the convergence, I will create a new PR. I apologize if I am spamming you with this :)

Issue #335, opened 2017-03-30
Minimum value to report


glouppe:
For the moment, the minimum value reported is the minimum value among the sampled points, for which we know the true value of the objective.

What do you think of also adding to the results the _expected_ location of the minimum, as modeled by the last model on the data points? (I.e., find the location of minimum over the (expected) decision surface and associate the expected function value to that minimum.)

If not packed into the results, a short helper `x, fx = expected_minimum(results)` would be handy.

betatim:
I like the idea. Adding a helper might be better than putting it in the result object. Mostly because I can't think of a good way to name the attribute that won't lead to it being confused with the "other" minimum. Not sure which one we'd want people to use but having `res.minimum` and `res.actual_minimum` or what ever seems like a recipe for confusion.

---

😕 decision surface? You mean the surrogate model no?

glouppe:
> You mean the surrogate model no?
Sorry, about that.

MechCoder:
How do you infer that directly from the results object?

glouppe:
> How do you infer that directly from the results object?

This is not immediate, but you can find the value x that minimizes that mean prediction of the GP.
(With a fine grid and/or by running lbfgs from the promising points.)

glouppe:
In other words, this amounts to return the next best query point, turning off exploration completely.

betatim:
I think I have a slight preference for [providing a helper](https://github.com/scikit-optimize/scikit-optimize/issues/335#issue-218215549) to do this.

glouppe:
I'll work on this today!

PR #336, opened 2017-03-31
[MRG] Early stopping

betatim:
This will implement one of the criteria mentioned in #6. The main purpose of this PR is to provide a mechanism to stop evaluating the objective before exhausting all `n_iter` calls.


betatim:
This implements early stopping based on `|x_t - x_t-1| < delta_x`.

Should we add an argument to the minimizer API that is called `early_stopping[_kwargs]` which collects the arguments? Otherwise in the future we will have lots and lots of arguments `delta_x`, `delta_y`, `threshold`, etc. What do you think?

Another (longer term) architecture question is how do we split this functionality out so that you can easily reuse it when using the ask-and-tell interface? Similar question as for #335 which could reuse a lot of the "minimize this surrogate model" functionality of the `Optimizer` class.

glouppe:
> Should we add an argument to the minimizer API that is called early_stopping[_kwargs] which collects the arguments? Otherwise in the future we will have lots and lots of arguments delta_x, delta_y, threshold, etc. What do you think?

Would it be possible maybe to have early stopping as a callback? That would be a way to factor out nicely all the logic.

betatim:
Do you know how much python magic we would need to use for something like this to work:
```python
def gp_optimize(...., stopping=...):
  ...
  if stopping(next_x, results_so_far):
    break

# user calls
gp_optimize(..., stopping=delta_x(0.1) or delta_y(1e-5) and y_below(1e-8))
```
because that would be pretty rad (as the yanks say). I know how to make something that returns a new callable but I don't know how much magic python sauce we need to support logical operators on callables that then form a new callable...

Cute but too much technical debt or something worth investigating?

glouppe:
Not convinced adding a syntax for the stopping criterion would be useful.

For the code structure, it seems that something like https://github.com/fchollet/keras/blob/master/keras/callbacks.py#L414 would be a good example as a callback. 

(The trick is for the callback to modify the state of the caller object.)

MechCoder:
+1 for early stopping as a callback.

betatim:
Thoughts on this?

glouppe:
Hmm, what I had in mind was something similar to our VerboseCallback.

betatim:
If you have a more complicated criteria you can implement it as a class with a `__call__` method no?

Interface wise I think all we need to agree is that a stopping criteria is callable and takes `next_x` and `results_so_far` as arguments, makes a decision somehow, returns `True`/`False`.

I would be -1 on allowing a list of stopping criteria to be passed in, because then you need to decide how to combine those criteria (AND or OR) so instead we should make the user write a short callable that combines them as they want it, pass that in.

MechCoder:
There are already too many arguments in the function call, I get distracted by the time I reach the end. Why not just create an `EarlyStoppingCallback` and make it public?

betatim:
Ah, I think now I understand what you mean...not a new argument with callbacks but use the existing callback parameter.

MechCoder:
+1.

betatim:
WDYT? I tried to keep the API compatible with what we had before.

Callbacks can now return `True` or `False` to participate in the decision to stop or return `None` to indicate that they don't have an opinion. The argument to `__call__` remains the same as before, through this you can access the results, models, space and optimizer object.

glouppe:
Thanks! I think it is much better like this :)

Now regarding the stopper in itself, do you think it is a good strategy? It is very likely that the current optimum wont improve because of exploration. I think it would be more useful if the criterion was defined as "stop if no improvement larger than delta within the N previous steps", where N is set to a default value larger than 2, say 10 or 20. What do you think?

betatim:
Yes, now that I finally get what you guys were talking about I also prefer it :)

> Now regarding the stopper in itself, do you think it is a good strategy? It is very likely that the current optimum wont improve because of exploration. I think it would be more useful if the criterion was defined as "stop if no improvement larger than delta within the N previous steps", where N is set to a default value larger than 2, say 10 or 20. What do you think?

For me this is just a first criterion to illustrate how we setup the technicalities. I will add more from https://github.com/scikit-optimize/scikit-optimize/issues/6#issuecomment-220965349

Note that the DeltaX criterion looks at the distance between positions at which the objective was evaluated. So this is about "we keep sampling in the same location, lets stop now".

glouppe:
> Note that the DeltaX criterion looks at the distance between positions at which the objective was evaluated. So this is about "we keep sampling in the same location, lets stop now".

Oh right, I read too fast. I thought you were considering |f(x_i) - f(x_{i-1})| < delta. 

+1 for merge, unless you want keep adding other criteria directly here.

betatim:
Let's add stopping criteria in parallel ;)

betatim:
Thanks for the patience and re-re-re-stating your idea till I got it

skopt/optimizer/optimizer.py
```diff
@@ -186,31 +186,15 @@ def ask(self):
...
-                    cat_eq = np.all(x_arr[cat_inds] == next_x_arr[cat_inds])
-                    non_cat_eq = np.allclose(
-                        np.array(x_arr[non_cat_inds], dtype=np.float32),
-                        next_x_non_cat)
-                    if cat_eq and non_cat_eq:
-                        warnings.warn("The objective has been evaluated "
-                                      "at this point before.")
+            min_delta_x = np.min([self.space.distance(next_x, xi)
+                                  for xi in self.Xi])
+            if np.allclose(min_delta_x, 0.):
```
MechCoder: Nice, thank you!

skopt/utils.py
```diff
@@ -53,6 +53,36 @@ def create_result(Xi, yi, space=None, rng=None, specs=None, models=None):
...
+    Returns
+    -------
+    * `decision` [bool]:
+        Decision of the callbacks whether or not to keep optimizing
+    """
+    stop = False
+    if callbacks:
+        for c in callbacks:
+            decision = c(result)
+            if decision is not None:
```
MechCoder: Same as?

```python
if decision is not None and decision:
    return True
```

(minor benefit being that is stops early? meta not intended :P )

MechCoder:
Thanks! Looks great.

MechCoder:
Just realized that you merged it already. Sorry for being late, struggling to keep up pace.

skopt/utils.py
betatim: heh, I think we have to evaluate all callbacks: imagine the callback list is `TimerCallback`, `DeltaXStopper`, `VerboseCallback`. You'd want to have all of them called all the time.

MechCoder: oh yeah, I was being too smart

skopt/callbacks.py
```diff
@@ -125,3 +140,39 @@ def __call__(self, res):
...
+
+    def _criterion(self, result):
+        if len(result.x_iters) >= 2:
+            if result.space.distance(result.x_iters[-2],
+                                     result.x_iters[-1]) < self.delta:
+                return True
+            else:
+                return False
+        else:
+            return None
```
betatim: I decided to make it explicit (also the extra else branch) as a way of showing the difference between `False` and `None`. Not helpful/unnecessary

betatim:
Addressed comments in c6c3210 as part of #353

betatim:
Regarding naming convention: I think having Callback in the name is not needed really as all callbacks are in the callback module so you'd end up with a lot of code like `callbacks.TimerCallback`. So I would vote for removing the Callback ending. Should we move this to another issue/PR?

PR #337, opened 2017-03-31
[MRG] Add test for two vs one dimensional problems with initial points

betatim:
The test I added would have failed before. The entries in `res.x_iters` had different dimensions for a 1D problem.

### DIFF
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -17,6 +17,7 @@ from skopt import gp_minimize
 from skopt import forest_minimize
 from skopt import gbrt_minimize
 from skopt.benchmarks import branin
+from skopt.benchmarks import bench1
 from skopt.benchmarks import bench4
 from skopt.benchmarks import bench5
 from skopt.space import Space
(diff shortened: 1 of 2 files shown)


skopt/tests/test_common.py
```diff
@@ -306,7 +306,26 @@ def test_repeated_x():
...
+def test_consistent_x_iter_dimensions():
+    # check that all entries in x_iters have the same dimensions
+    for minimizer in MINIMIZERS:
+        # two dmensional problem
+        res = minimizer(bench4,
+                        dimensions=[(0, 1), (2, 3)],
+                        x0=[[0, 2], [1, 2]], n_calls=3,
+                        n_random_starts=0)
+        assert len(set(len(x) for x in res.x_iters)) == 1
+        assert len(res.x_iters[0]) == 2
```
MechCoder: These tests pass for me on master. Did you write this just to increase test coverage?

skopt/tests/test_common.py
```diff
@@ -306,7 +306,26 @@ def test_repeated_x():
         assert_warns_message(
             UserWarning, "has been evaluated at", minimizer, lambda x: x[0],
             dimensions=[[0, 1]], x0=[[0], [1]], n_random_starts=0, n_calls=3)
+
         assert_warns_message(
             UserWarning, "has been evaluated at", minimizer, bench4,
-            dimensions=[("0", "1")], x0=["0", "1"], n_calls=3,
+            dimensions=[("0", "1")], x0=[["0"], ["1"]], n_calls=3,
```
betatim: This typo is what made me write the new test that explicitly checks that all `res.x_iters` entries have the same shape. `test_repeated_x` did pass before, but it wasn't doing what it should have.

betatim: Maybe we should also add some protection in the optimizer to catch this kind of mistake (I think the `assert` statements would cover this, but didn't test it.

betatim:
Added to the new test a bit,  I think this is better as it would have stopped the tests from passing.

glouppe:
Thanks! Good catch.

PR #341, opened 2017-04-06
[MRG] Add test for dump()/load() of Optimizer class

### LINKED ISSUE #316: skopt.load/dump can not handle `Optimizer` instances
`Optimizer` instances can not be serialised with `skopt.dump/load`. This is probably unexpected for users, or at the very leat inconvenient. More discussion in #305

It would make sense to extend `dump/load` to support "all" of the `skopt` classes etc.

Needs a bit of discussing in terms of API as this is mainly syntactic sugar on top of `pickle`, so we need to find a way that is actually nicer to use than the bare `pickle`.

betatim:
Fixes #316 

Added a test to check that the dump()ing and load()ing of `Optimizer` instances works., because "it would be a good idea" 😆 

cc @glouppe

### DIFF
--- a/skopt/tests/test_utils.py
+++ b/skopt/tests/test_utils.py
@@ -7,7 +7,10 @@ from sklearn.utils.testing import assert_true
 from skopt import gp_minimize
 from skopt import load
 from skopt import dump
+from skopt.benchmarks import bench1
 from skopt.benchmarks import bench3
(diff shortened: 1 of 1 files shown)


glouppe:
Tests pass. Let's merge!

betatim:
:clinking_glasses:

PR #339, opened 2017-04-04
[MRG] Supervised Learning benchmark

iaroslav-ai:
This is a continuation of https://github.com/scikit-optimize/scikit-optimize/pull/317 .
@MechCoder It seemed like the easiest solution was to just make a new fork of the repo and to copy the bench_ml.py file there. Otherwise I risk messing up things again with commits :)

PR #343, opened 2017-04-08
[MRG+1] Parallel optimization with skopt

iaroslav-ai:
A typical use of Bayesian optimization is in sequential mode, where one does only one experiment (function evaluation) at a time. However, one could hope to reduce the overall wall clock time by running a few function evaluations / experiments at a time.

Approaches that try such parallelisation typically yield speed up of convergence wrt wall clock time somewhere proportional to the number of experiments run in parallel, eg:

https://arxiv.org/pdf/1602.05149.pdf (see page 13)
https://people.orie.cornell.edu/pfrazier/Presentations/2015.02.CMU.pdf (see slide 32)

Currently such feature is not supported by skopt explicitly (correct me if I am wrong), but one could nonetheless attempt some simple approaches already. One such approach is to sample multiple points that minimize current surrogate, more precisely something like calling the below function multiple times, and evaluating results in parallel:

```
def get_point(X,Y):
    """
    :param X: All points evaluated so far
    :param Y: Objective values at corresponding points
    :return: next point to evaluate
    """
    opt = Optimizer(dimensions=dimensions,base_estimator=some_surrogate)
    opt.tell(X, Y)
    return opt.ask()
```

Assume for simplicity that at a single sequential step I can run multiple experiments in parallel. Even with simplistic approach above, I seem to be able to achieve speedups proportional to what is shown in the literature when I run multiple experiments in parallel. Consider example of optimizing 2 parameters of decision tree, averaged over 64 repeats (green: 1 experiment / step, blue 2, red: 4):

figure_1-7

You can reproduce these results with `/examples/skopt_parallel_example.py`.

I have more examples with similar results, for more dimensions (up to 11), other surrogates, other toy and “real” optimization problems etc. I also tried a bit more complicated approaches for obtaining points to evaluate in parallel, but they did not perform much better.

In light of this, I propose to make a minimalistic ipynb example describing such approach. While I cannot make theoretical argument that it should always yield speedup (probably there are counterexamples), this seems to be useful in practice and at least running multiple experiments instead of one should not do harm. Let me know if you would be interested or if someone is already working on something like this.


iaroslav-ai:
Uhm, I really need to learn git - there seem to be some commits that should have not been there (I talk about bench_ml.py), sorry for that! If someone could suggest some command magic that could solve this I would deeply appreciate it.

iaroslav-ai:
I am not certain why I see my commits from #339 . I created a new branch "parallel_example" from "master" branch of https://github.com/iaroslav-ai/scikit-optimize (after #339 was merged) , added there extra commits and then created this pull request. While benchmarks/bench_ml.py are the same in pull request and current repository, the comparison tool shows somehow that the code lines were added. I will try to figure out what is going on.

dhirschfeld:
Not sure if this will help, but maybe worth giving it a go:
```
git branch parallel_example_backup    # create a backup of your current branch, just in case
git checkout master
git fetch upstream
git reset --hard upstream/master      # ensure your master is up to date
git checkout parallel_example
git rebase master                     # rebase on top of updated master
```

dhirschfeld:
I'm just giving `scikit-optimze` a go for the first time. It seems great but I would really benefit from any way to parallelise it, hence I ended up here!

As a user of [`distributed`](https://distributed.readthedocs.io/en/latest) I was hoping there was an interface to that library a la [`dask-searchcv`](http://dask-searchcv.readthedocs.io/en/latest/). I didn't expect that there would be but I think there could be a lot of benefit in it similar to how `dask-searchcv` enables scaling out grid searches to clusters of machines.

betatim:
I'll take a look at your code when you fix the branch :)

I think one way to obtain several points to evaluate in parallel is to:

1. `ask` for a point `x1`
1. `tell` that point using the estimate of the surrogate as value of the objective
1. `ask` for a second point `x2`

now you have two points to evaluate. The thing to look out for is that once you know the actual value of the objective at `x1` you need to "untell" `x1` and `tell` it with the actual value of the objective you observed.

Being able to evaluate several points in parallel would be a nice addition.

---

Let's discuss dask distributed in a different issue maybe.

iaroslav-ai:
@dhirschfeld thanks for the suggestion! Will try it soon.

iaroslav-ai:
@betatim It is exactly what I tried albeit that in step 2 instead of surrogate estimate I tried using random value as a substitute for actual objective, as well as mean of all known objective values. This gives results quite similar to the ones that are depicted on the figure. Maybe one could do better with surrogate estimates as a substitute for the objective. 
Will try to fix the branch asap :)

iaroslav-ai:
Ok, unnecessary commits are no more. I realized however that there was a bug (which does not seem to change the results that I am getting though) and the code needs to be simplified a bit. I will post update here when I will make the code a bit more readable.

betatim:
We should think about how to add this to `Optimizer` so that as a user one can just call `opt.ask(n_points=3)` and be given three points to evaluate the objective on. Getting the API right will take a bit of work.

To me it is clear that this is a useful feature to have, so the hard work is implementing a good strategy and adding as little code as necessary to the `Optimizer` class (I think this is the right place to add this).

dhirschfeld:
Regarding distributed I'll just note that if this PR makes use of joblib then you can use distributed as a [backend for joblib](http://distributed.readthedocs.io/en/latest/joblib.html)

glouppe:
Just pinching in.

Shouldn't this be closely tied to the acquisition function? The current acquisition functions we have (EI, PI, LCB) do not lend themselves to the query of more than one point. If you `ask` several times the optimizer, these calls should all return identical values, possibly with some variance, but this only highlights issues we have with the optimization the acquisition function.

I think this is a great feature to have, but we should really investigate the literature before implementing strategies we are not sure about.

Entropy search for example (see #304) naturally extends to more than one query point: we ask for the N query points that together would reduce the most entropy of the minimizer.

betatim:
Starting off the literature search: https://hal.archives-ouvertes.fr/hal-00732512/document describes an algorithm that uses the "constant liar" where (simplified) you `ask` for one point, `tell` it with some lie, `ask` for another until you have `q` points to evaluate.

https://arxiv.org/abs/1602.05149 discusses q-EI which is EI adopted to `q` points.

iaroslav-ai:
It seems pretty easy to implement Constant Liar strategy with a wrapper around an instance of Optimizer class. I have that in the `examples/constant_liar.py`, together with example, seems fairly minimalist :) You can also use `examples/constant_liar_history.py` for comparison of results with different numbers of processes. 

So far it seems to be consistently beneficial to run things in parallel even with this simplistic strategy. This implementation could be used for further development, or as a baseline for comparison with more dedicated approaches.

glouppe:
The liar strategy seems like something simple enough to consider as a baseline, and also to design a good and simple API. What do others think?

iaroslav-ai:
Sorry that I am maybe rushing a bit before everyone has a say, but I extended the ask() method of Optimizer in this PR so that you can specify n_jobs (like @betatim suggested) and it will return a list of points using "Constant Liar" strategy. Long story short, ask() is now:

```
def ask(self, n_jobs=None, strategy="CL"):
    if n_jobs is None:
        return self._ask()
    opt = copy.deepcopy(self)
    opt.rng = self.rng  # set random generator to not generate same points
    x = []
    for i in range(n_jobs): # generate points
        x.append(opt.ask())
        y_lie = np.min(opt.yi) if len(opt.yi) > 0 else 0.0  # CL-min lie
        opt.tell(x[-1], y_lie)  # lie to the optimizer
        self._n_random_starts = max(0, self._n_random_starts - 1)
    return x
```

where _ask is an old ask() method. The argument `strategy` is a dummy argument for now. 
The minimal example with new Optimizer would be:

```
import numpy as np
from multiprocessing.pool import ThreadPool
from skopt.space import Real
from skopt import Optimizer

pool = ThreadPool()
optimizer = Optimizer(
    base_estimator=GaussianProcessRegressor(),
    dimensions=[Real(-3.0, 3.0) for i in range(10)]
)

n_jobs, n_steps, Y = 4, 20, []

for i in range(n_steps):
    x = optimizer.ask(n_jobs)
    y = pool.map(lambda x: np.sum(np.array(x) ** 2), x)
    optimizer.tell(x, y)
    Y.extend(y) # keep objectives history here
    print min(Y)

```

This seems to work fine, I get decent improvement when I increase n_jobs. You can see for yourself by running `examples/constant_liar_test.py` (might take a minute or two)
It took ~5 mins to implement so if we go with something else its not big deal :) But this could be used already for tests etc.

iaroslav-ai:
Hmm, I will rename later `n_jobs` to something else, as it is a bit confusing. For now I will keep it here to keep the code in comments and in the PR consistent.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                 "Expected acq_optimizer to be 'lbfgs' or 'sampling', "
                 "got %s" % acq_optimizer)
 
-    def ask(self):
+    def ask(self, n_jobs=None, strategy="CL"):
```
betatim: +1 for renaming `n_jobs`. Maybe `n_samples` or `n_points`?

betatim:
API wise some things that I thought of that we should take care of:

* what should happen if the user asks for several points but then `tell()`s a point that wasn't in that set?
* should we mark points that were added with a lie instead of a real observation?
* should the constant liar find the best `y` amongst those already evaluated or should it try and minimize the surrogate (connected a little to #335)?

So I think my biggest question is about what should the state of the optimizer be after I `ask` for several points? What should the outcome be of `X1 = opt.ask(n_points=3); X2 = opt.ask(n_points=3)`? Should it equal `X3= opt.ask(n_points=6)`? If you ask for several points in one go, should you be allowed to tell only a subset of them or even a point that isn't part of the ones you got from ask and if so what do we do with the lies?

---

We should move the example to a notebook like the other ones so that it is rendered as part of the webpage. Also, we should move the "testing" example to a unit test :)

iaroslav-ai:
In my view, we can separate everything in two options wrt api behaviour. One with remembering the lie points in between `ask()` calls, and another one without. 

**tl;dr:** So far it seems that the “remember” option is harder to implement and there seem to be more opportunities for bugs on user side (eg forgetting to remove lie points), so I am leaning more towards the option of not remembering the lie points in between calls to `ask`. The “remember” option could be implemented if users ask for it or if it performs to be better than the more simple options on the benchmarks.

If we want to remember points, then one needs to keep track of which points are lie or not, and be able to handle situations where some experiment failed, and thus a lie point needs to be removed. One idea would be if y=None is provided into `tell` method which means that that the corresponding point should be removed. Also it might happen that user looses track of what points are actually failed to evaluate, and then would require method like `clear()` which removes all the lie points.
I think it is important to carefully keep track of lie points in “remember” case, as otherwise “forgotten” lie points can mess up variance function. Worst case example would be lie point which was by [un]lucky coincidence placed close to a global optimum, which was assigned some large value. 

In contrast to this, one could discard lie points outside `ask()` method (eg current implementation), which removes the need for tracking the lie points, and makes it easier for user if the experiment has failed - the user simply does not report the results of the experiment. The usage would be to ask for n points to evaluate, evaluate them all in parallel, tell them to the optimization algorithm, ask for next n points, etc. Waiting for all points to finish evaluating and be reported seems important in order to make sure that same point is not evaluated twice. With this approach however the evaluation time for n points is determined by evaluation time of the “slowest” point, which could be inefficient given that evaluation times for different points might differ exponentially. One could be more efficient with “remember” method, where one could run new experiment as soon as resources to do so become available (eg worker thread finishes evaluating a point.)

To see how two approaches compare, I came up with a code [here](https://gist.github.com/iaroslav-ai/93aa78385f55a3bcee8b1dd7c46a2215). There `batch_computation` variable switches between two outlined options. Interestingly, while more points are evaluated per unit of time with “remember” approach due to better use of resources, it does not seem to lead to results as good as with waiting for all n points to finish evaluating (on the toy examples that I tried though). 

--- 

I think the same goes for what kind of lie should be used - benchmarks should determine that :)
Indeed there is a bit of a mess right now with what is where. Will fix it soon. 
P.S. I like it how you use horizontal rules, I will do that too :)

examples/constant_liar.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import numpy as np
+from multiprocessing.pool import ThreadPool
+from skopt.space import Real
+from skopt.learning import GaussianProcessRegressor
+from skopt import Optimizer
+
+# ThreadPool is used for parallel computations
+pool = ThreadPool()
```
betatim: (thinking about the rest so just a small comment) can we ditch this as with python this won't actually run in parallel and for the example it should be ok to illustrate the idea of getting several points, evaluating them, then telling them without having to add actual parallel execution. (or we should use the process pool)

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+        """
+        if n_points is None:
+            return self._ask()
+
+        opt = copy.deepcopy(self)
+        opt.rng = self.rng  # set random generator to not generate same points
+        x = []
+        for i in range(n_points):
+            x.append(opt.ask())
+            y_lie = np.min(opt.yi) if len(opt.yi) > 0 else 0.0  # CL-min lie
```
MechCoder: This should be a parameter defined by the user.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                 "Expected acq_optimizer to be 'lbfgs' or 'sampling', "
                 "got %s" % acq_optimizer)
 
-    def ask(self):
+    def ask(self, n_points=None, strategy="CL"):
```
MechCoder: I suppose `n_points=None` should be `n_points=1`?

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+            Is ignored if n_jobs = None.
+        """
+        if n_points is None:
+            return self._ask()
+
+        opt = copy.deepcopy(self)
+        opt.rng = self.rng  # set random generator to not generate same points
+        x = []
+        for i in range(n_points):
+            x.append(opt.ask())
```
MechCoder: Please use `opt._ask()` directly.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+            y_lie = np.min(opt.yi) if len(opt.yi) > 0 else 0.0  # CL-min lie
+            opt.tell(x[-1], y_lie)  # lie to the optimizer
+            self._n_random_starts = max(0, self._n_random_starts - 1)
```
MechCoder: Why are you doing this here, every call to `ask` reduces `n_random_starts` right?

examples/constant_liar.py
```diff
@@ -0,0 +1,31 @@
...
+
+# configure number of threads to be used in parallel, and overall # of computations
+n_points, n_steps, Y = 4, 20, []
+
+for i in range(n_steps):
+    x = optimizer.ask(n_points)
+    # evaluate n_points in parallel
+    y = pool.map(lambda x: np.sum(np.array(x) ** 2), x)
+    # tell points and corresponding objectives to the optimizer
+    optimizer.tell(x, y)
```
MechCoder: So, the paper just describes getting `q` no. of values. While feeding these values back, note that every `x` has two function values which can be interpreted as variance in `f(x)`.  Should we reset the state of the `GP` after `n_points` are asked for or when the points told to the GP are told internally?

MechCoder:
Thanks for starting this up. It seems pretty useful!

MechCoder:
With respect to the API, I suppose we should do the following:

1. `asking` for points repeatedly without telling should provide the same points if the random state is fixed. (current behavour in master).
2. This means if multiple points are asked for at `ask` time, the state of the surrogate model should be reset to where it was before the `ask`ing.

examples/constant_liar.py
MechCoder: (Sorry skipped the long discussion, I've summarized my understanding below)

skopt/optimizer/optimizer.py
iaroslav-ai: If `n_points` is set, I return a list of points, and single point if it is none. I think this might be better for users who might want to experiment with different number of points (including n_points=1), as then they do not need to account for special case when `n_points=1`.

skopt/optimizer/optimizer.py
iaroslav-ai: You mean the type of lie that is fed to the surrogate, right?

skopt/optimizer/optimizer.py
iaroslav-ai: I call `ask` on copy of surrogate, not the surrogate itself. Thus I need to reduce this variable.

iaroslav-ai:
I agree with @MechCoder . It just seems like not keeping the lie points in between `ask` is easier to implement and it offers less possibilities for bugs on user side.

examples/constant_liar.py
iaroslav-ai: I drop the ThreadPool, and use joblib instead, I think this way it is also consistent with the rest of the code in repository. I think it is  good to have example with actual parallel computation going on, as then all user needs to do is substitute the function of their own to have things run in parallel.

skopt/optimizer/optimizer.py
MechCoder: ok, thanks!

skopt/optimizer/optimizer.py
MechCoder: yes.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+            Otherwise a list of points to evaluate is returned of size n_jobs.
+        * `strategy` [string]:
+            Strategy to use in order to query multiple points to evaluate.
+            Is ignored if n_jobs = None.
+        """
+        if n_points is None:
+            return self._ask()
+
+        opt = copy.deepcopy(self)
+        opt.rng = self.rng  # set random generator to not generate same points
```
MechCoder: You need not have to make a copy here. `rng` is a RandomState and not an integer. It will produce the same sequence of random numbers across runs and not the same random numbers is a single run.

MechCoder:
Since `tell` modifies the state of `Xi`, `yi` and `models`, one way to handle this would be something like this.

```python
def _tell(self, x, y):
    # Optimize
    return est, next_x

def tell(self, x, y):
    est, self._next_x = self._tell(x, y)
    self.models.append(est)
```

and then call `_tell` multiple times.

MechCoder:
Also, can you rebase with master?

MechCoder:
Once we get the API cleared, it would be great to reproduce Table 1 of  http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=D48CA9F3BDBACE2E306F2E60257B0BD6?doi=10.1.1.332.879&rep=rep1&type=pdf  . Do not hesitate to ping if no-one replies for a few days, sometimes I look at github comments through mails and I forget to reply.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,32 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                 "Expected acq_optimizer to be 'lbfgs' or 'sampling', "
                 "got %s" % acq_optimizer)
 
-    def ask(self):
+    def ask(self, n_points=None, strategy="CL"):
+        """Query point or multiple points to evaluate.
+
+        * `n_jobs` [int or None]:
```
betatim: n_jobs -> n_points

betatim: and we should update the text describing what this does. I'll think about a suggestions.

betatim:
I am still feeling a bit uncomfortable with the "copy the optimizer" approach :-/ It somehow doesn't feel "right".

Here are some thoughts:
* `x1 = opt.ask(n_points=3); x2 = opt.ask(n_points=3); assert x1 == x2` should work
* calling `ask` should leave the state of the optimizer unchanged (for example it should not reduce the number of function evaluations left)
* I can tell any point I like, even one I didn't `ask` for before, this will update the state of the optimizer

Here a suggestion, not sure it is better but would be good to hear some feedback from you guys:

```
class Optimizer:
  def _remove(self, n=1):
    """Remove last n points from the history. Restore the state of the optimizer to what it was."""
    self.X = self.X[:-n]
    self.models = self.models[:-n]
    ...etc..
    # recompute the next point that `ask` returns?

  def ask(self, n_points=None):
    if n_points is not None:
      xs = []
      for n in range(n_points):
        xs.append(self._ask())
      self._remove(n_points)
      return xs
    else:
      return self._ask()
```
WDYT?

betatim:
Having it written down like that ... maybe it does make sense to "fork" the optimizer when asking for multiple points as it should make it easier to do all the "rewinding" (because it is automatic).

betatim:
Maybe a compromise that satisfies my suspicion of the `copy` module is that we construct a new `Optimizer` instance from the existing one instead of copying it.

This has the advantage of making a copy (no counters to reset, repeated `ask` calls return the same thing) but doesn't have the downside of requiring twice as much memory (because we deep copy all the history of the existing `Optimizer`). I think this is my favourite solution so far.

We should also encode some of the "invariants" I listed above as unit tests to make sure they are actually true.

---

Not sure why I don't like `copy` ... it just feels like you shouldn't ever have to use it in pythonic code...but yeah that is more of a feeling/opinion.

MechCoder:
I removed my previous comment because I realized how stupid it was. I don't like the "deep-copying" the Optimizer solution either.

My preference would be the solution mentioned here (https://github.com/scikit-optimize/scikit-optimize/pull/343#issuecomment-299386343). We could maybe rename the `_remove` to `reset` to make it seem better.

iaroslav-ai:
@betatim I agree with you, it just seems easier to make a copy of the optimizer class, so that you do not need to worry about someone adding something that also needs to be copied, and then the code needs to be updated every time :)
A possible disadvantage that I could imagine with copy is that it might be slow. Not that it really matters in the contexts of intended use of scikit - learn, but non the less it might be annoying to have slow tests.
Indeed it seems as if constructing new optimizer is the better way to go. This also does not copy the .models, which should make things [at least a bit] faster compared to deepcopy. 
I was also thinking about having a function in an Optimizer class which takes as input list of explored points and returns a new point to explore based on the provided list of points. Constructing new optimizer kinda does this anyways, but then maybe one does not need to worry about copying the parameters from one optimizer to another one (it is probably easy, but I did not look into it).

MechCoder:
What are the disadvantages of the reset approach?

MechCoder:
I might be wrong here but how do you construct a new Optimizer instance without explicitly copying the history of points and at-least the last model fit?

Copying the Optimizer in fhe case of a random forest, copies `O(n_calls X n_trees)` memory.

dhirschfeld:
It seems to me that implementing your own shallow copy would work and would remove any need to carefully manage state (adding then removing points):
```python
In [69]: class obj:
    ...:     def __repr__(self):
    ...:         return "obj({})".format(id(self))

In [70]: class Optimizer:
    ...:     X = None
    ...:     def __init__(self, X):
    ...:         self.X = X
    ...:     def copy(self):
    ...:         """Returns a shallow copy of the class"""
    ...:         return Optimizer(list(self.X))
    ...:     def __repr__(self):
    ...:         return "Optimizer(X={!r})".format(self.X)

In [71]: c = Optimizer([obj() for _ in range(3)])

In [72]: cc = c.copy()

In [73]: cc.X.append(obj())

In [74]: c
Out[74]: Optimizer(X=[obj(2028073363104), obj(2028090527928), obj(2028090529608)])

In [75]: cc
Out[75]: Optimizer(X=[obj(2028073363104), obj(2028090527928), obj(2028090529608), obj(2028090593688)])
```
...as can be seen above appending on to the list `X` doesn't affect the original `Optimizer` class but the shallow copy contains references to the original `obj`s rather than copies (the ids are the same as in the original list)

iaroslav-ai:
@MechCoder it just seems to make it easier with a copy to keep track of the state and updating the code when some new things are added, that is all.
I agree that it would also be good to reproduce results in the literature to make sure that everything works fine. 
@dhirschfeld that is interesting actually. Indeed, this should not take much space and I would expect it to be quite fast also.
---
I will try to do something that does not involve deepcopying stuff.

skopt/optimizer/optimizer.py
iaroslav-ai: Hmm, sorry for confusing response, by surrogate I mean optimize.

MechCoder:
Taking a cue from @dhirschfeld , here is a minimal suggestion.

```python
class Optimizer(object):

     def copy(self):
         optimizer = Optimizer(
             dimensions=self.space, base_estimator=self.base_estimator,
          ...)
          optimizer.tell(self.Xi, self.yi)
          return optimizer
```

This will not work for `acq_func="gp_hedge"` which involves some more hackery such as copying the gains.

betatim:
:tada: @MechCoder something like what you posted is what I had in mind. Basically doing our own "copy" so that we don't copy all of the models (and save some memory).

Making sure we get copies where needed and elsewhere references will be the part where bugs creep in. I think we definitely need to copy `Xi`, `yi` and `rng`. Most of the other things are either strings (immutable by default) or we can use references.

Happy weekend y'all!

iaroslav-ai:
Sorry, could not comment earlier.
I use now these two lines instead of deepcopying: 

```
opt = copy.copy(self)
opt.__dict__ = {k:copy.copy(v) for k,v in self.__dict__.items()}
```
First one creates shallow copy of optimizer, and second one copies all fields from self into new copy of class as shallow copies. 
This seems to do exactly what we want: it creates a copy of optimizer, where fields which are arrays contain references to points in original optimizer. I just checked in the debugger - adding new points to opt does not adds new points to self.
I compared this approach to deepcopy using trees surrogate. On my computer, deepcopy gradually takes more and more time as more points are explored, starting from 0.05 s and ending in 0.041 s for 200 points. In contrast to this, new approach results in a constant value of 0.0002 seconds. For comparison, one ask with n_points=4 takes at least 0.5 s.
This looks like a good approach as then one does not need to define explicitly the copy() method. Do you think this could work?
@betatim thank you, you too and everyone else have a splendid weekend :)

betatim:
I would vote for @MechCoder's suggestion of explicitly creating the forked optimizer instance instead of reaching into the internals of `opt.__dict__` and such.

iaroslav-ai:
Do you like it more because you explicitly control what you copy? I am fine with either option. 
@MechCoder what do you say?

MechCoder:
Are you asking rhetorically if I would prefer the solution that I wrote? ;)

It should be better to explicitly creating a copy using as less memory as possible. Such at thing can also be used in a `CheckpointCallback`. The requirement being 

```python
opt1 = Optimizer(...)
opt2 = opt1.copy()
for i in range(10):
    opt1.ask() == opt2.ask()
```
Is there anything else? We can move that to a separate PR if needed.

iaroslav-ai:
I implemented copy in this branch so that I can see if everything works fine together, but feel free to open a new PR if you think something needs to be changed, I will simply rebase then.

iaroslav-ai:
Currently the parallelization strategy is passed as a string parameter to the constructor of optimizer, and number of points is passed to the `ask` method as an argument. Do you think there should be any further changes to the way the api is?

jcrist:
I'm excited to see this work coming along. Once this reaches a usable state, I hope/plan to implement a wrapper around it for performing distributed hyper-parameter optimization in [dask-searchcv](https://github.com/dask/dask-searchcv). Based on the example in this PR, I expect this to be fairly simple to do (hooray for well designed libraries :)). Is there anything I can do to help move things along here?

MechCoder:
Are you a SymPy developer? I contributed a few years back :)

It's just reviewer availability. I'm in the last week of my graduate program so I hope to get to reviewing this soon..

jcrist:
I used to be, but haven't contributed in a while. No rush on my end, was just offering to help in any way needed. Good luck on finishing up grad school :).

glouppe:
> Is there anything I can do to help move things along here?

If you have time, helping with reviewing this PR is a good place to start :)

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +189,74 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+        opt = self.copy()
+
+        # when _n_random_starts > 0, random state does not change
+        # during tell(), but only during ask(). As ask() is called
+        # on a copy of the optimizer only, the random state of original
+        # optimizer does not change during _n_random_starts > 0. If
+        # the random state is not shared with original optimizer,
+        # then same points are generated, which leads to exceptions
+        # with fitting of some models.
+        if self._n_random_starts > 0:
```
MechCoder: I think I understood the problem, but can you give an example just for some clarity. Do you mean that if this line is removed?

```python
opt = Optimizer(n_random_starts=2)
Xi = opt.ask(n_points=2)
opt.tell(Xi, yi)
x = opt.ask()
opt.tell(Xi + x, yi + yi[0])
```

will fail because Xi[0] is the same as x?

MechCoder:
Sorry about the delay @iaroslav-ai !  Hope you are still around. I believe the API is much clearer now. Can you please start adding some tests?

iaroslav-ai:
Hi, I am still around :D Was a bit busy with my stuff also.
Will take a look asap at what tests I can add and at your comments.

skopt/optimizer/optimizer.py
```diff
@@ -94,6 +95,16 @@ class Optimizer(object):
     * `acq_optimizer_kwargs` [dict]:
         Additional arguments to be passed to the acquistion optimizer.
 
+    * `parallel_strategy` [string, default=`"cl_min"`]:
```
iaroslav-ai: It seems that it will give more flexibility to a user, so I feel it is good to have it in ask. Moved it there.

examples/constant_liar.py
```diff
@@ -0,0 +1,32 @@
+"""
```
iaroslav-ai: Yup, will do so.

skopt/optimizer/optimizer.py
```diff
@@ -94,6 +95,16 @@ class Optimizer(object):
...
+        Method to sample multiple points for evaluation in parallel. This parameter
+        is ignored if n_jobs = None. Supported options are `"cl_min"`, `"cl_mean"`
```
iaroslav-ai: Ooops my bad :) Fixed.

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +189,74 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+
+        * `n_jobs` [int or None]:
+            Number of function evaluations that can be done in parallel.
+            If the value is None, a single point to evaluate is returned.
+            Otherwise a list of points to evaluate is returned of size n_jobs.
+        """
+        if n_points is None:
+            return self._ask()
+
+        opt = self.copy()
```
iaroslav-ai: Yes, I added a comment there.

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +189,74 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+        # with fitting of some models.
+        if self._n_random_starts > 0:
+            opt.rng = self.rng
+
+        X = []
+        for i in range(n_points):
+            x = opt.ask()
+            X.append(x)
+            if self.parallel_strategy == "cl_min":
+                y_lie = np.min(opt.yi) if len(opt.yi) > 0 else 0.0  # CL-min lie
```
iaroslav-ai: I see what you mean. I wander is this is an issue though; You only use 0.0 lie if you do not know any objective value. As soon as you have any values, the opt.yi should have at least some values, and thus in your example the lie will be negative. Also the lie is not preserved after the call to ask. Kinda seems like there should be no issue - you only use 0.0 lie initially in order to make sure that you select all points with large variance (but you do not know anything yet about objective). Anyone has any ideas on this?

skopt/optimizer/optimizer.py
iaroslav-ai: Here is a fairly self contained example, which prints (#random restarts, generated points):
```python
from copy import deepcopy

optimizer = Optimizer(
    base_estimator=GaussianProcessRegressor(),
    dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
    acq_optimizer='sampling'
)

n_points, n_steps, Y = 2, 10, []

for i in range(n_steps):
    opt = deepcopy(optimizer)
    #opt.rng = optimizer.rng # if you uncomment this, the different points generated
    X = []

    for j in range(n_points):
        x = opt.ask()
        X.append(x)
        lie = np.min(opt.yi) if len(opt.yi) > 0 else 0.0
        opt.tell(x, lie)

    print(optimizer._n_random_starts, X)
    optimizer.tell(X, [branin(v) for v in X])
    optimizer._n_random_starts = max(0, optimizer._n_random_starts - len(X))

```

In the pull request I duplicate self.rng in copy() method of Optimizer, so that at least
when _n_random_starts == 0, multiple calls to ask() result in same set of points.
However as I copy the random number generator, the random points
for _n_random_starts > 0 are the same even if you `tell` them to
the Optimizer (as is shown in the example above). Thus I added the line which takes a reference to the original number generator when  _n_random_starts > 0.

iaroslav-ai:
Not sure what else to add - please review.

iaroslav-ai:
Hmm it feels that this PR will be in mergable state soon. I change the name accordingly.

skopt/optimizer/optimizer.py
MechCoder: I think I meant the same thing. You can further strip your example to this to convey your point.

```python
optimizer = Optimizer(
    base_estimator=GaussianProcessRegressor(),
    dimensions=[(-5.0, 10.0), (0.0, 15.0)])

for i in range(10):
    opt = optimizer.copy()
    print(opt.ask())
```

MechCoder:
I'm afraid of sharing a common random state between the optimizer and its copy.

Since the problem of generating common points among the `n_points` number of optimizers seem to happen only when `n_random_starts` is not exhausted, what do you think about generating `n_random_starts` number of points at once during the first call to `ask`?

MechCoder:
I was suggesting something similar to this.

```python
def __init__(self, ...):
    self._n_points = 0

def ask(self, n_points...)
    if self._n_points = 0:
        self._X_rand = self.space.rvs(random_state=self.rng, size=self.n_random_starts)

    if n_points is None:
        if self._n_random_starts > 0:
            self._n_points += 1
            return self._X_rand[self._n_points - 1]
        else:
            return self._next_x
    else:
      Xi = []
       if self._n_random_starts > 0:
           start = self._n_points + 1
           end = start + np.min(n_points, self._n_random_starts)
           Xi = self._n_random_starts[start: end]
           self._n_random_starts -= len(Xi)
           rem = n_points - len(Xi)
           self._n_points += len(Xi)
      if rem == 0:
          return Xi

     opt = self.copy()
     if Xi:   
         opt.tell(Xi, lie)
     for i in range(rem):
         x = opt.ask()
         Xi.append(x)
         opt.tell(x, lie)
    return Xi
```

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +183,95 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+
+    def ask(self, n_points=None, parallel_strategy="cl_min"):
+        """Query point or multiple points to evaluate.
+
+        * `n_points` [int or None]:
+            Number of function evaluations that can be done in parallel.
+            If the value is None, a single point to evaluate is returned.
+            Otherwise a list of points to evaluate is returned of size n_jobs.
+
+        * `parallel_strategy` [string, default=`"cl_min"`]:
```
MechCoder: `parallel_strategy` to just strategy? 

I understand that you mean parallel as in "parallel in evaluation", but it can also mean that the strategy is parallel in nature but in reality it is sequential

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +183,95 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+
+
+
+    def ask(self, n_points=None, parallel_strategy="cl_min"):
+        """Query point or multiple points to evaluate.
+
+        * `n_points` [int or None]:
+            Number of function evaluations that can be done in parallel.
+            If the value is None, a single point to evaluate is returned.
+            Otherwise a list of points to evaluate is returned of size n_jobs.
```
MechCoder: n_jobs -> n_points

MechCoder:
The code written by me above looks unnecessarily complicated. Another simple way to fix this, is to just use a different random state of the copied Optimizer during initialization.

I'm +1 to that approach but in that case it might make sense to rename `copy` to something else.

skopt/optimizer/optimizer.py
iaroslav-ai: Oops yet another typo of mine sorry :) I will double check the descriptions for those

skopt/optimizer/optimizer.py
iaroslav-ai: Haha I see what you mean :D I will rename it to strategy.

iaroslav-ai:
I really like the approach of generating the _n_random_starts points during first call, feels like it could resolve the issues. Will give it a try.

iaroslav-ai:
Actually using a different random state should also fix everything in a rather simple fashion. 
Some time ago there was a discussion that two consecutive calls to `ask` should result in same generated points (or sets of points for n_points not None). This is not true if we use a new random state every time the optimizer is copied. Do you think this is a big issue? If not, then I would think we should go with just generating new random state. But if it is,  With cached  _n_random_starts points as I would imagine it right now the same points should be returned with consecutive calls to `ask`.

MechCoder:
Let's just go with caching the results then.

iaroslav-ai:
Great. Should be ready in some minutes  :)

iaroslav-ai:
The `copy` method now takes `copy_random_state` parameter, and I introduced caching for n_points not None (for the other case it is not necessary). Let me know what you think.

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+def test_constant_liar_runs():
+    optimizer = Optimizer(
+        base_estimator=ExtraTreesRegressor(),
+        dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
+        acq_optimizer='sampling'
+    )
+
+    n_points, n_steps, Y = 2, 13, []
+
+    for i in range(n_steps):
```
betatim: what does `n_steps` do? It seems like it is used to iterate through the different strategies for parallel evaluation? For that we should use [`@parametrize`](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions)

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+        strategy = optimizer.par_strats[i % len(optimizer.par_strats)]
+        x = optimizer.ask(n_points, strategy)
```
betatim: we should probably test that `x` contains `n_points` points :)

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+        optimizer.tell(x, [branin(v) for v in x])
+
```
betatim: `pep8` wants an extra blank line (there are a few more `pep8` things to fix up too)

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+        x = optimizer.ask(n_points, strategy)
+        optimizer.tell(x, [branin(v) for v in x])
+        distances = pdist(x)
+        if any(distances < tolerance):
+            raise ValueError("Same points were generated!")
+
+
+def test_same_set_of_points_ask():
+    """For n_points not None, tests whether two consecutive calls to ask
+    return the same sets of points."""
```
betatim: I think we should stick to using a comment to explain what the test is about instead of also having docstrings. Not sure why we adopted this in the rest of the code base but now that we have it ... :)

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+
+    optimizer = Optimizer(
+        base_estimator=ExtraTreesRegressor(),
+        dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
+        acq_optimizer='sampling'
+    )
+
+    n_points, n_steps, Y = 8, 5, []
```
betatim: `Y` is not used

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+
+    for i in range(n_steps):
+        strategy = optimizer.par_strats[i % len(optimizer.par_strats)]
+        xa = optimizer.ask(n_points, strategy)
+        xb = optimizer.ask(n_points, strategy)
+
+        optimizer.tell(xa, [branin(v) for v in xa])
+
+        if not (str(xa) == str(xb)):
```
betatim: why not compare the return values directly (or use `assert_equal`) instead of turning them into strings?

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+            raise ValueError("Different sets of points were generated for two consecutive calls to ask.")
```
betatim: Generally in tests I wouldn't raise an exception but use `assert` statements.

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+    optimizer = Optimizer(
+        base_estimator=ExtraTreesRegressor(),
+        dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
+        acq_optimizer='sampling'
+    )
+
+    n_points, n_steps, Y = 8, 5, []
+
+    for i in range(n_steps):
+        strategy = optimizer.par_strats[i % len(optimizer.par_strats)]
```
betatim: as above: use `@parametrize`

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+               for a point, and the point is told to the copy of optimizer with
+               some fake objective (lie), the next point is asked from copy,
+               it is also told to the copy with fake objective and so on.
+               The type of lie defines different flavours of `cl_x` strategies.
+
+        """
+        if n_points is None:
+            return self._ask()
+
+        if not strategy in self.par_strats:
```
betatim: `if not x in y:` -> `if x not in y:` reads nicer and is what python does for you internally anyway :)

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+               https://hal.archives-ouvertes.fr/hal-00732512/document
+
+               With this strategy a copy of optimizer is created, which is then asked
+               for a point, and the point is told to the copy of optimizer with
+               some fake objective (lie), the next point is asked from copy,
+               it is also told to the copy with fake objective and so on.
+               The type of lie defines different flavours of `cl_x` strategies.
+
+        """
+        if n_points is None:
```
betatim: What should the right behaviour be if `n_points == 1`? I think it should return a list with just one value, which makes it different from `None`. I think we agree but wanted to double check

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+
+        if len(self.Xi) > 0:
+            optimizer.tell(self.Xi, self.yi)
+
+        return optimizer
+
+    def ask(self, n_points=None, strategy="cl_min"):
+        """Query point or multiple points at which objective should be evaluated.
+
+        * `n_points` [int or None]:
```
betatim: needs a `default=None`

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+
+        if copy_random_state:
+            params['random_state'] = copy.copy(self.rng)
+
+        optimizer = Optimizer(**params)
+
+        if hasattr(self, "gains_"):
+            optimizer.gains_ = np.copy(self.gains_)
+
+        if len(self.Xi) > 0:
```
betatim: In python we can write `if some_list` directly instead of checking if the length is larger than zero. (here are a few more of these in the PR)

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                              "'sampling', got {0}".format(acq_optimizer))
         self.acq_optimizer = acq_optimizer
 
-    def ask(self):
+    def copy(self, copy_random_state=True):
+        """Creates a shallow copy of optimizer."""
```
betatim: More docs :) Also "Create ..." instead of "Creates ..." https://www.python.org/dev/peps/pep-0257/ has a few more suggestions on wording (even as a native speaker I keep checking that I follow it)

skopt/optimizer/optimizer.py
```diff
@@ -152,6 +155,13 @@ def __init__(self, dimensions, base_estimator,
 
         self.n_jobs = n_jobs
 
+        # list of all strategies for parallelization. Used internally for testing.
+        # See description of `ask` method to see what parallel strategy means.
+        self.par_strats = ["cl_min", "cl_mean", "cl_max"]
```
betatim: not a big fan of the name and we only use it once in `Optimizer` so can we define this as a tuple in the check that is done instead?

betatim:
Sorry for all those little comments. I will think about the random state copying during :zzz: :sleeping:

examples/parallel-optimization.ipynb
```diff
@@ -0,0 +1,205 @@
...
+    "    \n",
+    "    # plot results of optimization \n",
+    "    plt.plot(range(len(Y)), Y, label=\"points=\" + str(n_jobs))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now lets use the function and see how sequential mode performs against the parallel mode:"
```
MechCoder: well this is kind of cheating because in the parallel there are two calls to the objective function at every call to tell. Do you know if this is the preferred way to compare parallel strategies?

(Or at least you should add a note saying that the assumption is that several jobs are always available and it saves time to suggest points in parallel to different jobs unlike this toy example for the naive reader)

examples/parallel-optimization.ipynb
```diff
@@ -0,0 +1,205 @@
...
+   "source": [
+    "# objective function to minimze\n",
+    "def objective(x):\n",
+    "    return np.sum(np.array(x) ** 2)\n",
+    "\n",
+    "def plot_optimization(n_jobs):\n",
+    "    optimizer = Optimizer(\n",
+    "        base_estimator=ExtraTreesRegressor(),\n",
+    "        dimensions=[Real(-3.0, 3.0) for i in range(3)],\n",
+    "        acq_optimizer='sampling'\n",
```
MechCoder: I believe we should start the same point for both?

examples/parallel-optimization.ipynb
```diff
@@ -0,0 +1,205 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Problem statement\n",
+    "\n",
+    "In many practical cases, you can run evaluations of your objective in parallel, which you would be \n",
+    "great to leverage in order to speed up the black box optimization. \n",
+    " \n",
+    "Think about it like this: if you want to find a perfect mix of ingredients for a perfect cookie, you would rather \n",
```
MechCoder: Another example can be distribute data across different jobs and running computations in parallel..?

MechCoder:
I think that is it from me.

Is there a way to mark a feature as experimental?

iaroslav-ai:
Thanks for a detailed feedback everyone! :) Will take a look soon.

betatim:
> Is there a way to mark a feature as experimental?

I'd be -1 on this. Either we are confident in it or not :) What we could do (eg if we want to hammer out the public facing invariants some more) is not expose the multiple points functionality in `ask()` and only have a `_ask(n_points=...)` for the moment. This way it is clear that the public API doesn't yet support it but keen people can dig into the internals ... but I"d be -0 on this approach as well.

betatim:
Some thoughts on `ask()` providing the same points:

```python
et = ExtraTreesRegressor(random_state=2)
opt = Optimizer([(0, 10.), (0.,10.)], et, acq_optimizer='sampling')
x1 = opt.ask()
x2 = opt.ask()
assert x1 != x2
```

so while we are still in the random sampling phase two calls to `ask()` will result in different points. Once a model has been fit I would expect/hope we always return the same point no matter how often you call `ask()` until the user provides new information via a call to `tell()`. (There is actually a bug somewhere so I can't test it right now)

I think if we call `ask` repeatedly during the random startup phase then we should get new points, once there is a model we should get the same points unless `tell` is called.

I think this means we just pass `rng` into the copy and advance that instance (which is shared with the original). I think this means we can get rid of the `copy_random_state` parameter in the `copy()` method right?

MechCoder:
By getting rid of the parameter `copy_random_state`, you mean replacing it with another parameter that can set the random state right? See (https://github.com/scikit-optimize/scikit-optimize/pull/343#discussion_r119003903)

skopt/tests/test_parallel_cl.py
iaroslav-ai: Thanks for the reference - I switched to the parametrize now. 
n_steps defines a number of calls to ask() method, where at every call n_points are generated, evaluated and fed back to the optimizer.

skopt/tests/test_parallel_cl.py
iaroslav-ai: Yup :) Added this to the test.

skopt/tests/test_parallel_cl.py
iaroslav-ai: Yeah, my `pep8`-fu is not strong as of now. I guess I need to look more into it - I think there are issues with that in other parts of my code.

skopt/tests/test_parallel_cl.py
iaroslav-ai: Yeah, some remnants of the older version of the code. Removed.

skopt/tests/test_parallel_cl.py
iaroslav-ai: That is what I ask myself now also :) Switched to `assert_equal`.

skopt/tests/test_parallel_cl.py
iaroslav-ai: Ok, I removed any kind of explicit raising in the code and switched to `assert`.

skopt/tests/test_parallel_cl.py
iaroslav-ai: Yes, using `@parametrize` now.

skopt/tests/test_parallel_cl.py
iaroslav-ai: pep8 package does not say anything about the test script now - I assume that it is fixed.

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                              "'sampling', got {0}".format(acq_optimizer))
         self.acq_optimizer = acq_optimizer
 
-    def ask(self):
+    def copy(self, copy_random_state=True):
+        """Creates a shallow copy of optimizer."""
+
+        params = {
+            'dimensions':self.space.dimensions,
```
iaroslav-ai: Yup, `pep8` is my current weakness. I will use heavily `pep8` package, so these should go away in nearby future.

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+    n_points, n_steps, Y = 8, 5, []
+    tolerance = 1e-3 # minimum distance at which points are considered to be the same
+
+
+    for i in range(n_steps):
+        strategy = optimizer.par_strats[i % len(optimizer.par_strats)]
+        x = optimizer.ask(n_points, strategy)
+        optimizer.tell(x, [branin(v) for v in x])
+        distances = pdist(x)
+        if any(distances < tolerance):
```
iaroslav-ai: It actually seems that `assert_greater` assumes that constants are provided as arguments that are compared. I use `assert all(distances > tolerance)` now.

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+from skopt import Optimizer
+from skopt.benchmarks import branin
+from scipy.spatial.distance import pdist
+
+
+def test_constant_liar_runs():
+    optimizer = Optimizer(
+        base_estimator=ExtraTreesRegressor(),
+        dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
+        acq_optimizer='sampling'
```
iaroslav-ai: yes, the random state is set now.

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,70 @@
...
+        x = optimizer.ask(n_points, strategy)
+        optimizer.tell(x, [branin(v) for v in x])
+
+def test_all_points_different():
+    # tests whether the parallel optimizer always generates different points to evaluate
+
+    optimizer = Optimizer(
+        base_estimator=ExtraTreesRegressor(),
+        dimensions=[Real(-5.0, 10.0),Real(0.0, 15.0)],
+        acq_optimizer='sampling'
```
iaroslav-ai: Added the random state.

skopt/optimizer/optimizer.py
iaroslav-ai: oops missed that, added now

skopt/optimizer/optimizer.py
iaroslav-ai: Indeed it is nicer. I changed it here and in other parts of the code where I was able to find similar thing.

skopt/optimizer/optimizer.py
iaroslav-ai: Switched to your suggestion.

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
                              "'sampling', got {0}".format(acq_optimizer))
         self.acq_optimizer = acq_optimizer
 
-    def ask(self):
+    def copy(self, copy_random_state=True):
```
iaroslav-ai: I think this is a good idea, switched to random_state parameter

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +186,101 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+            return self._ask()
+
+        if not strategy in self.par_strats:
+            raise ValueError(
+                "Expected parallel_strategy to be one of " + str(self.par_strats) + ", "
+                "got %s" % strategy)
+
+        # Caching the result. If some new parameters are provided to the ask,
+        # the cache is not used.
+        if not self.cache is None:
```
iaroslav-ai: Nice. I use this now.

examples/parallel-optimization.ipynb
iaroslav-ai: Yeah, I do not like how I structured the example and the text as such. I will rewrite taking into account your suggestions.

iaroslav-ai:
@betatim currently the caching of outputs of ask for n_points not None is introduced, which means that  with either random state sharing or not, calling ask twice results in same sets of points (see `test_same_set_of_points_ask` function in test_parallel_cl.py which is passed). Hence right now I do not copy the random state. Everything else seems to work fine with this approach. Let me know if you think this needs to change.

iaroslav-ai:
I made a pass through all your comments. Please take a look and let me know if you wish for any changes.

examples/parallel-optimization.ipynb
```diff
@@ -0,0 +1,129 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "For many practical black box optimization problems expensive objective can be evaluated in parallel at multiple points. This allows to get more objective evaluations per unit of time, which reduces a time necessary to reach good objective values when appropriate optimization algorithms are used, see for example results in [1] and the references therein. \n",
```
MechCoder: nitpick: `a time` to `the time`

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +183,109 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
+        # deletion of points with "lie" objective (the copy of
+        # oiptimizer is simply discarded)
+        opt = self.copy()
+
+        X = []
+        for i in range(n_points):
+            x = opt.ask()
+            X.append(x)
+            if strategy == "cl_min":
+                y_lie = np.min(opt.yi) if opt.yi else 0.0  # CL-min lie
```
MechCoder: Should we just raise an error when `n_random_starts` is lesser than `n_points`. Where else is `opt.yi` an empty list?

skopt/optimizer/optimizer.py
```diff
@@ -152,6 +155,10 @@ def __init__(self, dimensions, base_estimator,
 
         self.n_jobs = n_jobs
 
+        # the cache of responses of ask. For n_points not None, this ensures
+        # that multiple calls to ask return same sets of points.
+        self.cache = {}
```
MechCoder: Note that the behaviour between `n_points=None` and `n_points=x` is different when `n_random_starts` is not exhausted and `ask` is called multiple times. Also can you make it `self.cache_`

MechCoder: (Can you also add that this is reset at `tell` time?)

MechCoder:
That's it from me. LGTM!

examples/parallel-optimization.ipynb
iaroslav-ai: Yup fixed

examples/parallel-optimization.ipynb
```diff
@@ -0,0 +1,129 @@
...
+    " \n",
+    " \n",
+    "One such example task is a selection of number and type of neurons of a neural network used for some machine learning problem. For such task, multiple neural networks with different combinations of number of neurons and type can be evaluated at the same time in parallel on different cpu cores / computational nodes. \n",
```
iaroslav-ai: Right, a bit confusing. Rewritten with activation instead of neuron type.

skopt/optimizer/optimizer.py
iaroslav-ai: Yes, I added all points to the comment and renamed `cache_`.

skopt/optimizer/optimizer.py
iaroslav-ai: Hmm it might happen that the `n_points` could be something large (eg 64 if you run computations on a 64 core machine), so I  think situation when `n_random_starts` < `n_points` should not raise the error. 
Normally `yi` is empty when no points were generated yet. Do you suggest to make a distinction between the case when `n_random_starts` has been exhaused or not? Right now I do not make this distinction explicitly, but allow this to happen in `_ask` method, hence I am also checking if the list `yi` is empty or not.

iaroslav-ai:
@MechCoder has finished reviewing 
I pushed the changes upstream. 
The code might be ready for merging... 
What say you of it, @betatim? :) 
 
P.S. Sorry, just had an inspiration to ping you in an unusual way :) But I do not mean to rush things - take a look when you are around.

skopt/optimizer/optimizer.py
```diff
@@ -176,7 +185,109 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
...
         self.acq_optimizer = acq_optimizer
 
-    def ask(self):
+    def copy(self, random_state=None):
+        """Create a shallow copy of an instance of the optimizer.
+
+        Parameters
+        ----------
+        * `random_state` [int, RandomState instance, or None (default)]:
+        Set random state of the copy of optimizer.
```
betatim: needs a few SPACE characters

skopt/tests/test_parallel_cl.py
```diff
@@ -0,0 +1,121 @@
...
+from scipy.spatial.distance import pdist
+import pytest
+
+# list of all strategies for parallelization
+supported_strategies = ["cl_min", "cl_mean", "cl_max"]
+# Extract available surrogates, so that new ones are used automatically
+available_surrogates = [
+    getattr(sol, name) for name in sol.__all__
+    if "GradientBoostingQuantileRegressor" not in name
+]  # excluding the GradientBoostingQuantileRegressor, will open issue later
```
betatim: -> new issue

betatim:
What a mega PR!

betatim:
Thank you everyone for the patience, many rounds of comments and many rounds of addressing comments! I recommend 🍰 for breakfast, because it is the breakfast of champions :)

---

The two inline comments I left are to remind us about a small formatting issue (new PR) and a mystery comment promising a new issue.

glouppe:
Thanks for the great work @iaroslav-ai !

iaroslav-ai:
Indeed a big one :D :tada:
@betatim and @MechCoder, thanks so much for reviewing! @betatim , I will do exactly as you suggest :D
Big thanks to everyone else for useful suggestions and words of encouragement!

skopt/tests/test_parallel_cl.py
iaroslav-ai: Actually now that I think about it I believe I already opened an issue about this in #348 - somehow default acquisition function does not play well with this surrogate.

Issue #338, opened 2017-04-03
Default base_estimator for Optimizer


glouppe:
Currently, the constructor of `Optimizer` expects `base_estimator`. 

It would be a good default I think to have the same GaussianProcessRegressor as in `gp_minimize` when `base_estimator=None`. 

(Reproducing that logic is not trivial for the user, hence that would be a nice shortcut.)

PR #345, opened 2017-04-09
Use all surrogates in bench_ml.py

iaroslav-ai:


MechCoder:
Thanks! Could you also add the results if you have the time.

Could you also provide a random state parameter to MLBench and use that random state to do things:
1. The train test split.
2. And also fix the random state if the model has a random state parameter in the constructor over here (https://github.com/scikit-optimize/scikit-optimize/blob/master/benchmarks/bench_ml.py#L214)

(It was my mistake that I removed the random state, and it might be better to assume that the black-box function is non-noisy for reproducible results)

benchmarks/bench_ml.py
```diff
@@ -352,7 +353,7 @@ def run(n_calls=32, n_runs=1, save_traces=True, n_jobs=1):
     * `n_jobs`: int
         Number of different repeats of optimization to run in parallel.
     """
-    surrogates = [GaussianProcessRegressor]
+    surrogates = [RandomForestRegressor, ExtraTreesRegressor, GradientBoostingQuantileRegressor, GaussianProcessRegressor]
```
MechCoder: You could remove the `RandomForestRegressor` since we use the ExtraTreesRegressor by default. This is to create a smoothing of the decision surface.

iaroslav-ai:
Yup, will fix.
As for the benchmark results, somehow http://archive.ics.uci.edu/ml/ is down. I will see where else I could get the data.

iaroslav-ai:
I removed 
```
from urllib.parse import quote
from urllib.request import urlopen
```
as the imports are not used. Let me know if this breaks anything (seems to work fine without them).

iaroslav-ai:
Benchmark seems to work fine now. I will update this pull request with results as soon as I will have them.

iaroslav-ai:
First results with GP are there. However, I cannot use `GradientBoostingQuantileRegressor` and `ExtraTreesRegressor` for default values of the parameters of `Optimizer`, as it throws an exception (that predict does not have `return_mean_grad` parameter). I will set  `acq_optimizer='sampling'` for now, I assume that is what Optimizer is intended to do anyway.

iaroslav-ai:
Currently I have results for `GaussianProcessRegressor` and `ExtraTreesRegressor`. I do not have results for `GradientBoostingQuantileRegressor` due to #348 . Maybe we just start with these results, and add results for the missing surrogate later.
I also have logs of the optimization procedure, from which one can extract performance for some particular configurations of the dataset / model / surrogate. I currently have them in the repository, but plan to move the data to my own storage until it is needed, unless you have a better idea.

glouppe:
Suggestion: instead of passing the surrogate models, would it be possible instead to iterate over the functions `gp_minimize`, `forest_minimize`, `gbrt_minimize`, etc? The advantage is that these functions have been configured internally with good defaults for the surrogate models. Using instead the `Optimizer` class with its current interface leads to having to reproduce the logics of putting good defaults, which we should try to avoid.

This connects to #338 

What do you think?

iaroslav-ai:
@glouppe Yeah should be doable. Especially if you say that the parameters are optimized for these functions, then of course it makes a lot of sense.
Personally I like "ask-tell" interface more because it is in my experience more flexible. Maybe it would be a good idea to have something like GPOptimizer class, where one does not need to configure the surrogate models, yet retaining the flexibility of Optimizer class.

iaroslav-ai:
Or maybe just transfer the parameters from the gp_minimize to default parameters of corresponding surrogate, in case this is sufficient.

benchmarks/README.md
```diff
@@ -12,17 +12,20 @@ repeated 10 times.
 ------------------|------------|-----------|---------------------|--------------------|-----------------------
 |Branin| 0.398 +/- 0.000 |0.398 | 33.1 | 5.7 | 27
 |Hart6| -3.299 +/- 0.048|-3.322 | 57.3 | 5.658 | 47
+|ML_hypp_tune| -0.860 +/- 0.045 | -0.898 | NA | NA | NA
```
MechCoder: We should split the regressors and classifiers, else this is tough to interpret.

MechCoder: Also as @glouppe suggested, we should use `log_loss` from `sklearn.metrics` since he accuracy is not continuous in the hyperparameter space.

iaroslav-ai: Yeah, I think splitting up the results obtained with the benchmark in general is a good idea, so that one can easily see results for different types of learning problems.

iaroslav-ai: I will give cross entropy a try. Do you expect that there will be some problems with the optimization algorithms if the objective is not continuous wrt hyperparameters? Should one in general try to avoid objectives that are not continuous?

iaroslav-ai:
Results are coming today or latest tomorrow. I will commit all the changes to the code together with new results (using cross entropy and using x_minimize methods).

benchmarks/README.md
MechCoder: No, it is just that the GP makes a prior assumption that `f(x)` follows a multivariate gaussian and hence a continuous function is easier to model for the GP. However I don't think that is a concern for tree-based methods.

It would be interesting to try both and see if tree-based methods have a significant advantage if the function is discontinuous.

benchmarks/bench_ml.py
```diff
@@ -301,14 +313,14 @@ def evaluate_optimizer(surrogate, model, dataset, n_calls, random_state):
...
-    problem = MLBench(model, dataset)
+    problem = MLBench(model, dataset, random_state)
     space = problem.space
 
     # initialization
     estimator = surrogate(random_state=random_state)
     dimensions_names = sorted(space)
     dimensions = [space[d][0] for d in dimensions_names]
-    solver = Optimizer(dimensions, estimator, random_state=random_state)
+    solver = Optimizer(dimensions, estimator, random_state=random_state, acq_optimizer='sampling', acq_func='EI')
```
MechCoder: Can you change this to `acq_optimizer` = lbfgs if the model used is a gaussian process?

glouppe: It would really be better if the optimization was carried out through the `*_minimize` functions, for which all these good default values are already set.

iaroslav-ai: Yup, as @glouppe suggested, I used *_minimize methods, which make it quite simple with setting up the right parameters.

iaroslav-ai:
I updated the readme with new results. Maybe I also calculate `f_calls` metrics; While I agree that the results rather represent average over many benchmarks rather than one benchmark, it might still be useful for comparison of algorithms.

iaroslav-ai:
Hmm I realized that I run the benchmark with n_calls = 50, while it should be 200. I will re - run the benchmark.

MechCoder:
50 calls are okay. I am fine with merging this as is.

MechCoder:
Do we need the jsoin files?

benchmarks/bench_ml.py
```diff
@@ -301,36 +328,37 @@ def evaluate_optimizer(surrogate, model, dataset, n_calls, random_state):
...
         # the result of "evaluate" is accuracy / r^2, which is the more the better
         objective_at_point = -problem.evaluate(point_dct)
 
-        if best_y > objective_at_point:
-            best_y = objective_at_point
+        if shared_values[-1] > objective_at_point:
+            shared_values[-1] = objective_at_point
+
+            # remember the point, objective pair
+        trace.append((point_dct, objective_at_point, shared_values[-1]))
```
MechCoder: The `results` instance as returned from the minimize function provides all the information that you need. I would prefer this to be done outside the objective function.

benchmarks/bench_ml.py
```diff
@@ -211,17 +218,34 @@ def evaluate(self, point):
         point_mapped = {}
         for param, val in point.items():
             point_mapped[param] = self.space[param][1](val)
+
+        # this is necessary to make things run in parallel
```
MechCoder: I thought this was for reproducible results.

benchmarks/bench_ml.py
```diff
@@ -268,11 +295,11 @@ def fmt(float_value):
         output = "|".join([fmt(min_mean) + " +/- " + fmt(min_std), fmt(min_best)])
         result = table_template + output
         print("")
-        print(algorithm)
+        print(key)
         print(result)
 
 
-def evaluate_optimizer(surrogate, model, dataset, n_calls, random_state):
+def evaluate_optimizer(surrogate_minimize, model, dataset, n_calls, random_state):
```
MechCoder: Can you update documentation for this?

MechCoder:
Looks fine. Thanks for your work. Bit surprised GBRT minimize beats the others for regression!

benchmarks/bench_ml.py
iaroslav-ai: This was a comment from some code experimentation I had earlier, I just forgot to update the comment :)
Relevant to this: the parallel computation did not work, I am not sure why, but it seems that it was due to same instance of estimator used in multiple processes. I switched to passing classes to threads instead of instances, and everything seems to work fine now.

benchmarks/bench_ml.py
iaroslav-ai: updated

iaroslav-ai:
I am happy if you merge this as is. I will make PR if the results that I get with 200 calls are drastically different. 
Thank you for reviewing my code!
Indeed it is surprising that GBRT beats everything else.

betatim:
🌳  for lyfe! Thanks for the contribution and reviewing!

PR #353, opened 2017-04-12
[MRG] Add delta y early stopping callback

betatim:
Follow up on #6 and #336. This adds a "stop if delta y of N best points is less than delta".

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -158,6 +158,26 @@ class EarlyStopper(object):
         """
         return self._criterion(result)
 
+    def _criterion(self, result):
+        """Compute the decision to stop or not.
+
+        Classes inheriting from `EarlyStop` should use this method to
+        implement their decision logic.
+
+        Parameters
+        ----------
+        * `result` [`OptimizeResult`, scipy object]:
+            The optimization as a OptimizeResult object.
+
+        Returns
+        -------
+        * `decision`:
+            Return True/False if the criterion can make a decision or `None` if
+            there is not enough data yet to make a decision.
+        """
+        raise NotImplementedError("The _criterion method should be implemented"
+                                  " by subclasses of EarlyStopper.")
+
 
 class DeltaXStopper(EarlyStopper):
     """Stop the optimization when |x1 - x2| < `delta`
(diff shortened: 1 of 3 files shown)


skopt/callbacks.py
```diff
@@ -176,3 +178,28 @@ def _criterion(self, result):
...
+    objective values is less than `delta`.
+    """
+    def __init__(self, delta, n_best=5):
+        super(EarlyStopper, self).__init__()
+        self.delta = delta
+        self.n_best = n_best
+
+    def _criterion(self, result):
+        if len(result.x_iters) >= self.n_best:
+            x_iters = np.sort(result.x_iters)
```
MechCoder: Should this be `result.func_vals`?

skopt/callbacks.py
```diff
@@ -176,3 +178,28 @@ def _criterion(self, result):
...
+            worst = x_iters[self.n_best - 1]
+            best = x_iters[0]
+
+            if np.abs(worst - best) < self.delta:
```
MechCoder: worst is always greater than best.

skopt/callbacks.py
```diff
@@ -176,3 +178,28 @@ def _criterion(self, result):
...
             return None
+
+
+class DeltaYStopper(EarlyStopper):
+    """Stop the optimization if the `n_best` minima are within `delta`
+
+    Stop the optimizer if the absolute difference between the `n_best`
+    objective values is less than `delta`.
+    """
+    def __init__(self, delta, n_best=5):
```
MechCoder: Is `n_best=2` a better default? That is stop when the new point is in delta of the previous point.

skopt/callbacks.py
betatim: Yes! I need better 👓

skopt/callbacks.py
betatim: Not sure. This implements "delta between N best", with N=2 that would be "delta between best and last added". If we want "delta between last two points added" that would be something different.

My thinking was that because of exploration you might not want "delta between last N" but "delta between N best".

MechCoder: ok cool

MechCoder:
Looks fine except the inline comments.

betatim:
:clinking_glasses:

PR #354, opened 2017-04-15
MRG: raises a ValueError whenever base_estimator is tree-based and acq_optimizer evaluates to False

mirca:
This is an attempt to fix #351.

Let me know if that's the correct approach. I will be happy to add more unit tests accordingly.

PR #360, opened 2017-04-18
MAINT: remove deprecation warning when acq_optimizer='auto'

mirca:
Based on the chat in #354, I'm removing the deprecation warning when ``acq_optimizer='auto'``.

I think after #354, it would make sense to let ``acq_optimizer='auto'`` as default... What do you think?

Does this need an entry in the change log?

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -162,6 +162,10 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
+    if acq_optimizer == "auto":
+        warnings.warn("The 'auto' option for the acq_optimizer will be "
+                      "removed in 0.4.")
+
     optimizer = Optimizer(dimensions, base_estimator, n_random_starts,
                           acq_func, acq_optimizer, random_state,
                           acq_optimizer_kwargs=acq_optimizer_kwargs,
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -162,10 +162,6 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
-    if acq_optimizer == "auto":
-        warnings.warn("The 'auto' option for the acq_optimizer will be "
-                      "removed in 0.4.")
-
     optimizer = Optimizer(dimensions, base_estimator, n_random_starts,
                           acq_func, acq_optimizer, random_state,
                           acq_optimizer_kwargs=acq_optimizer_kwargs,
--- a/skopt/optimizer/forest.py
+++ b/skopt/optimizer/forest.py
@@ -7,9 +7,12 @@ from ..learning import ExtraTreesRegressor
 from ..learning import RandomForestRegressor
 
 
-def forest_minimize(func, dimensions, base_estimator="ET", n_calls=100,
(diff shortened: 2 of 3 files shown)


MechCoder:
Yes, can you add an entry in the change-log.

Also, just FYI the "auto" option was there to configure the `acq_optimizer` option on the basis of the space configuration which was removed by me over here (https://github.com/scikit-optimize/scikit-optimize/pull/285/)

MechCoder:
It should be set to sampling here as well (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/forest.py#L12).  :/

betatim:
We removed `auto` as a valid parameter value and are now adding it back with a different meaning. How confused are users going to be?

---

> It should be set to sampling here as well (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/forest.py#L12).  :/

can we remove it all together from `forest_optimize` (and the other tree based optimize calls if they list it). For now `sampling` is the correct and only choice so we don't need a parameter :)

MechCoder:
There are two things.

1. The `acq_optimizer` argument as supplied to the Optimizer directly. Since the Optimizer class is a new class the "auto" argument for the `acq_optimizer` did not exist before.
2.  The deprecation warnings are for the `acq_optimizer` as supplied through the *_minimize functions depending upon the Space configuration.

We can move the warning out of the `Optimizer` class and into the `base_minimize` function.

mirca:
@MechCoder @betatim Just to be crystal clear, should add a deprecation warning for ``acq_optimizer=auto`` in ``base_minimize``?

skopt/optimizer/forest.py
```diff
@@ -7,12 +7,10 @@
 from ..learning import RandomForestRegressor
 
 
-def forest_minimize(func, dimensions, base_estimator="ET",
-                    n_calls=100, n_random_starts=10,
-                    acq_func="EI", acq_optimizer="auto",
```
MechCoder: This will break user-code. Let's keep it for now and hopefully we will add something better later on?

MechCoder:
Yes, that is what I had meant. WDYT?

skopt/optimizer/base.py
```diff
@@ -162,6 +162,10 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
+    if acq_optimizer == "auto":
+        warnings.warn("The 'auto' option for the acq_optimizer will be "
+                      "removed in 0.4.")
```
betatim: Shouldn't we remove this? The next release will be v0.4 so we can't remove the option?

MechCoder: Yes indeed we should remove it from `base_minimize`

skopt/optimizer/optimizer.py
```diff
@@ -69,11 +69,15 @@ class Optimizer(object):
                 - After fitting the surrogate model with `(X_best, y_best)`,
                   the gains are updated such that $g_i -= \mu(X_i)$
 
-    * `acq_optimizer` [string, `"sampling"` or `"lbfgs"`, default=`"lbfgs"`]:
+    * `acq_optimizer` [string, `"sampling"`, `"lbfgs"`, or `"auto"`, default=`"lbfgs"`]:
```
MechCoder: you should remove if from here as well.

MechCoder:
Merging as non-controversial.

MechCoder:
Thanks!

betatim:
🍻 thanks for this!

mirca:
@betatim @MechCoder Awesome! x)

PR #362, opened 2017-04-18
[MRG] Updated benchmark results from bench_ml.py

iaroslav-ai:
These are updated results with n_calls = 200 and dummy_minimize.
For most part, minimize algorithms are better than random strategy, but the difference is not as big as I expected. This might be due to the problems being fairly simple (but I am not sure). 
To get a more in - depth look at things, I made a plot of algorithm iteration vs average optimum value. Here it is for classification, all algorithms appear to be better than random:

classification

Here it is for regression, somehow forest does not do too good, not sure if results are reliable though (warning - color / method assignment is not preserved from previous graph): 

regression

I will rerun the benchmarks with eg n_runs = 50 to be more confident in results that I am getting. Also it seems that n_calls = 50 should be sufficient, as differences seem to be pretty consistent across different iterations. 

Maybe some modifications would be good to benchmarks eg in what datasets are used in order to make tasks somewhat "harder", maybe then the difference would be larger.


iaroslav-ai:
Hmm yeah it seems that indeed more n_calls are necessary especially with Regression. I made plots similar to the above, however I also plotted standard error of measurement () for dummy_minimize (I assume normal distribution of errors for simplicity, maybe you let me know if this is not reasonable). So for classification there are algorithms that are clearly outside the SEM of dummy_minimize. In the plot below the sticks around dummy_minimize points indicate standard error of measurement:

clsf_sem

For regression it is not really the case (gbrt is somewhere on the boundary):

regr_sem

So more n_calls are necessary. And indeed, maybe some tougher ML problems with one clear global optimum are necessary in order to have a more clear distinction between the minimization methods and the random strategy (with less n_calls that is).

betatim:
`n_calls` is the number of times you repeated the whole minimisation correct? (just double checking because I wasn't paying close attention)

As the points in one series are highly correlated with each other maybe it is enough to put the uncertainty only on the last point? But this is just visual nitpicking.

benchmarks/README.md
```diff
@@ -12,23 +12,30 @@ repeated 10 times.
...
-|ML, Regression| -0.755 +/- 0.171 | -0.806 | NA | NA | NA
+|ML, Classification| 1.285 +/- 0.043|1.229 | NA | NA | NA
+|ML, Regression| -0.781 +/- 0.026 | NA | NA | NA
+
+## dummy_minimize
+
+|Blackbox Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min
+------------------|------------|-----------|---------------------|--------------------|-----------------------
+|ML, Classification| 1.284 +/- 0.029 | 1.241 | NA | NA | NA
+|ML, Regression| -0.772 +/- 0.023 | -0.802 | NA | NA | NA
```
betatim: will you add branin and hart6 here as well?

betatim:
Maybe expand on the sentence in `README.md` that explains to people how to reproduce the benchmark numbers. I'd put the full command line for the benchmarking script. After that merge?!

iaroslav-ai:
@betatim oooops sorry instead of n_calls I meant n_runs

benchmarks/README.md
iaroslav-ai: yeah should not be too hard

iaroslav-ai: that is, yes I will add them as well :)

glouppe:
I was wondering, why do we have all these "NA"s in table?

MechCoder:
Do you want to add the other benchmarks over here? I also think it is okay to plot the error bars at the end.

The problems seem to be easy and there is a clear trend for example with the regularization in SVC or max_depth for a DecisionTree.

I would also suggest to split the problems on the basis of the underlying regressor or classifier in order to have a clearer idea.

iaroslav-ai:
@glouppe: some time ago me and MechCoder decided to remove number of calls until best minimum and related metrics, I do not remember exactly the reason to be honest. I actually have the code in my working branch that calculates these, but for different best minima these are not always useful to compare (eg better minima could be found in somewhat larger number of steps). I will put those here so that you can take a look. 
@betatim @MechCoder: I have seen some examples of problems that I am experimenting right now where SEM are quite different for different iterations. I find it useful also in order to see how difference with random strategy depends on number of iterations (n_calls).
@MechCoder I have some results now with n_runs = 40 (all but gp), interestingly I see a fairly consistent trend in algorithm performance across different models. So it might even be sufficient to concentrate on one / few problems among set of all we got. 

--- 

There could be a few ways how to improve complexity of tasks:
* Do feature selection together with parameter selection. Complexity could be improved further by adding random features and by restricting number of features that could be selected, so that algorithm needs to learn which features are “most useful” 
* Add parameters that enable / disable normalization
* One could even add parameters which if not set right cause objective to drop eg to 0.0, but this might be a bit too much :)

iaroslav-ai:
Fixed some bugs and updated the benchmarks with dummy_minimize results. I also restructured the way results are presented - as it is right now I find it easier to compare algorithms.

benchmarks/README.md
```diff
@@ -6,29 +6,49 @@ functions provided in Table 2 of http://www.automl.org/papers/13-BayesOpt_Empiri
 All optimizers are run with default values and with `n_calls=200`. Runs are
 repeated 10 times.
 
-## gp_minimize
+## Branin
 
-|Blackbox Function| Minimum | Best minimum | Mean f_calls to min | Std f_calls to min | Fastest f_calls to min
+To reproduce, run `python bech_branin.py`
```
MechCoder: Should be bench

iaroslav-ai: Oops fixed :)

betatim:
Time for review or is there something left to add?

iaroslav-ai:
Please review. Further changes would be to the tasks as such, for which I think a new pull request is more appropriate.

betatim:
Cool. In the future can you change the title of the PR to "[MRG] ..." when things are ready to be merged? We copied this idea from scikit-learn: [WIP] tag for PRs that are being worked on, then [MRG] when it is ready to be merged/reviewed. 📋

betatim:
Tried to run the benchmark locally and it seems to get deadlocked?
```
$ python bench_hart6.py                                                 (skopt)
gp_minimize
0
Iteration No: 1 started. Evaluating function at random point.
Iteration No: 1 ended. Evaluation done at random point.
...
Time taken: 3.6441
Function value obtained: -3.3222
Current minimum: -3.3224
Iteration No: 128 started. Searching for the next optimal point.
^CProcess ForkPoolWorker-2845:
Process ForkPoolWorker-2842:
Process ForkPoolWorker-2844:
Process ForkPoolWorker-2843:
Process ForkPoolWorker-2841:
Process ForkPoolWorker-2838:
Process ForkPoolWorker-2840:
Process ForkPoolWorker-2839:
```
I killed it after it sat there for minutes. Full trace back here: https://gist.github.com/betatim/6d93c4ecdcbf4bba3fd91cb6de616ad8

Any ideas?

betatim:
If I set `n_jobs=1` for `gp_minimize` it seems to work fine. 😕

iaroslav-ai:
@betatim  Sorry, will use "MRG" next time, I still got much to learn :smile: 
Wrt gp_minimize: I am running it for ~20 mins by now with n_jobs=-1, and so far it works a ok on my Ubuntu 16 machine. Maybe this is some system specific thing?

betatim:
Let's make a new issue then to deal with the threading issue. The other observation (from looking at `top` briefly) is that the python process doesn't seem to use more than one CPU -> are we sure setting `n_jobs>1` actually gives a speed up?

(For history: this is on my laptop/OSX 10.11/python 3.5)

iaroslav-ai:
In my case all 8 of my CPU cores are loaded ~100% when I run gp_minimize with n_jobs =-1. Not sure about multiple CPU's though (did you mean more than one CPU core by any chance?).

betatim:
I was just looking at % of CPU used for the process which is usually >100% if it has multiple threads.

betatim:
screen shot 2017-04-28 at 08 58 41
like this

iaroslav-ai:
I see. This is what top command outputs for me (relevant part) when I run the benchmark:

image

Seems to be utilizing most of cores, but I am running on python 2.7.

MechCoder:
LGTM. It seems to work fine for me.

betatim:
Weird. On my linux machine it works.

Merge?!

PR #368, opened 2017-04-25
[MRG+1] Converts tests to pytest

thomasjpfan:
Converts tests from nosetests to pytest. The only real change needed was to convert the yield statements that nosetests uses to function calls. pytest shows the duration of test runs by default. nose is still a test requirement because of `assert_array_equal` from the sklearn library.

#220

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ install: source build_tools/travis/install.sh
 
 script:
   - if [ ${COVERAGE} == "true" ];
-    then pytest --cov=skopt --durations=10; else
-    pytest --durations=10;
+    then travis_wait 25 pytest --cov=skopt --durations=10; else
+    travis_wait 25 pytest --durations=10;
     fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ install: source build_tools/travis/install.sh
 
 script:
   - if [ ${COVERAGE} == "true" ];
-    then travis_wait 25 pytest --cov=skopt --durations=10; else
-    travis_wait 25 pytest --durations=10;
+    then pytest --cov=skopt --durations=10; else
+    pytest --durations=10;
     fi
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ install: source build_tools/travis/install.sh
 
 script:
   - if [ ${COVERAGE} == "true" ];
-    then nosetests --with-coverage --cover-package=skopt --with-timer --timer-top-n 10; else
-    nosetests --with-timer --timer-top-n 10;
+    then pytest --cov=skopt --durations=10; else
+    pytest --durations=10;
     fi
 
 after_success:
--- a/skopt/learning/gaussian_process/tests/test_kernels.py
+++ b/skopt/learning/gaussian_process/tests/test_kernels.py
@@ -3,6 +3,7 @@ from scipy import optimize
 from scipy.spatial.distance import pdist, squareform
 from sklearn.utils.testing import assert_array_almost_equal
 from sklearn.utils.testing import assert_array_equal
+import pytest
 
 from skopt.learning.gaussian_process import GaussianProcessRegressor
 from skopt.learning.gaussian_process.kernels import ConstantKernel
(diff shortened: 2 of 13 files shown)


glouppe:
This is great! Thanks a lot :)

I'll try to have a look tomorrow, unless @MechCoder or @betatim come by in the meantime.

.travis.yml
```diff
@@ -17,8 +17,8 @@ install: source build_tools/travis/install.sh
 
 script:
   - if [ ${COVERAGE} == "true" ];
-    then nosetests --with-coverage --cover-package=skopt --with-timer --timer-top-n 10; else
-    nosetests --with-timer --timer-top-n 10;
+    then travis_wait 25 pytest --cov=skopt --durations=10; else
+    travis_wait 25 pytest --durations=10;
```
betatim: for my education, what does `travis_wait` do? Does it time out if it runs for more than 25s?

betatim: Ah, just saw your commit message :)

betatim:
While we are at it, one thing that was tedious about our old test setup (and extensive use of `yield`) was that nosetest would tell you that `check_xyz` with arguments `<foobar repr...>` failed. Which often felt like I had to debug the tests to help me debug the real problem. Basically it was too hard to tell which of the arguments combos to the `check_func` failed.

Here my question: should we be switching to using https://docs.pytest.org/en/latest/example/parametrize.html? This seems to generate nicer messages in the output but it does mean switching a lot of stuff around. (Even if we decide yes this is a good idea we might want to delay it to a future PR and not slow this one down.) Just hoping to have the attention of someone with some pytest knowledge to get some advice :)

.travis.yml
thomasjpfan: In Python 2, one of the tests in `skopt/tests/test_gp_opt.py` took longer than 10 mins to finish. I added this 25 minute buffer, so that Travis CI does not error out when a test is taking too long.

thomasjpfan:
My latest commit refactors the original tests into smaller tests using pytest's parametrize framework. These smaller tests accept parameters through the `@pytest.mark.parametrize` decorator, which were originally iterated by a `for` loop. If a test were to fail, pytest will display the parameters that caused that test to fail.

thomasjpfan:
Refactoring the benchmark tests into smaller tests allows for all tests to run less than 10 minutes. This removes the need for `travis_wait 25` in the .travis.yml file.

skopt/learning/gaussian_process/tests/test_kernels.py
```diff
@@ -69,12 +70,12 @@ def check_gradient_correctness(kernel, X, Y, step_size=1e-4):
...
+def test_gradient_correctness(kernel):
     rng = np.random.RandomState(0)
     X = rng.randn(5)
     Y = rng.randn(10, 5)
-    for kernel in KERNELS:
-        yield check_gradient_correctness, kernel, X, Y
+    check_gradient_correctness(kernel, X, Y)
 
 
 def test_gradient_finiteness():
```
betatim: `test_gradient_finiteness()` -> `test_gradient_finiteness(kernel)`?

betatim: as well as `random_state` as parameter?

betatim:
LGTM.

glouppe:
Thanks a lot for contributing :)

This looks good to me too... Merging!

betatim:
:tada: this is super! Now we can mark the heavy tests as heavy -> faster local execution!

skopt/learning/gaussian_process/tests/test_kernels.py
thomasjpfan: Ah yes, I missed that one. That test would become:
```
@pytest.mark.parametrize("random_state, kernel", product([0, 1], KERNELS))
def test_gradient_finiteness(random_state, kernel):
    """
    When x is the same as X_train, gradients might become undefined because
    they are divided by d(x, X_train).
    Check they are equal to numerical gradients at such points.
    """
    rng = np.random.RandomState(random_state)
    X = rng.randn(5).tolist()
    Y = [X]
    check_gradient_correctness(kernel, X, Y, 1e-6)
```

MechCoder:
Thanks!

PR #369, opened 2017-04-26
[MRG] Model expected inverse time along with function

MechCoder:
- [x] Add tests
- [x] Support `return_grad=True` when `per_second=True`


betatim:
Is there a good rule of thumb how to weight time and other score? I'm imagining a model where you reduce the number of layers/max_depth of trees which will lead to faster training times but potentially worse performance. Trying to understand how this is taken into account in EI per second.

skopt/optimizer/base.py
```diff
@@ -127,13 +129,26 @@ def base_minimize(func, dimensions, base_estimator,
...
+        is the time taken for the function call at x.
+        Note that
+
+            - `log(t) ~ N(m, sigma)`
+            - `-log(t) ~ N(-m, sigma)`
+            - `log(1/t)` ~ N(-m, sigma)
+
+        This implies `1/t` follows a lognormal distribution with mean `-m`
+        and `sigma` or `E(1/t) = exp(-m + sigma**2)`
+        The next point suggested is the point that minimizes `acq(x)*E(-1/t(x))`
```
glouppe: Just curious, do you have a reference for this strategy? 

A smart strategy, but which would require quite some changes in our codebase, is freeze-thaw bayesian optimization (http://arxiv.org/abs/1406.3896) where the idea is basically to extrapolate the function wrt some its parameters (like n_epochs). This is in turn allows to do early stopping, thereby reducing the total execution time.

MechCoder: Yes, see section 3.2 of https://arxiv.org/pdf/1206.2944.pdf . 

I'll have to look at the other paper though.

glouppe:
While in general I am in favor of taking running time into account, I also think we should be cautious in what we end up implementing. Most of those approaches are very state-of-the-art -- that is none are well-established. So before adding yet another option to the minimizers, I think it would be wise to really establish first whether this is worth it.

MechCoder:
I removed the `include_time` parameter and added support for acquisition functions based on the suffix "ps". I still need to finish manually compute gradients for optimizations.

I'll add tests and some experiments in a while.

MechCoder:
Can someone make a pass for correctness?

MechCoder:
I wrote a toy example here, the function optimized is a constant function while the time is proportional to the input. Here are the "EI" and "EIps" values after 3 points have been seen.

```python
def contant_func(x):
    return 10.0

def time_func(x):
    """
    Hard-code such that time is x[0]
    """
    return contant_func(x), log(x[0])

x = [3, 5, 7]
x_1D = np.reshape(x, (-1, 1))
y = [contant_func(xi) for xi in x]
x_pred = np.linspace(1, 9, 100)
x_pred1D = np.reshape(x_pred, (-1, 1))

# Naive EI.
gpr = GaussianProcessRegressor(random_state=0)
gpr.fit(x_1D, y)
y_pred = gpr.predict(x_pred1D)
ei_vals = gaussian_ei(x_pred1D, gpr, y_opt=10.0)

# EI that takes into account time.
y_2D = [time_func(xi) for xi in x_1D]
mor = MultiOutputRegressor(gpr)
mor.fit(x_1D, y_2D)
eips_vals = gaussian_ei(x_pred1D, mor, y_opt=10.0, per_second=True)

plt.plot(x, y, "ro")
plt.plot(x_pred, y_pred, label="mean")
plt.plot(x_pred, ei_vals, label="EI")
plt.plot(x_pred, eips_vals, label="EIps")
plt.legend()
plt.show()
```

eips

betatim:
I can take a closer look maybe this afternoon (EU time) or next week. Had a quick look now and was wondering if we should make the infrastructure to support "per second" already flexible enough to handle arbitrary multi-objective problems? I think that would mainly influence decision in `Optimizer`, the user facing parts would stay the same. Thoughts?

MechCoder:
We can do that but I don't think the present acquisition function machinery can handle multitask problems. I'd prefer writing separate acquisition functions as and when we decide to support that.

MechCoder:
Updated to [MRG]. Ready for review from my side.

MechCoder:
Can I attract some attention here?

betatim:
I don't really know much about the science side of EI per second etc. Would be good to have @glouppe comment on that, at least he seems to know that none are established methods??

MechCoder:
Since @glouppe 's initial comment, I have moved the option to include time as an option to the minimizers to a new acquisition function. Not sure what you mean by "the science side of EI", can you please elaborate? Also, surely my links to the original bayesian optimisation paper and the matlab toolbox that allows such options make it a more established method than gbrt_minimize which our codebase supports which hasn't been published?

MechCoder:
In the graph (https://github.com/scikit-optimize/scikit-optimize/pull/369#issuecomment-305696871), you can see a constant function being modelled by the GP, with time proportional to the logarithm of x. Including the time, makes sure that if there are two points to the left and right of the origin, it is likelier to pick the point on the left.

You are right that one cannot fine-tune the balance between the function and time, but I would prefer implementing methods that are already there rather than something new as a starting point.

MechCoder:
Maybe @iaroslav-ai has the time to look?

iaroslav-ai:
Yeah, I will take a look

iaroslav-ai:
Review incoming in a day or two - I just need to formulate my thoughts a bit more on this

skopt/acquisition.py
```diff
@@ -34,23 +34,40 @@ def _gaussian_acquisition(X, model, y_opt=None, acq_func="LCB",
...
     else:
         raise ValueError("Acquisition function not implemented.")
 
+    if return_grad:
+        return acq_vals, acq_grad
+
+    if per_second:
+        mean, std = time_model.predict(X, return_std=True)
+        inv_t = np.exp(-mean + 0.5*std**2)
+        return acq_vals * inv_t
```
iaroslav-ai: Could the acquisition function value `acq_vals` be positive? In this case longer expected computation leads to a smaller expected objective according to this expression. This seems to be the opposite of what we pursue, we are looking for minimizing value here, right? Also as @betatim suggested would be good if user can specify a tradeoff parameter between objective and time.

MechCoder: nice catch. acq_vals here is the negation of the actual acquisition values. So:

1. PI is always positive, (it is probability)
2. EI is always positive, since it is the expected value of a non-negative function.

I don't think any such bounds apply for LCB (and gp-hedge). I'll remove the function "LCBps" and "gp-hedgeps"

MechCoder:
@iaroslav-ai I've removed references to LCBps and gphedgeps

skopt/acquisition.py
iaroslav-ai: What about gradient of acquisition function in this case? Do you plan to implement it later? Maybe an exception would be nice here (if you do not have it somewhere esle). Product rule gives something like

d(acq_vals * inv_t) / dx = -acq_grad * inv_t + acq_vals *  d(inv_t)/dx,

where x is some point and as it seems that acq_grad = -d(acq_vals)/dx. Seems like it is possible to have generic expression for many acquisitions.

skopt/optimizer/base.py
```diff
@@ -63,6 +63,11 @@ def base_minimize(func, dimensions, base_estimator,
         - `"LCB"` for lower confidence bound,
         - `"EI"` for negative expected improvement,
         - `"PI"` for negative probability of improvement.
+        - ``"EIps"`` for negated expected improvement per second to take into
+          account the function compute time. Then, the objective function is
```
iaroslav-ai: Nitpick: maybe drop "Then, ", seems a bit redundant

skopt/optimizer/base.py
```diff
@@ -222,9 +227,14 @@ def base_minimize(func, dimensions, base_estimator,
...
-            raise ValueError(
-                "`y0` elements should be scalars")
+        if acq_func not in ["EIps", "Pips"]:
+            if not all(map(np.isscalar, y0)):
+                raise ValueError(
+                    "`y0` elements should be scalars")
+        else:
+            if np.ndim(y0) != 2 and np.shape(y0)[1] != 2:
+                raise ValueError(
+                    "`y0` elements should be a tuple of 2 values.")
```
iaroslav-ai: Nitpick: one might get confused by this message by assuming that `y0` as such should be a tuple of 2 values. Maybe use something like "Every element in y0 should be a tuple or list containing 2 scalar values"

skopt/acquisition.py
```diff
@@ -34,23 +34,40 @@ def _gaussian_acquisition(X, model, y_opt=None, acq_func="LCB",
...
         if return_grad:
-            return -func_and_grad[0], -func_and_grad[1]
+            acq_vals = -func_and_grad[0]
+            acq_grad = -func_and_grad[1]
         else:
-            return -func_and_grad
+            acq_vals = -func_and_grad
 
     else:
         raise ValueError("Acquisition function not implemented.")
```
iaroslav-ai: Consider maybe adding here also some code which prints the list of supported acquisition function identifiers. Might be useful for someone who is familiar with BO but not with skopt, or for someone lazy who does not want to look up how to spell the acquisition name properly.

skopt/acquisition.py
iaroslav-ai: Clarification, "Maybe an exception would be nice here": in case return_grad = True, but ps acquisition is used

skopt/tests/test_acquisition.py
```diff
@@ -17,9 +20,27 @@
 
 class ConstSurrogate:
     def predict(self, X, return_std=True):
+        X = np.array(X)
         return np.zeros(X.shape[0]), np.ones(X.shape[0])
 
 
+class MultiOutputSurrogate:
```
iaroslav-ai: Its more like a two output surrogate, we might want to add multi - output surrogates later, might be better to spare this name for now to use it later. Could you use something along the lines of "TwoOutputSurrogate"?

skopt/tests/test_acquisition.py
```diff
@@ -74,3 +95,16 @@ def test_acquisition_gradient():
...
+def test_acquisition_per_second():
+    X = [[3], [5], [7]]
+    y = [[1, log(3)], [1, log(5)], [1, log(7)]]
+    mos = MultiOutputSurrogate()
+    mos.fit(X, y)
+
+    X_pred =  [[1], [2], [4], [6], [8], [9]]
+    for acq_func in ["EIps", "PIps"]:
+        vals = _gaussian_acquisition(X_pred, mos, y_opt=1.0, acq_func=acq_func)
+        for fast, slow in zip([0, 1, 2], [5, 4, 3]):
```
iaroslav-ai: Could you here just check if vals is a sorted list? Something like 

`all(vals[i] <= vals[i+1] for i in range(len(vals)-1))`

Or maybe skip some elements to be sure 

`all(vals[i] <= vals[i+2] for i in range(len(vals)-2))`

iaroslav-ai: ... to make it a bit more easy to pass the test for the surrogate - while I would expect that most of the time it should learn the right thing, it might not just because data is not too much

skopt/tests/test_common.py
```diff
@@ -31,7 +31,7 @@
 # treated separately
 MINIMIZERS = [gp_minimize]
 ACQUISITION = ["LCB", "PI", "EI"]
-
+ACQ_FUNCS_PS = ["LCBps", "PIps", "EIps", "gp_hedgeps"]
```
iaroslav-ai: Don't forget to remove here the unsupported acquisitions

skopt/acquisition.py
iaroslav-ai: Sorry just saw the gradient on your todo list :)

iaroslav-ai:
Just thinking out loud: a user one can always have a tradeoff between value of time and objective in principle by just multiplying the time returned in the objective function by some constant. This would be good to clarify in documentation somewhere. Still explicit tradeoff parameter would be better I think.

skopt/acquisition.py
MechCoder: I have added detailed error messages in the Optimizer instance. It's unlikely that anyone will use this private function directly.

skopt/optimizer/base.py
```diff
@@ -226,9 +231,6 @@ def base_minimize(func, dimensions, base_estimator,
         if len(x0) != len(y0):
             raise ValueError("`x0` and `y0` should have the same length")
 
-        if not all(map(np.isscalar, y0)):
```
MechCoder: these checks were moved inside `Optimizer`

MechCoder:
@iaroslav-ai I've addressed all your comments and have added support for gradient with gradient checks. I would postpone adding an example and allowing the user to configure the tradeoff to another pull request. Please have another look.

skopt/optimizer/forest.py
```diff
@@ -74,6 +75,11 @@ def forest_minimize(func, dimensions, base_estimator="ET", n_calls=100,
         - `"LCB"` for lower confidence bound.
         - `"EI"` for negative expected improvement.
         - `"PI"` for negative probability of improvement.
+        - `"EIps" for negated expected improvement per second to take into
+          account the function compute time. Then, the objective function is
+          assumed to return two values, the first being the objective value and
+          the second being the time taken.
+        - `"PIps"` for negated probability of improvement per second.
```
iaroslav-ai: Consider adding here that the objective function is assumed to be the same as for 'EIps', or maybe describe "EIps" and "PIps" jointly. Otherwise someone who might be in a rush or tired can just miss the implied similarity of objective function as is for the 'EIps', which might lead to some confusion.

skopt/optimizer/optimizer.py
```diff
@@ -370,24 +388,40 @@ def tell(self, x, y, fit=True):
             only be fitted after `n_initial_points` points have been told to the
             optimizer irrespective of the value of `fit`.
         """
+        check_x_in_space(x, self.space)
+
+        if "ps" in self.acq_func:
+            if is_2Dlistlike(x):
+                if np.ndim(y) == 2 and np.shape(y)[1] == 2:
+                    y = [[val, log(t)] for (val, t) in y]
```
iaroslav-ai: A bit of a nitpick: maybe use log(t + 1e-3), for example in case user measures time in minutes, and then it might happen that for some task the time is zero minutes (< 60 seconds).

skopt/tests/test_acquisition.py
```diff
@@ -1,26 +1,52 @@
...
+from skopt.utils import cook_estimator
 
 
 class ConstSurrogate:
     def predict(self, X, return_std=True):
+        X = np.array(X)
         return np.zeros(X.shape[0]), np.ones(X.shape[0])
 
 
+class ConstantGPRSurrogate(object):
```
iaroslav-ai: Add a comment somewhere around to explain why you need those.

skopt/tests/test_acquisition.py
```diff
@@ -81,3 +107,34 @@ def test_acquisition_gradient():
 
     for acq_func in ["LCB", "PI", "EI"]:
         check_gradient_correctness(X_new, gpr, acq_func, np.max(y))
+
+
+@pytest.mark.fast_test
+@pytest.mark.parametrize("acq_func", ["EIps", "PIps"])
+def test_acquisition_per_second(acq_func):
```
iaroslav-ai: You check here whether the acquisition function is working properly with an example surrogate ConstantGPRSurrogate by checking if the time function wrt x is learned, right? Would be good to have a comment elaborating on this

iaroslav-ai:
LGTM when my latest comments are addressed!

skopt/optimizer/optimizer.py
MechCoder: Maybe it is easier to document that time should be in seconds?

skopt/tests/test_acquisition.py
MechCoder: Added below but the diff view does not collapse it.

MechCoder:
Will merge when tests pass.

PR #372, opened 2017-05-02
[MRG+1]: Remove acq_optimizer from forest_minimize

### LINKED ISSUE #367: acq_optimizer is not documented in forest_minimize
The argument ``acq_optimizer`` is not presented in the docstring of ``forest_minimize``.

Additionally, ``acq_optimizer`` is later passed as ``acq_optimizer="sampling"`` to ``base_minimize``. Although that makes sense because ``forest_minimize`` uses a tree-based regression model, I think the user should be aware of that behavior, at least. 

Should ``acq_optimizer`` be removed from ``forest_minimize``?

mirca:
As described in #367, ``acq_optimizer`` is not documented in ``forest_minimize`` and its value does not have any effect on ``forest_minimize``.

In this PR, I propose removing ``acq_optimizer`` from  ``forest_minimize``. As mention by @betatim, one caveat arises when another option (besides ``'sampling'``) for ``acq_optimizer`` is implemented. That can be solved by adding ``acq_optimizer`` back to the list of arguments and proper docstrings.

Fixes #367.

### DIFF
--- a/skopt/optimizer/forest.py
+++ b/skopt/optimizer/forest.py
@@ -7,12 +7,10 @@ from ..learning import ExtraTreesRegressor
 from ..learning import RandomForestRegressor
 
 
-def forest_minimize(func, dimensions, base_estimator="ET",
-                    n_calls=100, n_random_starts=10,
-                    acq_func="EI", acq_optimizer="auto",
-                    x0=None, y0=None, random_state=None, verbose=False,
-                    callback=None, n_points=10000, xi=0.01, kappa=1.96,
-                    n_jobs=1):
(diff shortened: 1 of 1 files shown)


betatim:
I think this is a good move but would like to hear @MechCoder or @glouppe's opinion.

The only thing to think about is should we keep the argument for two more releases and raise a deprecation warning before we remove it. Or do we just get rid of it because it was "wrong"?

MechCoder:
lets just remove.

MechCoder:
lgtm

MechCoder:
I assumed users supplying `forest_minimize(...., acq_optimizer="sampling")` ~= 0

betatim:
And another nice contribution!

PR #373, opened 2017-05-03
[MRG] Add 'copy constructor' to Space

### LINKED ISSUE #371: Optimizing over a list defining points in the search space?
Is there a feature where, instead of using dimensions to define the search space, skopt will optimize on a list of points which define the search space?

This would be a useful feature if I wanted to optimize in a subset of the possible points of a space (rather than the entire space).

betatim:
Fixes #371 

Create a new Space instance from an existing one. The goal of this PR is to allow people to pass a `Space` instance everywhere we accept a list of dimensions as argument.

### DIFF
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -148,6 +148,27 @@ def test_minimizer_api_random_only(minimizer):
     check_minimizer_bounds(result, n_calls)
 
 
+@pytest.mark.parametrize("minimizer", MINIMIZERS)
+def test_minimizer_with_space(minimizer):
+    # check we can pass a Space instance as dimensions argument
+    n_calls = 7
+    n_random_starts = 4
+
+    space = Space([(-5.0, 10.0), (0.0, 15.0)])
+    space_result = minimizer(branin, space, n_calls=n_calls,
+                             n_random_starts=n_random_starts, random_state=1)
+
+    check_minimizer_api(space_result, n_calls)
+    check_minimizer_bounds(space_result, n_calls)
+
+    dimensions = [(-5.0, 10.0), (0.0, 15.0)]
+    result = minimizer(branin, dimensions, n_calls=n_calls,
+                       n_random_starts=n_random_starts, random_state=1)
+
+    assert np.allclose(space_result.x_iters, result.x_iters)
+    assert np.allclose(space_result.func_vals, result.func_vals)
+
+
 @pytest.mark.parametrize("n_random_starts, optimizer_func",
                          product([0, 5],
                                  [gp_minimize,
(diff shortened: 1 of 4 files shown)


betatim:
@MechCoder do you have an idea where I should go and check for this failure:

```python
(Pdb) p space_result.x_iters
[[1.2553300705386103, 10.804867401632373], [-4.9982843777398269, 4.5349885894775968], [-2.798661637743304, 1.3850789215319672], [-2.2060968293349363, 5.1834109056457169], [1.266321887020273, 10.806829307463431], [-2.2779669130728766, 5.1862295242560386], [-2.0526999728871478, 5.1772280358548404]]
(Pdb) p result.x_iters
[[1.2553300705386103, 10.804867401632373], [-4.9982843777398269, 4.5349885894775968], [-2.798661637743304, 1.3850789215319672], [-2.2060968293349363, 5.1834109056457169], [1.2631908785389683, 10.79850547188739], [-1.9719421658503988, 5.3208778886226966], [-1.8862238180288384, 7.3176360832275558]]
```

The test in `test_common.py` fails because for `gp_minimize` the objective gets evaluated at different points. The first four points are random samples and then we get slight differences. Not quite sure why. A random state that isn't passed on properly??

MechCoder:
https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/optimizer.py#L312 :(

MechCoder:
Could you please add a common test for reproducibility?

betatim:
Will add a test. Thanks for finding this.

skopt/tests/test_common.py
```diff
@@ -148,6 +148,47 @@ def test_minimizer_api_random_only(minimizer):
...
+
+    space = [(-5.0, 10.0), (0.0, 15.0)]
+    result1 = minimizer(branin, space, n_calls=n_calls,
+                        n_random_starts=n_random_starts, random_state=1)
+
+    dimensions = [(-5.0, 10.0), (0.0, 15.0)]
+    result2 = minimizer(branin, dimensions, n_calls=n_calls,
+                        n_random_starts=n_random_starts, random_state=1)
+
+    assert np.allclose(result1.x_iters, result2.x_iters)
```
MechCoder: You can also use `assert_array_almost_equal` over here.

skopt/tests/test_space.py
```diff
@@ -252,6 +252,16 @@ def test_space_api():
...
 
 
+def test_space_from_space():
+    # can you pass a Space instance to the Space constructor?
+    space = Space([(0.0, 1.0), (-5, 5),
+                   ("a", "b", "c"), (1.0, 5.0, "log-uniform"), ("e", "f")])
+
+    space2 = Space(space)
+
+    assert space == space2
```
MechCoder: I think it is a good idea to use `assert_equal` over here.

MechCoder:
LGTM

betatim:
Switched both of the assert statements.

Do you know what the advantage/disadvantages are? Seems like in pytest everything (or most things) are written with simple assert statements (based on looking around what others do).

MechCoder:
I do it just for prettier error messages.

PR #375, opened 2017-05-19
Typo, "f of may be noisy"  → "of f may be noisy"

Naereen:
Typo, "f of may be noisy"  → "of f may be noisy" in https://scikit-optimize.github.io/notebooks/bayesian-optimization.html


MechCoder:
thx!

PR #376, opened 2017-05-22
[MRG] Add scikit-garden as a dependency

glouppe:


glouppe:
Travis is happy :)

betatim:
You want to remove cython from the requirements again :) ?

glouppe:
Haha, yeah, except this was needed to get skgarden to compile :x (hence the addition for travis and circle ci)

betatim:
Weird, shouldn't installing skgarden install cython?

Related: should we also move the GBRT wrapper?

MechCoder:
Yeah, it should. I did not add `install_requires` to `setup.py` :/

glouppe:
I have now added the necessary requirements in scikit-garden's setup files, such that cython gets automatically installed from there if absent. We should still include cython in Travis and CircleCI scripts, to speed up things. 

So I guess this is ready for merge once we release 0.2 for scikit-garden.

MechCoder:
I pushed 0.1.2

build_tools/circle/install.sh
```diff
@@ -30,9 +30,11 @@ popd
 # Configure the conda environment and put it in the path using the
 # provided versions
 conda create -n testenv --yes python pip pytest nose \
-   numpy scipy scikit-learn matplotlib
+   numpy scipy scikit-learn matplotlib cython
 source activate testenv
 
+pip install -r requirements.txt
```
betatim: do we need this? Especially after #383 where we switched to using `pip install -e.` I think we should be good without this. (same in travis)

betatim:
Need to add the new dependency in `setup.py` as well. Python is good for DRY, except when it comes to specifying dependencies ...

glouppe:
Let's see how happy Travis is this morning :)

MechCoder:
lgtm

glouppe:
Merging :)

PR #381, opened 2017-05-30
[MRG] Add test to check we fit a model when a point is provided

betatim:
Fixes the bug I found in https://github.com/scikit-optimize/scikit-optimize/pull/343#issuecomment-304805380

Seems like we only fit a model if you call `tell` after exhausting the random points, not if you call it in between and then never again.

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -7,6 +7,7 @@ It is sufficient that one re-implements the base estimator.
 import copy
 import inspect
 import numbers
+import warnings
 from collections import Iterable
 
 import numpy as np
@@ -18,7 +19,7 @@ from ..utils import eval_callbacks
 
 
 def base_minimize(func, dimensions, base_estimator,
-                  n_calls=100, n_random_starts=10,
+                  n_calls=100, n_initial_points=10, n_random_starts=None,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
@@ -53,9 +54,12 @@ def base_minimize(func, dimensions, base_estimator,
     * `n_calls` [int, default=100]:
         Maximum number of calls to `func`.
 
-    * `n_random_starts` [int, default=10]:
+    * `n_initial_points` [int, default=10]:
         Number of evaluations of `func` with random initialization points
         before approximating the `func` with `base_estimator`.
+
+    * `n_random_starts` [int, default=10]:
+        DEPRECATED, use `n_initial_points` instead.
 
     * `acq_func` [string, default=`"EI"`]:
         Function to minimize over the posterior distribution. Can be either
@@ -162,8 +166,16 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
-    optimizer = Optimizer(dimensions, base_estimator, n_random_starts,
-                          acq_func, acq_optimizer, random_state,
+    if n_random_starts is not None:
+        warnings.warn(("n_random_starts has been renamed to "
+                       "n_initial_points."),
+                      DeprecationWarning)
+        n_initial_points = n_random_starts
+
+    optimizer = Optimizer(dimensions, base_estimator,
+                          n_initial_points=n_initial_points,
+                          acq_func=acq_func, acq_optimizer=acq_optimizer,
+                          random_state=random_state,
                           acq_optimizer_kwargs=acq_optimizer_kwargs,
                           acq_func_kwargs=acq_func_kwargs)
 
@@ -184,21 +196,21 @@ def base_minimize(func, dimensions, base_estimator,
         raise ValueError("`x0` should be a list, but got %s" % type(x0))
 
     n_init_func_calls = len(x0) if y0 is None else 0
-    n_total_init_calls = n_random_starts + n_init_func_calls
+    n_total_init_calls = n_initial_points + n_init_func_calls
 
     if n_calls < n_total_init_calls:
         raise ValueError(
             "Expected `n_calls` >= %d, got %d" % (n_total_init_calls,
                                                   n_calls))
 
-    if n_random_starts == 0 and not x0:
-        raise ValueError("Either set `n_random_starts` > 0,"
+    if n_initial_points == 0 and not x0:
+        raise ValueError("Either set `n_initial_points` > 0,"
                          " or provide `x0`")
 
     callbacks = check_callback(callback)
     if verbose:
         callbacks.append(VerboseCallback(
-            n_init=n_init_func_calls, n_random=n_random_starts,
+            n_init=n_init_func_calls, n_random=n_initial_points,
             n_total=n_calls))
 
     # User suggested points at which to evaluate the objective first
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -196,8 +196,7 @@ def base_minimize(func, dimensions, base_estimator,
                           acq_optimizer_kwargs=acq_optimizer_kwargs,
                           acq_func_kwargs=acq_func_kwargs)
 
-    if optimizer.space.n_dims == 1:
-        assert all(isinstance(p, Iterable) for p in x0)
+    assert all(isinstance(p, Iterable) for p in x0)
 
     if not all(len(p) == optimizer.space.n_dims for p in x0):
         raise RuntimeError("Optimization space (%s) and initial points in x0 "
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -7,7 +7,6 @@ It is sufficient that one re-implements the base estimator.
 import copy
 import inspect
 import numbers
-import warnings
 from collections import Iterable
 
 import numpy as np
@@ -19,7 +18,7 @@ from ..utils import eval_callbacks
 
 
 def base_minimize(func, dimensions, base_estimator,
-                  n_calls=100, n_initial_points=10, n_random_starts=None,
+                  n_calls=100, n_random_starts=10,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
@@ -54,14 +53,9 @@ def base_minimize(func, dimensions, base_estimator,
     * `n_calls` [int, default=100]:
         Maximum number of calls to `func`.
 
-    * `n_initial_points` [int, default=10]:
-        Number of evaluations of `func` with initialization points
-        before approximating it with `base_estimator`. Points provided as
-        `x0` count as initialization points. If len(x0) < n_initial_points
-        additional points are sampled at random.
-
     * `n_random_starts` [int, default=10]:
-        DEPRECATED, use `n_initial_points` instead.
+        Number of evaluations of `func` with random points before
+        approximating it with `base_estimator`.
 
     * `acq_func` [string, default=`"EI"`]:
         Function to minimize over the posterior distribution. Can be either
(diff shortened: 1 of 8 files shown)


betatim:
@glouppe at some point we decided not to fit any models until after we exhausted the random calls. Do you remember why?

Now I am thinking we should fit a model every time `tell()` is called and use it to produce a `next_x` that is used as soon as the random calls are exhausted.

glouppe:
> @glouppe at some point we decided not to fit any models until after we exhausted the random calls. Do you remember why?

Why shall we fit?

Not sure if this is the current API, but I would fit only once `n_random_starts` points have been `tell`'ed, and do nothing before that. Meanwhile, calling `ask` would return random points as long as this condition is not satisfied. What do you think?

betatim:
The reason to do a fit is because otherwise tests like the one I added in this PR fail. I don't think it should fail.

Returning random points until we have been told `n_random_starts` points is a nice idea. Not sure if the current API forbids it or not. (My guess would be that this is a edge case not many people run into anyway)

betatim:
7e784a5 implements this strategy. Let's see which tests start failing.

MechCoder:
I also believe that the current behaviour is correct, i.e  not fitting models till `n_random_starts` has been exhausted. The assumption being you need atleast `n_random_starts` to get a good approximation of the search space to suggest the next point.

betatim:
:+1: 

Next question I ran into: what if I add a few points by hand? I think following our logic we should keep suggesting random points until `yi` contains at least `n_random_starts` points.

(Maybe one day we want to rename `n_random_starts` to something that reflects this but that is a different conversation)

MechCoder:
I thought about that and I believe that is desired behaviour as well.

We should change a few things that we missed.

1. Stop decreasing `_n_random_starts` at `ask` time. It is not guaranteed that the same points at `ask` time will be "told".
2. Rename `n_random_starts` to "something-better" which suggests the first model will be fit after "something-better" number of points are passed to `tell`

MechCoder:
I admit I did not read the entire thread. It seems that there is consensus.

skopt/optimizer/optimizer.py
```diff
@@ -109,7 +109,7 @@ class Optimizer(object):
 
     """
     def __init__(self, dimensions, base_estimator,
-                 n_random_starts=10, acq_func="gp_hedge",
+                 n_random_starts=10, n_initial_points=10, acq_func="gp_hedge",
```
MechCoder: We can rename this to `n_starts`?

skopt/optimizer/optimizer.py
```diff
@@ -148,21 +148,27 @@ def __init__(self, dimensions, base_estimator,
                 self._cat_inds.append(ind)
             else:
                 self._non_cat_inds.append(ind)
-        self._check_arguments(base_estimator, n_random_starts, acq_optimizer)
+
+        if n_random_starts != 10:
```
MechCoder: A practice is to change the class instantiation to `n_random_starts=None`, so that we raise a warning when Optimizer(n_random_starts=10, ....) also raises a warning.

skopt/optimizer/optimizer.py
```diff
@@ -50,9 +50,9 @@ class Optimizer(object):
         argument, which returns `std(Y | x)`` along with `E[Y | x]`.
 
     * `n_random_starts` [int, default=10]:
-        Number of evaluations of `func` with random initialization points
-        before approximating the `func` with `base_estimator`. While random
-        points are being suggested no model will be fit to the observations.
+        Number of evaluations of `func` with random points before approximating
```
MechCoder: We should remove this parameter from the documentation.

skopt/tests/test_optimizer.py
```diff
@@ -105,3 +105,16 @@ def test_acq_optimizer(base_estimator):
...
+
+def test_exhaust_random_calls():
+    # check a model is fitted if there is at least one call to tell
+    base_estimator = ExtraTreesRegressor(random_state=2)
+    opt = Optimizer([(-2.0, 2.0)], base_estimator, n_random_starts=2,
+                    acq_optimizer="sampling")
+
+    x = opt.ask()  # random point
+    opt.tell(x, 3.)
+    opt.ask()  # random point
```
MechCoder: Is this is a smoke test? I would prefer to test two things.

1. Repeated calls to ask give different points.
2. `self.models` etc is empty till `n_points` are told.

MechCoder:
The tests fail because of the way `Optimizer` is instantiated in the `base_minimize`

betatim:
Yeah there is quite a bit missing from this still I think, worked on it on the train yesterday and haven't pushed it yet.

skopt/optimizer/optimizer.py
betatim: Don't really like `n_starts`, it doesn't tell me "what the parameter is about". `n_initial_points` is very long to type. Maybe there is a third name we can use?

MechCoder: `n_init`?

skopt/tests/test_optimizer.py
betatim: This is the usecase that lead to an error before this PR. Most of 2. is already checked (in a fairly complex way :-/) in `test_common.py`. I'll have a look if we test 1. anywhere already or not.

betatim:
Ready for review. Questions to discuss:

* `n_initial_points` is a long name, what is a better name?

skopt/optimizer/optimizer.py
```diff
@@ -50,9 +50,12 @@ class Optimizer(object):
         argument, which returns `std(Y | x)`` along with `E[Y | x]`.
 
     * `n_random_starts` [int, default=10]:
-        Number of evaluations of `func` with random initialization points
-        before approximating the `func` with `base_estimator`. While random
-        points are being suggested no model will be fit to the observations.
+        DEPRECATED, use `n_initial_points` instead.
+
+    * `n_initial_points` [int, default=10]:
+        Number of evaluations of `func` with random points before approximating
```
MechCoder: These points need not be `random` anymore.

skopt/optimizer/optimizer.py
```diff
@@ -179,11 +192,11 @@ def _check_arguments(self, base_estimator, n_random_starts, acq_optimizer):
     def ask(self):
         """Suggest next point at which to evaluate the objective.
 
-        Returns a random point for the first `n_random_starts` calls, after
-        that `base_estimator` is used to determine the next point.
+        Return a random point while not at least `n_random_starts` observations
```
MechCoder: `n_random_starts` -> `n_initial_points`

MechCoder:
My suggestion is to keep the `n_random_starts` parameter as it is in the `*_minimize` functions, while changing it only in the `Optimizer` class. The reason being that `n_random_starts` excludes the initial points `x0` etc in the minimize functions, while in the Optimizer the points told can be non-random and can be based on some prior-knowledge as well.

skopt/tests/test_optimizer.py
MechCoder: ok, the comment below states that # this is the first non-random points. I think it is sill random since there is only call to `tell`

betatim:
I added some more changes. They fix the comments made and (try to) simplify the code a bit. Overall I think we should call it `n_initial_points` everywhere (`*_minimize` and `Optimizer`). They behave the same (if not I made a mistake). My thinking is that the only reason to sample points at random is because you do not know any better. From this it follows that if you use `x0` to provide initial points then we should reduce the number of random points accordingly. 

The parameter `n_calls` sets the total number of times `func` can be called. So if we need to use some of these calls to evaluate initial points then we reduce the number of iterations in the bayesian optimisation loop.

This is what I tried to implement. Does it sound sensible?

MechCoder:
To summarize the changes:

1. In the `Optimizer` API, multiple calls to `ask` when `n_random_starts` are not exhausted, used to decrease the attribute the `n_random_starts` counter. That behaviour has been changed to decrease the `n_random_starts` counter when a point is being `tell`ed. This is a bug fix.
2. The name `n_random_starts` has been changed to `n_initial_points` in the Optimizer API.
3. In the `*_minimize` functions, however I feel that the current behaviour is well defined and was arrived after some amount of discussion, `x0` is the number of initial points and `n_random_starts` should be the number of random points sampled in the space excluding `x0`. What is your rationale behind changing this behaviour? In any case, it would be great if you remove this change to a separate PR for discussion and we can merge the changes that involve the first 2 points quickly.

betatim:
The thinking is that `n_random_starts` means the same thing in both cases `*_minimize` and `Optimiser`. Namely "we need some points to get this started and I have no good suggestion for where to evaluate `func`, so please pick some at random."

We could use the convention that in `*_minimise` you ask for random points in addition to the points you added via `x0` by setting `n_random_points`. At the moment I think it would make it more confusing to have these different names. And it makes the code a lot more confusing as you now need to somehow convert from the one convention to the other in `base_minimize`.

betatim:
Actually maybe we can think of `n_random_starts` for the `*_minimize` functions as number of random points to sample in addition to any initial points. (Though maybe we should rename it from `_starts` to `_points`?)

MechCoder:
> Actually maybe we can think of n_random_starts for the *_minimize functions as number of random points to sample in addition to any initial points.

Honestly, that is what I've been trying to explain since the beginning, both over  [here](https://github.com/scikit-optimize/scikit-optimize/pull/381#issuecomment-305783038) and [here](https://github.com/scikit-optimize/scikit-optimize/pull/381#issuecomment-306044596) but I agree that the doctring is slightly confusing as well. It should be number of approximations to the function using random points excluding `x0`:)

MechCoder:
> The thinking is that n_random_starts means the same thing in both cases *_minimize and Optimiser. Namely "we need some points to get this started and I have no good suggestion for where to evaluate func, so please pick some at random."

That is not completely true. In the `Optimizer`, the `tell` is controlled by the user, one can choose to ignore the random point returned by the `ask` method and supply any other point. This means `n_random_starts` is a bad name since all the points need not be randomly generated.

As said before such an argument does not apply to the minimize functions since the `n_random_starts` points are all randomly generated, (or at least before it was refactored into the Optimizer API)

betatim:
> That is not completely true. ...

You wanted to hear my thinking, so I told you about it. Last time I checked with my brain it was completely true that I was indeed thinking this 😃 

I think a large part of the confusion was that you were talking about how it is right now and I was talking about how I think it should be in the future.

---

Anyway, I modified the comment for `*_minimize` to say "additional random points".

betatim:
Tests pass again and changed back to the old parameter names for `*_minimize`.

betatim:
Maybe @iaroslav-ai has time for some reviewing?

skopt/optimizer/base.py
```diff
@@ -162,63 +162,68 @@ def base_minimize(func, dimensions, base_estimator,
...
+    n_initial_points = n_random_starts + len(x0)
+    optimizer = Optimizer(dimensions, base_estimator,
+                          n_initial_points=n_initial_points,
+                          acq_func=acq_func, acq_optimizer=acq_optimizer,
+                          random_state=random_state,
+                          acq_optimizer_kwargs=acq_optimizer_kwargs,
+                          acq_func_kwargs=acq_func_kwargs)
+
+    if optimizer.space.n_dims == 1:
+        assert all(isinstance(p, Iterable) for p in x0)
```
MechCoder: Why is this special-cased for `space.n_dims=1`?

skopt/optimizer/optimizer.py
```diff
@@ -151,7 +155,14 @@ def __init__(self, dimensions, base_estimator,
                 self._cat_inds.append(ind)
             else:
                 self._non_cat_inds.append(ind)
-        self._check_arguments(base_estimator, n_random_starts, acq_optimizer)
+
+        if n_random_starts is not None:
+            warnings.warn(("n_random_starts has been renamed to "
```
MechCoder: It's not just a rename right. Maybe we can rename this to `n_random_starts will be removed. Use n_initial_points instead`?

skopt/tests/test_optimizer.py
```diff
@@ -112,3 +112,31 @@ def test_acq_optimizer(base_estimator):
...
+    # check a model is fitted and used to make suggestions after we added
+    # at least n_initial_points via tell()
+    base_estimator = ExtraTreesRegressor(random_state=2)
+    opt = Optimizer([(-2.0, 2.0)], base_estimator, n_initial_points=2,
+                    acq_optimizer="sampling", random_state=1)
+
+    x0 = opt.ask()  # random point
+    x1 = opt.ask()  # random point
+    assert x0 != x1
+    print("tell1")
```
MechCoder: Can you make these print statements into inline comments?

MechCoder:
LGTM apart from some cosmetic comments.

iaroslav-ai:
Yup will take a look

skopt/optimizer/base.py
betatim: Good question. Removing the special casing.

skopt/tests/test_common.py
```diff
@@ -197,12 +197,12 @@ def test_minimizer_with_space(minimizer):
...
 @pytest.mark.slow_test
 @pytest.mark.parametrize("n_random_starts, optimizer_func",
-                         product([0, 5],
+                         product([0, 1, 2, 3, 4, 5],
                                  [gp_minimize,
                                   forest_minimize,
                                   gbrt_minimize]))
 def test_init_vals_and_models(n_random_starts, optimizer_func):
-    # test how many models are fitted when using initial points, values
+    # test how many models are fitted when using initial points, y0 values
```
iaroslav-ai: Maybe a bit more clear would be "initial points and corresponding objectives", otherwise this might be confusing to someone. Also maybe something like "different number of initial random starts"
Edited: saw the other test.

betatim:
Merge once circleci is happy?

skopt/tests/test_common.py
betatim: Let's keep this in mind for the next time we pass by this test (or nearby).

MechCoder:
Great, thanks for the patience @betatim!

betatim:
🌴 this is like live on a tropical island :)

PR #383, opened 2017-05-30
[MRG] Convert README from md to rst

### LINKED ISSUE #382: Rewrite README.md using restructured text so that it renders nicely on the PyPi page
For eg: https://pypi.python.org/pypi/requests

Credits: @Naereen from https://github.com/MechCoder/mondrian-art/commit/fcb639dbbc055fdb41e8ffc283a18856dbc637be#commitcomment-22325181

betatim:
Fixes #382 

This is to support nicer rendering on PyPI which does not understand markdown.

Github doesn't seem to mind if you use MD or RST, and we don't need to edit it often.

PyPI page will get updated with the next release.


betatim:
Needed to fix some formatting issues, so would be good if someone takes another look at the rendered version: https://github.com/betatim/scikit-optimize/blob/c39dc0032890a8a3693e6785ac2fa7af3621c306/README.rst otherwise ready

mirca:
Under the section ``Development``, I would just change
```
pip install -r requirements
python setup.py develop
```

to ``pip install -e .`` based on the section ``"Editable" Installs`` of https://pip.pypa.io/en/stable/reference/pip_install/

MechCoder:
thanks!

Naereen:
Great! Looking forward for a nicer [PyPi](https://pypi.python.org/pypi/scikit-optimize) page!

PR #384, opened 2017-05-31
Benchmarks for logistic regrssion on mnist

MechCoder:


betatim:
First impression: do we really need `keras` to do logistic regression? I'd prefer if we could do the same using scikit-learn :-/

Can we modify the output/add a parameter so that it outputs the same quantities as the other benchmark scripts so we can add to the overview table?

betatim:
Can you post the plots it produces as well :)

MechCoder:
Then how can I showcase my deep-learning skills?

More seriously, we can use the `SGDClassifier` with `log-loss` but it does not allow to tune the batch-size. automatically (the use-case being more of online-learning). We can allow it to by doing something like this.

```python
clf = SGDClassifier(loss="log", n_iter=1, shuffle=False, ....)
shuffle(X)
for X_batch, y_batch in generate_batches(X, y, batch_size=):
    clf.partial_fit(X_batch, y_batch)
```

But ultimately we need keras if we want to bench convnets etc, so why not use it now?

benchmarks/logreg_on_mnist.py
```diff
@@ -0,0 +1,122 @@
...
+        opt = forest_minimize
+
+    min_vals = []
+    all_vals = []
+    all_times = []
+    timer = TimerCallback()
+    res = opt(
+        log_reg_on_mnist, bounds,
+        verbose=1, random_state=random_state, callback=timer)
+    del res["specs"]
```
betatim: does the pickling fail with specs in the result? seems like something we should explain in a comment/fix

betatim:
> Then how can I showcase my deep-learning skills?

by contributing to keras :rofl: 

I'm just a fan of not adding dependencies, is all.

More important I think is that we get the output formatting so we can easily add it to https://github.com/scikit-optimize/scikit-optimize/blob/master/benchmarks/README.md and then we might have to start a new issue to actually discuss how to tune our defaults, etc

MechCoder:
Hmm, the benchmarks are independent of the actual-package so we need not add it as a dependency. We can just add a comment saying that if you need to run this benchmark, we need keras installed.

I'll change the code a bit, right now it runs only for one random state so that I can run jobs across multiple processes and aggregate the results.

iaroslav-ai:
An idea that I had for some time: how about creating a separate repo for scikit-optimize benchmarks (maybe something not original like scikit-optimize-benchmarks)? Then we could go with all kinds of wild dependencies there, and skopt will not be affected. Could also contain some datasets, so that we do not worry about broken links or mldata being down all of a sudden. What do you think?

MechCoder:
Yes, that is a good idea, but the problem is running all these benchmarks and caching the data. Random search seems to perform well for this particular problem and we need cifar10 that takes a significantly longer time as compared to mnist.

iaroslav-ai:
I was thinking about something like we train a few times on cifar10 for example, and hence collect data about points and correpsonding objectives, and then either train predictive model to estimate what the objective would be for a point, or just train a surrogate from scikit optimize. Such surrogate should be very fast to evaluate, and with sufficient amount of data I expect would represent accurately behaviour of the actual objective.

iaroslav-ai:
See also discussion in #458. Not sure if someone will pick this up, so closing this for now (but feel free to reopen!).

Issue #293, opened 2017-01-09
Faster tests by deselcting slow tests


betatim:
Running the complete set of tests takes a long time. We can use [test attributes](http://nose.readthedocs.io/en/latest/plugins/attrib.html) to mark tests that take a long time to run, this gives people a chance to run a lightweight test quickly by deselecting those tests. Travis would still run all tests.

ccauet:
@betatim Implemented by #386

PR #386, opened 2017-05-31
Implements pytest attributes to select fast and slow running tests

ccauet:
This PR implements changes proposed in #293 by adding pytest attributes to all tests depending on their run time.

I picked the threshold between slow and fasts tests to be at 1 second. This yields all fast tests to run in approx 10s while the slow tests take another 10min. If one test is performed multiple times with different input parameters, only the slowest run is taken into account.

* To only run the fast tests: `pytest -m 'fast_test'`
* To only run the slow tests: `pytest -m 'slow_test'`
* To only run tests w/o any attribute: `pytest -m 'not slow_test and not fast_test'`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,7 +79,11 @@ The development version can be installed through:
     cd scikit-optimize
     pip install -e.
 
-Run the tests by executing ``pytest`` in the top level directory.
+Run all tests by executing `pytest` in the top level directory.
+
+To only run the subset of tests with low run time, you can use `pytest -m 'fast_test'`. On the other hand `pytest -m 'slow_test'` is also possible. To exclude all slow running tests try `pytest -m 'not slow_test'`.
+
+This is implemented using pytest [attributes](https://docs.pytest.org/en/latest/mark.html). If a tests runs longer than 1 second, it is marked as slow, else as fast.
 
 All contributors are welcome!
 
--- a/skopt/learning/gaussian_process/tests/test_gpr.py
+++ b/skopt/learning/gaussian_process/tests/test_gpr.py
@@ -1,4 +1,6 @@
 import numpy as np
+import pytest
+
 from scipy import optimize
 
 from sklearn.utils.testing import assert_almost_equal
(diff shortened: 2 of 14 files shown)


MechCoder:
lgtm

glouppe:
Thanks for the changes! Could you rebase your branch on top of master?

betatim:
Nice!

If I run `pytest` it will run all tests right? So to specifically exclude slow stuff I could do `pytest -m 'not slow_test'` :racing_car:

ccauet:
> If I run pytest it will run all tests right? So to specifically exclude slow stuff I could do pytest -m 'not slow_test'

Correct.

ccauet:
> Could you rebase your branch on top of master?

@glouppe I did, but to be honest I don't understand what is happening here. I tried following: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

At first things seemed to work as expected. I added an upstream remote, then `git fetch upstream` + `git rebase upstream/master`. I saw the recent changes to the upstream master, i.e. `README.md` was replaced by `README.rst` and the `test_forest.py` was removed. I applied my changes to the `README.rst` file and then tried to push everything. 

This was not possible until another pull from my forked repo. Here I had to apply the same changes again, then push. 

This updated the PR as you can see, but still left the `README.rst` with a merge conflict. 

I then did the whole procedure again, still the conflict in `README.rst` is present...

MechCoder:
You need to submit pull requests from branches other than master to make it work. However, since there is only one commit this might work assuming your upstream points to `scikit-optimize/scikit-optimize`.

```
# Create backup just in case
git checkout backup

# Resests master to upstream
git checkout master
git reset --hard upstream/master

# Pick necessary commits and push.
git cherry-pick 75885b4
git push -f origin master
```

ccauet:
@MechCoder Ok, this seems to work. Sorry for the mess and thanks for the help.

MechCoder:
thanks!

Issue #78, opened 2016-06-06
Add scikit-learn compatible BayesSearchCV


amueller:
Hey.
Do you intent to provide a GridSearchCV plug-in replacement or only the optimizer?
The thing is that it might take a while to get that into scikit-learn, and it would be nice if people had access to it.

Cheers,
Andy

glouppe:
Would be a nice thing to have indeed! There is already a rudimentary example at https://github.com/scikit-optimize/scikit-optimize/blob/master/examples/plot_random_vs_gp.py but it would be even better to wrap to into a `BayesSearchCV` object.

betatim:
We should take inspiration from the discussion and design in scikit-learn/scikit-learn#5491

betatim:
http://arxiv.org/abs/1603.06560 via http://www.argmin.net/2016/06/20/hypertuning/ so we just wrap `dummy_minimize`? IMHO we should do some benchmarking 📊 😀

glouppe:
Random search for parameters is already in scikit-learn :)

betatim:
I was practicing my sarcasm, before it becomes illegal after the 🇬🇧  votes to leave 💶  (really money is the only euro symbol we have? 😢)

betatim:
On a more serious note, reading the blog a bit more carefully I think what they found is that if you can run twice as many points when doing random sampling then that beats BO. Slightly 🍎 to 🍑  comparison.

amueller:
Why do you think so? It is just trying to put some perspective into what the baseline of hyperparameter search should be.
I think "how much better than random is state of the art smbo?" is a good question, and the answer is that it is between 1x and 2x faster, but in no setting they looked at was it more than 2x faster.

betatim:
I think it is an interesting investigation for sure. The way the paper is phrased it comes across a bit dismissive of BO. The advice "evaluate your expensive functions twice as often and you will get a better result" is a bit impractical if your function is expensive to evaluate which is presumably why you are bothering with something complex like BO. That is all.

glouppe:
#112 demonstrates how to use `skopt` for hyper-parameter optimisation. 

Might still be worth also providing something like a `SkoptSearchCV` object though, behaving like an actual scikit-learn estimator.

glouppe:
Bumping this. Having a `SkoptSearchCV` would be nice.

betatim:
Not sure about the name, but otherwise like the idea!

[...]

Issue #400, opened 2017-06-12
Normalize Space dimensions to (0, 1) in Optimizer


MechCoder:
It seems like we do not normalize the dimensions of the space in the ask-and-tell API. Noted by @cranmer in https://github.com/scikit-optimize/scikit-optimize/issues/399#issuecomment-307887516

betatim:
If the `estimator` is a GP then we need to always normalise to (0, 1)? Asking because using `Optimizer` directly probably means you are more of an expert who wants to do exactly what they want with as little interference from the library. At least we should allow such a way of working. How do we make both easy switching from `*_minimize` to `Optimizer` possible and let experts do expert things?

iaroslav-ai:
Just pitching in my 2 cents. 
Maybe using something like a pipeline as a surrogate could be a solution, for example:
```python
from skopt.learning import GaussianProcessRegressor
from skopt.space import Real
from skopt import Optimizer

import numpy as np

from sklearn.pipeline import Pipeline
from sklearn.preprocessing import MinMaxScaler

# modified Pipeline class to handle `return_std` etc argument for predict
class SurrogatePipeline(Pipeline):
    def predict(self, X, *args, **kwargs):
        Xt = X
        for name, transform in self.steps[:-1]:
            if transform is not None:
                Xt = transform.transform(Xt)
        return self.steps[-1][-1].predict(Xt, *args, **kwargs)

# optimizer that uses pipeline as a surrogate
opt = Optimizer(
    dimensions=[Real(-2.0, 2.0) for i in range(10)],
    base_estimator=SurrogatePipeline([
        ('Scaler', MinMaxScaler()),
        ('Surrogate', GaussianProcessRegressor()),
    ])
)

# example optimization
obj = lambda x: np.sum(x)

for i in range(32):
    p = opt.ask()
    opt.tell(p, obj(p))
    print(min(opt.yi))
```
Instead of `MinMaxScaler`, a dedicated scaler can be used that normalizes based on dimensions of a problem.  This should make it possible for experts to modify the pipeline if necessary to remove normalization. Maybe also nice in general in someone wants to have a complex surrogate pipeline.

MechCoder:
> How do we make both easy switching from *_minimize to Optimizer possible and let experts do expert things?

Good point. For starters, we can maybe support `base_estimator` to be strings and for those cases we can handle things similar to what we do in our minimize functions.

MechCoder:
[...]

PR #405, opened 2017-06-17
[MRG+1] Scikit - Optimize based GridSearchCV plug-in replacement 

iaroslav-ai:
A class as discussed in #78 under the tentative name of `SkoptSearchCV` (discussion on what might be a better name is welcome :) 

Minimalist usage example as of right now:

```python
from skopt.wrappers import SkoptSearchCV

opt = SkoptSearchCV(
    SVC(),
    [{
        'C': Real(1e-6, 1e+6, prior='log-uniform'),
        'gamma': Real(1e-6, 1e+1, prior='log-uniform'),
        'degree': Integer(1, 8),
        'kernel': Categorical(['linear', 'poly', 'rbf']),
    }],
    n_jobs=1, n_iter=32,
)

opt.fit(X_train, y_train)
print(opt.score(X_test, y_test))
```

If something is missing in todos let me know. 

- [x] add `skopt.wrappers` to `setup.py`
- [x] implement basic wrapper similar to `RandomizedSearchCV` using `BaseSearchCV`
- [x] support multiple search spaces as list of dicts, similar to `GridSearchCV`
- [x] support parallel search 
- [x] add draft of example usage in `sklearn-wrapper`
- [x] add all necessary data to `cv_results_`
- [x] add tests
- [x] fix the docstrings and add comments 
- [x] review and improve example usage


iaroslav-ai:
Also let me know if you do not like the location of the class (cuttently `skopt.wrappers`)

iaroslav-ai:
As of right now PR contains working Implementation of `*SearchCV` according to #78 and scikit-learn/scikit-learn#5491 .
Could someone review? I am not sure who would be the best to ask to do this, as all of you guys seemed to be pretty active about this in referenced issue and PR :)

betatim:
Thanks for this! Code, example, and tests all in one!

> Also let me know if you do not like the location of the class (cuttently `skopt.wrappers`)

I'm not mad keen on the name. I'd make it available at the top level like `*_minimize`. But then `skopt.SkoptSearchCV` is a bit doppelt gemoppelt. `SMBOSearchCV` is cryptic, ... will think a bit and see what others have used. I think once we know a good name we can use it for the class and replace `wrappers` with it as well.

betatim:
[Looks like we need to be smarter about our tests](https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/244194986) so that they can run in less time.

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
+from skopt import Optimizer
+from skopt.learning import GaussianProcessRegressor, ExtraTreesRegressor
+
+from sklearn.base import clone
+from sklearn.externals.joblib import Parallel, delayed, cpu_count
+import sklearn.model_selection._search as skms
+
+import numpy as np
+from collections import *
```
betatim: no stars please :)

skopt/tests/test_searchcv.py
```diff
@@ -0,0 +1,80 @@
...
+from skopt import Optimizer
+from skopt.benchmarks import branin
+import skopt.learning as sol
+
+from scipy.spatial.distance import pdist
+import pytest
+
+# Extract available surrogates, so that new ones are used automatically
+available_surrogates = [
+    getattr(sol, name) for name in sol.__all__
```
glouppe: I would rather  be explicit and write down the full list of surrogates, instead of relying on some black magic that could easily break down as we add/update things.

skopt/tests/test_searchcv.py
```diff
@@ -0,0 +1,80 @@
...
+available_surrogates += ["auto"]
+
+
+@pytest.mark.parametrize("surrogate", available_surrogates)
+@pytest.mark.parametrize("n_jobs", [1, -1])  # test in sequential and parallel
+def test_searchcv_runs(surrogate, n_jobs):
+    """
+    Tests whether the cross validation search wrapper around sklearn
+     models runs properly with available surrogates and with single
+     or multiple workers.
```
glouppe: indentation issue

skopt/wrappers/__init__.py
```diff
@@ -0,0 +1 @@
+from .search_cv import SkoptSearchCV
```
glouppe: missing line break

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+        -------
+        optimizer: Instance of the `Optimizer` class used for for search
+            in some parameter space.
+
+        """
+        # convert search space from dict to list
+        dimensions = [params_space[k] for k in sorted(params_space.keys())]
+
+        if self.surrogate == "auto":
+            surrogate = GaussianProcessRegressor()
```
glouppe: Is that a good kernel? 
I think it would be nice to solve #338 to reuse a sensible default here.

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
+from skopt import Optimizer
+from skopt.learning import GaussianProcessRegressor, ExtraTreesRegressor
+
+from sklearn.base import clone
+from sklearn.externals.joblib import Parallel, delayed, cpu_count
+import sklearn.model_selection._search as skms
```
glouppe: can we find a better name than `skms`?

skopt/tests/test_searchcv.py
```diff
@@ -0,0 +1,80 @@
...
+from skopt.benchmarks import branin
+import skopt.learning as sol
+
+from scipy.spatial.distance import pdist
+import pytest
+
+# Extract available surrogates, so that new ones are used automatically
+available_surrogates = [
+    getattr(sol, name) for name in sol.__all__
+    if "GradientBoostingQuantileRegressor" not in name
```
betatim: Why can't we use this model?

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+        (dict, int > 0). The keys of every dictionary are parameter
+        names (strings) and values are skopt.space.Dimension instances
+        (Real, Integer or Categorical) which represents search space
+        for particular parameter.
+        If a list of dictionary objects is given, then the search is
+        performed sequentially for every parameter space with maximum
+        number of evaluations set to self.n_iter. Alternatively, if
+        a list of (dict, int > 0) is given, the search is done for
+        every search space for number of iterations given as a second
+        element of tuple.
```
betatim: This means if I give a list of three dictionaries there will be a total of `3*n_iter` iterations? What is scikit-learn's behaviour wrt this?

I thought sklearn treats the situation of having a list of `dict`s as having an extra implied categorical dimension.

We should try and be as much plug'n'play as possible.

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+                 pre_dispatch='2*n_jobs', random_state=None,
+                 error_score='raise', return_train_score=True):
+        self.search_spaces = search_spaces
+        self.n_iter = n_iter
+        self.random_state = random_state
+        self.surrogate = surrogate
+
+        # this dict is used in order to keep track of skopt Optimizer
+        # instances for different search spaces (str(space) is used as key)
+        self.optimizer = {}
```
betatim: -> `_optimizer`? It is 'private' to the internal process no?

skopt/wrappers/search_cv.py
betatim: We should solve https://github.com/scikit-optimize/scikit-optimize/issues/400#issuecomment-309020353 or we need to add the same kernel setup and space conversion as is in `gp_minimize`

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+
+        # account for case n_jobs < 0
+        if n_jobs < 0:
+            n_jobs = max(1, cpu_count() + n_jobs + 1)
+
+        for elem in self.search_spaces:
+
+            # if tuple: (dict: search space, int: n_iter)
+            if isinstance(elem, tuple):
+                psp, n_iter = elem
```
betatim: `psp` -> `search_space` (I think) in general we should use proper words for the variable names instead of abbreviations

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+            # if dict: represents search space
+            elif isinstance(elem, dict):
+                psp, n_iter = elem, self.n_iter
+            else:
+                raise ValueError("Unsupported type of parameter space. "
+                                 "Expected tuple or dict, got " + str(elem))
```
betatim: switch to `"got %s." % (elem)`

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+
+            # do the optimization for particular search space
+            while n_iter:
```
betatim: `bool(-3) == True` so this won't stop if `n_iter` isn't dividable by `n_jobs`. I think we should check if there are enough iterations left to sample `n_jobs` more points, if not reduce it to the number left and then stop the loop. (and add a unit test to check this works)

skopt/wrappers/search_cv.py
```diff
@@ -0,0 +1,490 @@
...
+        Returns
+        -------
+        params_dict: dictionary with parameter values.
+        """
+        params_dict = {
+            k: v for k, v in zip(sorted(params_space.keys()), params)
+        }
+        return params_dict
+
+    def step(self, X, y, param_space, groups=None, n_jobs=1):
```
betatim: I'd have to read the code for `RandomSearchCV` but doesn't it have a step function we can re-use/hook into so we don't have to duplicate all the code for recording the results?

skopt/wrappers/search_cv.py
iaroslav-ai: The documentation for `GridSearchCV` says that if the list of dicts is provided, then grid spanned by every dict is explored sequentially. Hence I do the same in here.

iaroslav-ai:
Addressed most of comments. I use now `_fit` method from `BaseSearchCV` inside step function, which simplifies stuff a bit, but might be a bit hacky (because _fit resets results_cv_). Also I renamed the class to BayesSearchCV, as per discussion in scikit-learn/scikit-learn#5491 . If someone has an idea for better name let me know.

skopt/wrappers/search_cv.py
iaroslav-ai: Yup I use now strings for surrogate for Optimizer which cooks me estimators like a chef :stuck_out_tongue_winking_eye:

iaroslav-ai:
Just double checked - this should address all the comments. Reduced the tests a bit - only one estimator class is tested, but generalization to other estimators is tested with the consistency of estimators interface in other tests. Also more general `optimizer_kwargs` is used instead of `base_estimator`.

iaroslav-ai:
Please take another look and tell me what you think.

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+import sklearn.model_selection._search as sk_model_sel
+
+import numpy as np
+from collections import defaultdict
+
+
+class BayesSearchCV(sk_model_sel.BaseSearchCV):
+    """Bayesian optimization over hyper parameters.
+
+    SkoptSearchCV implements a "fit" and a "score" method.
```
betatim: -> BayesSearchCV

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+        self.random_state = random_state
+
+        if optimizer_kwargs is None:
+            self.optimizer_kwargs = {}
+        else:
+            self.optimizer_kwargs = optimizer_kwargs
+
+        # this dict is used in order to keep track of skopt Optimizer
+        # instances for different search spaces (str(space) is used as key)
+        self.optimizer_ = {}
```
betatim: I like this as a way to support sub-search spaces (link #108 for reference). Maybe we can use this idea to build #108.

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+        # convert search space from dict to list
+        dimensions = [params_space[k] for k in sorted(params_space.keys())]
+
+        kwargs = self.optimizer_kwargs.copy()
+        kwargs['dimensions'] = dimensions
+        optimizer = Optimizer(**kwargs)
+
+        return optimizer
+
+    def _skopt_to_dict(self, params_space, params):
```
betatim: should we make this guy available in `skopt.utils`? Was discussed in #140 and might be a useful tool in general for people building things themselves?

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+        """Converts list of parameter values into the dictionary with
+        keys as parameter names and corresponding values of parameter.
```
betatim: Would change to:
"""
Convert a list of parameter values into a dictionary.

Parameter names are used as the key for that parameter.
...
"""

skopt/searchcv.py
betatim: (together with the inverse function?)

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+
+        # convert n_jobst to int > 0 if necessary
+        if n_jobs < 0:
+            n_jobs = max(1, cpu_count() + n_jobs + 1)
+
+        # use the cached optimizer for particular parameter space
+        key = str(param_space)
+        if key not in self.optimizer_:
+            self.optimizer_[key] = self._make_optimizer(param_space)
+        optimizer = self.optimizer_[key]
```
betatim: -> `optimizer = self.optimizer_.get(key, self._make_optimizer(param_space))`

skopt/searchcv.py
```diff
@@ -0,0 +1,448 @@
...
+        -------
+        params_dict: dictionary with parameter values.
+        """
+
+        # convert n_jobst to int > 0 if necessary
+        if n_jobs < 0:
+            n_jobs = max(1, cpu_count() + n_jobs + 1)
+
+        # use the cached optimizer for particular parameter space
+        key = str(param_space)
```
betatim: do we need to turn the space into a str before using it as a key? Maybe we should leave a comment where `self.optimizer_` is defined explaining why we can't use `space` directly.

betatim:
Having one optimizer for each entry in the list of "spaces" means we can't use the surrogate to choose between the spaces or have some kind of "cross learning" (a word I just made up).

If each subspace had the same dimensions we could simply add an additional categorical dimension to select which sub-space to look at. However we don't have that.

Can't we make one big space out of all the subspaces + one categorical dimension to select which subspace we want and then have the objective select the parameters of interest based on that categorical. Ideally the surrogate would then learn that for certain values of the categorical space the settings of some parameters are irrelevant.

iaroslav-ai:
@betatim, I think this is an interesting idea, but I avoided it for the following two reasons:
1. It is a bit easier to implement it as it is right now
2. I am not sure if it will yield better results with converting to one large space for the intended use case of BayesSearchCV.

Explanation for 2:
I assume that different search spaces span parameters for different ML models, and hence are largely independent. Thus concatenation of such spaces has a special structure, which is not assumed by the optimization algorithms in skopt. Thus the algorithm needs to learn such structure accurately, in order to understand that many parameters are independent, and that particular parameter selects different subspaces, and changing parameters in only selected subspace has effect. In the end it might not yield advantage, as subspaces are (assumed to be largely) independent anyway. If they are not, it is unclear how many evaluations it takes for the model to figure this out. 

A different way to see it: let a model parameter space A consists of categorical value a := {1,2,3,4} and b := {4,5,6,7}, and space B of c := {6,7,8,9}. A combination of both has 4 parameters, with total 128 parameter combinations to explore. When explored separately, this results only in 20 combinations to explore. 

Also such search strategy might require large n_calls, in which case fitting surrogate can become rather computationally expensive, eg for GP. 

I propose to consider this later, maybe in context of #108 , where I would actually argue for the approach that you propose. If it is implemented, then we can compare on benchmarks different approaches.

skopt/searchcv.py
iaroslav-ai: Yeah actually, space is a dict, which is unhashable type, hence I str it. Will add comment.

skopt/searchcv.py
iaroslav-ai: I agree. Will put it in there somewhere.

skopt/searchcv.py
iaroslav-ai: Ooops, will fix :D

skopt/searchcv.py
iaroslav-ai: ... and its str seems to be unique for different spaces

iaroslav-ai:
Made a pass over the comments. Please take a look and let me know if you have any other comments.

betatim:
Thanks for the long explanation. Sounds sensible. Maybe we can add to the doc string a bit on the separate number of `n_iter` for each subspace and that this gives a "weighting" to each?

Travis is unhappy :(

betatim:
Let's see if we can get @glouppe to have a second pass

betatim:
There are a few GridSearchCV left over which should be renamed.

skopt/searchcv.py
```diff
@@ -0,0 +1,425 @@
+from skopt import Optimizer
+from skopt.utils import params_asdict, dims_aslist
+
+from sklearn.base import clone
+from sklearn.externals.joblib import Parallel, delayed, cpu_count
+import sklearn.model_selection._search as sk_model_sel
+
+import numpy as np
+from collections import defaultdict
```
betatim: Can you sort the imports by stdlib, major package, minor package, our package

skopt/tests/test_searchcv.py
```diff
@@ -0,0 +1,72 @@
...
+    * `n_jobs` [int]:
+        Number of parallel processes to use for computations.
+
+    """
+    from skopt.space import Real, Categorical, Integer
+    from skopt import BayesSearchCV
+
+    from sklearn.datasets import load_iris
+    from sklearn.svm import SVC
+    from sklearn.model_selection import train_test_split
```
betatim: import these at the top of the file instead of in the function

skopt/utils.py
```diff
@@ -292,3 +292,75 @@ def cook_estimator(base_estimator, space=None, **kwargs):
 
     base_estimator.set_params(**kwargs)
     return base_estimator
+
+
+def dims_aslist(params_space):
```
betatim: nitpick: spell out params and dims. I'd prefer that despite it being more to type

skopt/utils.py
```diff
@@ -292,3 +292,75 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+    """Convert dictionary of parameter names : skopt.space.Dimension
+    a list of skopt.space.Dimension, that can be used in Optimizer.
```
betatim: I will think about a different way to formulate this. Right now I had to read it several times despite knowing what the function does :-/

""Convert a dictionary of parameters to a list.

The list of dimensions is created from the values of the
dictionary, sorted by the name of the dimensions used
as keys.

Counterpart to `parameters_asdict`.
"""

Maybe this?

iaroslav-ai: omg lol yeah it was late when I was writing it :D
I will use your version - thanks for coming up with this!

skopt/searchcv.py
iaroslav-ai: This is a standard thing to do right? Will do so elsewhere also

iaroslav-ai:
Addressed comments by @betatim . Please let me know if there is something else.

skopt/searchcv.py
betatim: Each optimizer is associated to one of the spaces defined by the user (as a list). So could we make `self.optimizer_` a list as well? Where the order of optimizers corresponds to the order of spaces? Only thought of this yesterday night while pondering how likely it is that we will get subtle bugs because of using `str()` (In my experience constructing a way to hash an unhashable object is the path to pain and subtle bugs)

iaroslav-ai: I would go with your intuition on this, I do not have much experience with hashing in python. Actually some time ago I also had some doubts about this, but forgot about it. I will come up with something that does not require construction of a way to hash a dict with space parameters.

skopt/__init__.py
```diff
@@ -61,7 +61,7 @@ def f(x):
 from .utils import load
 from .utils import dump
 from .utils import expected_minimum
-
+from .searchcv import BayesSearchCV
```
glouppe: Can you put this in the alphabetical order in the list above?

skopt/searchcv.py
```diff
@@ -0,0 +1,425 @@
...
+
+import numpy as np
+
+from sklearn.base import clone
+from sklearn.externals.joblib import Parallel, delayed, cpu_count
+import sklearn.model_selection._search as sk_model_sel
+
+from skopt import Optimizer
+from skopt.utils import point_asdict, dimensions_aslist
```
glouppe: those should be local imports

skopt/searchcv.py
```diff
@@ -0,0 +1,425 @@
...
+
+    If `n_jobs` was set to a value higher than one, the data is copied for each
+    parameter setting(and not `n_jobs` times). This is done for efficiency
+    reasons if individual jobs take very little time, but may raise errors if
+    the dataset is large and not enough memory is available.  A workaround in
+    this case is to set `pre_dispatch`. Then, the memory is copied only
+    `pre_dispatch` many times. A reasonable value for `pre_dispatch` is `2 *
+    n_jobs`.
+
+    See Also
```
glouppe: Could we add a short `Example` section in the docstring? (showing that things are in fact simple, despite the huge list of arguments)

skopt/searchcv.py
glouppe: Hmm, not sure what is done inside `self._make_optimizer`, but we should ensure there is no border effects... Your suggestion is not equivalent to @iaroslav-ai's code, in case `self._make_optimizer` changes the state of `self` internally.

skopt/searchcv.py
```diff
@@ -0,0 +1,425 @@
...
+
+    Parameters
+    ----------
+    estimator : estimator object.
+        A object of that type is instantiated for each search point.
+        This object is assumed to implement the scikit-learn estimator api.
+        Either estimator needs to provide a ``score`` function,
+        or ``scoring`` must be passed.
+
+    search_spaces : list of dict or tuple
```
glouppe: While this is nice, the 90% use case is to search over a single space. I think we should support directly providing a dict. This would also better match with the API of `GridSearchCV`.

skopt/tests/test_searchcv.py
```diff
@@ -0,0 +1,67 @@
...
+        }],
+        n_jobs=n_jobs, n_iter=11,
+        optimizer_kwargs=optimizer_kwargs
+    )
+
+    opt.fit(X_train, y_train)
+
+    # this normally does not hold only if something is wrong
+    # with the optimizaiton procedure as such
+    assert_greater(opt.score(X_test, y_test), 0.9)
```
glouppe: We should have a test evaluating the multiple search spaces API.

skopt/searchcv.py
```diff
@@ -0,0 +1,547 @@
...
+
+            # 2. check all the dicts for correctness of contents
+            for subspace in dicts_only:
+                for k, v in subspace.items():
+                    if not isinstance(v, Dimension):
+                        raise TypeError(
+                            "Parameter of space should be an instance of"
+                            "skopt.space.Dimension (Real, Integer, ...),"
+                            "but in subspace %s the dimension %s is %s" %
+                            (subspace, k, v)
```
glouppe: Could we instead use `skopt.space.check_dimension`? In this way, we would have a consistent API with `Optimizer/*_minimize` in terms of how dimensions are specified. (Also, this would allow much less boilerplate code for the users.)

iaroslav-ai:
This should address all the comments. Will make a second pass over the code later to double check.

glouppe:
In the notebook, could you update the minimal example in order to use the simplest API? (directly feeding the dict, with dimensions specified as pairs)

iaroslav-ai:
Yup will do so

iaroslav-ai:
Updated! :)

glouppe:
Thanks! This looks very nice :)

One more thing though... in the last part of the notebook, it seems almost nothing is learned. Within the very first iterations, the optimizer reaches a good value and does not improve from there. Have you tried with more iterations to see if things eventually improve?

iaroslav-ai:
I did not try, but my suspicion would be that it will not improve much, as the dataset is not really that complex and is used mainly as a simple example. Maybe I could see what other datasets could be used.

skopt/tests/test_searchcv.py
```diff
@@ -42,6 +42,15 @@ def test_searchcv_runs(surrogate, n_jobs):
     # None search space is only supported when only `step` function is used
     assert_raises(ValueError, BayesSearchCV(SVC(), None).fit, (X, y))
 
+    # check if invalid dimensions are raising errors
+    def constructor(x): BayesSearchCV(*x)
+    assert_raises(
+        ValueError, constructor, (SVC(), {'C':'1 ... 100.0'})
+    )
```
betatim:
```python
with pytest.raises(ValueError):
  BayesSearchCV(args_here)
```
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions for soem more examples

skopt/searchcv.py
```diff
@@ -0,0 +1,547 @@
...
+                        )
+        else:
+            raise TypeError(
+                "Search space should be provided as a dict or list of dict,"
+                "got %s" % search_space)
+
+    def add_spaces(self, spaces, names):
```
betatim: Not sure it needs to be public, but if it is it needs docs. Which might motivate making it private :)

skopt/utils.py
```diff
@@ -292,3 +292,99 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+    -------
+    point_as_list: list with point values.The order of
+        parameters in the list is given by sorted(params_space.keys()).
+        Example output with example inputs:
+            [0.66, 3, -0.15]
+    """
+    point_as_list = [
+        point_as_dict[k] for k in sorted(search_space.keys())
+    ]
+    return point_as_list
```
betatim: Needs a unit test

betatim: Maybe in `test_utils.py` we can add a test that just tries to round trip a dict to a list and back

skopt/tests/test_searchcv.py
iaroslav-ai: ahaaa so that is how you do it thx :)

MechCoder:
Feel free to merge this, I'll try to have a look next week and then make some comments if any. Not sure I will have much to say in any case.

skopt/tests/test_searchcv.py
betatim: Trying to spread the pytest way of life :)

glouppe:
@betatim Was this ok with you? I think this PR is good enough to be merged in :) We can polish things further later.

glouppe:
:beers: Great work @iaroslav-ai ! Thanks a lot for this, I am pretty sure this will be super useful to many :)

betatim:
🍻 and 🍰 !

This is a big new feature! Thanks for the work and patience with the many little comments spread out over days :)

I think we can do a bit more work on improving the doc strings etc but let's do that in a new PR (or several).

iaroslav-ai:
A huge thank you to @glouppe  and @betatim for taking your time to review my code! Indeed quite some PR it is :)
I am also testing this internally for some of my things, seems quite useful so far! :)

amueller:
Awesome! Did someone do any benchmarks with how this compares to GridSearchCV in particular cases?

betatim:
Don't think so. The best we have is https://github.com/scikit-optimize/scikit-optimize/tree/master/benchmarks#ml-classification

Would be great if someone could do perform some benchmarks on "realistic" data sets.

I once saw some benchmarks in https://github.com/rhiever/tpot maybe we can start from there instead of having to do everything ourselves. Not related to tuning ML algos: https://github.com/sigopt/evalset but maybe faster to execute.

PR #409, opened 2017-06-20
[MRG + 1]: Provide intuitive deduction for Dimension type

### LINKED ISSUE #391: Unintuitive deduction of dimension type
```python
optimizer = Optimizer([(0, 1.124124)], base_estimator, n_random_starts=5)
```

to me it is wrong that this would create an `Integer` dimension. We should check if both or any of the limits is a float and then create a `Continous` dimension.

mirca:
Fixes #391 

I also did a few stylistic changes. Let me know if they are not desired.

Cheers :)


skopt/space/space.py
```diff
@@ -66,21 +66,29 @@ def check_dimension(dimension, transform=None):
...
+            return Real(*dimension, transform=transform)
+        elif all([isinstance(dim, int) for dim in dimension]):
+            return Integer(*dimension, transform=transform)
+        else:
+            raise ValueError("Invalid dimension {}. Read the documentation for"
+                             " supported types.".format(dimension))
 
-    if len(dimension) > 2 or isinstance(dimension[0], str):
-        return Categorical(dimension, transform=transform)
+    if len(dimension) == 3:
```
MechCoder: you can use elif here and below

skopt/space/space.py
```diff
@@ -66,21 +66,29 @@ def check_dimension(dimension, transform=None):
...
+        if (any([isinstance(dim, (float, int))
+                 for dim in dimension[:2]]) and isinstance(dimension[2], str)):
```
MechCoder:
```python
dimension[2] in ["uniform", "log-uniform"]
```

skopt/tests/test_space.py
```diff
@@ -160,6 +160,7 @@ def test_space_consistency():
     s3 = Space([Real(0, 1)])
     s4 = Space([(0.0, 1.0)])
     s5 = Space([(0.0, 1.0, "uniform")])
+    s6 = Space([(0, 1.0)])
```
MechCoder: Can you add a test for

```python
Space((1.0, "r"))
Space((1, "r"))
```

as well?

mirca: Well, according to the definition of ``Categorical``, those should be different because ``type(1) != type(1.0)``.

MechCoder: haha no, I meant to test the behaviour that both of them are treated as Categorical, (the current behaviour on master breaks)

mirca: Oh, ok ;)

betatim:
For my education, what is the current recommendation from the python gurus on `class Foo(object)` vs `class Foo:` in a python code base that also wants to support legacy python?

mirca:
@betatim In Python 3, it doesn't make any difference inheriting from ``object`` or not (inheritance takes place implicitly). In Python 2, not inheriting from ``object`` generates a few troubles, such as

```
>>> class Hi: pass
>>> class Oi: pass
>>> h = Hi()
>>> o = Oi()
>>> type(h) is type(o)
True
```

betatim:
Merci!

Was hoping you'd tell me that python2 finally added something so we don't need to do that hack anymore... 😆 Secretly I ❤️ legacy python.

mirca:
@betatim hahaha, sorry :disappointed:

PR #410, opened 2017-06-21
[MRG] Fix install instructions to explicitly install numpy

### LINKED ISSUE #404: Add 'empty' environment build slot?
In our CI setup we install (some of) the dependencies "by hand" in `build_tools/*/install.sh`. This means we don't notice problems like scikit-garden/scikit-garden#23.

Should we add a new entry to the build matrix that tries to install things "as a user would" into a otherwise empty env? This would also test whether or not we are keeping out `setup.py` up to date.

A different way of asking this is: should we remove the explicit installing of dependencies from `install.sh`?

betatim:
Closes #404 and adds a (temporary?) workaround to the install instructions.

We need to install numpy before trying to install scikit-optimize
as numpy is needed to run the `setup.py` of scikit-garden which
we depend on. While scikit-garden declares numpy as a dependency the seutp.py can't be executed without it being installed, which creates some kind of infinite loop.

The discussion of having numpy in `setup.py` is in scikit-garden/scikit-garden#23

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,7 @@ Install
 
 ::
 
+    pip install numpy # install numpy explicitly first
     pip install scikit-optimize
 
 Getting started
--- a/build_tools/circle/execute.sh
+++ b/build_tools/circle/execute.sh
@@ -2,6 +2,10 @@ export PATH="$HOME/miniconda3/bin:$PATH"
 source activate testenv
 export SKOPT_HOME=$(pwd)
 
+python --version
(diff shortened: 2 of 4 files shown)


betatim:
Let's see if travis picks up the new config files to test the build procedure.

MechCoder:
LGTM this fixes https://github.com/scikit-optimize/scikit-optimize/issues/404 as well?

betatim:
> LGTM this fixes https://github.com/scikit-optimize/scikit-optimize/issues/404 as well?

Correct.

PR #411, opened 2017-06-22
[MRG] Reset PyPI password

betatim:
I reset my PyPI password so needed to update the credentials stored in our travis config


betatim:
Merging this so we don't forget. We should keep this at the back of our minds when we cut the next release and deploying to PyPI fails :-/

PR #412, opened 2017-06-22
check if transformer is instance of CategoricalEncoder in Pipeline.fit

mirca:
I propose we check that every ``transformer`` in the ``Pipeline`` class is an instance of ``CategoricalEncoder`` and raise a helpful error message if any of them are not.

### DIFF
--- a/skopt/space/transformers.py
+++ b/skopt/space/transformers.py
@@ -152,7 +152,11 @@ class Pipeline(Transformer):
 
     def fit(self, X):
         for transformer in self.transformers:
-            transformer.fit(X)
+            if isinstance(transformer, CategoricalEncoder):
+                transformer.fit(X)
+            else:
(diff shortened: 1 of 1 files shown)


glouppe:
I am rather -0 on this. I'd rather rely on duck typing and not impose types anywhere. 

Also, not clear for me what you want to prevent with this?

mirca:
I agree with you, duck typing is the best. I just wanted to see a clear error message rather than "AttributeError: transformer has no attribute 'fit'"

betatim:
We could catch the `AttributeError` and print a message saying "A transformer has to implement `fit`"?

-0 on checking types.

mirca:
@betatim yeah, bad idea checking types, sorry :|

betatim:
Alternative idea: we couldcheck that each transformer has a `fit` method in `__init__` of the pipeline. To give a better error message earlier to the user. Just an idea.

PR #416, opened 2017-06-28
Remove n_random_starts references in tests

MechCoder:


betatim:
Thanks for spotting these!

Issue #417, opened 2017-06-29
`test_expected_minimum` failure


betatim:
One test for the expected minimum function fails somewhere deep in scipy. Any ideas on how to track this down?

```
$ pytest --pdb -x -m 'not slow_test' skopt/tests/test_utils.py          (skopt)
============================= test session starts ==============================
platform darwin -- Python 3.5.2, pytest-3.0.7, py-1.4.31, pluggy-0.4.0
rootdir: /Users/thead/git/scikit-optimize, inifile:
collected 3 items 

skopt/tests/test_utils.py ..F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

    @pytest.mark.fast_test
    def test_expected_minimum():
        res = gp_minimize(bench3,
                          [(-2.0, 2.0)],
                          x0=[0.],
                          noise=0.0,
                          n_calls=20,
>                         random_state=1)

skopt/tests/test_utils.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
skopt/optimizer/gp.py:238: in gp_minimize
    callback=callback, n_jobs=n_jobs)
skopt/optimizer/base.py:249: in base_minimize
    result = optimizer.tell(next_x, next_y, fit=fit_model)
skopt/optimizer/optimizer.py:407: in tell
    est.fit(self.space.transform(self.Xi), self.yi)
skopt/learning/gaussian_process/gpr.py:194: in fit
    super(GaussianProcessRegressor, self).fit(X, y)
../../anaconda/envs/skopt/lib/python3.5/site-packages/sklearn/gaussian_process/gpr.py:217: in fit
    bounds))
../../anaconda/envs/skopt/lib/python3.5/site-packages/sklearn/gaussian_process/gpr.py:424: in _constrained_optimization
    fmin_l_bfgs_b(obj_func, initial_theta, bounds=bounds)
../../anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/lbfgsb.py:193: in fmin_l_bfgs_b
    **opts)
../../anaconda/envs/skopt/lib/python3.5/site-packages/scipy/optimize/lbfgsb.py:330: in _minimize_lbfgsb
    f, g = func_and_grad(x)
[...]

Issue #348, opened 2017-04-10
GradientBoostingQuantileRegressor: non-broadcastable output operand with shape (3,) doesn't match the broadcast shape (3,3)


iaroslav-ai:
When I run the following code:
```
from skopt import Optimizer
from skopt.learning import GradientBoostingQuantileRegressor
from skopt.space import Real

opt = Optimizer([Real(-1,1)],GradientBoostingQuantileRegressor(),acq_optimizer='sampling')

def obj(x):
    return sum(x)

for i in range(128):
    x = opt.ask()
    y = obj(x)
    opt.tell(x, y)
```
it throws an exception:
```
/usr/bin/python2.7 /home/iaroslav/scikit-optimize/benchmarks/min_gbrt.py
Traceback (most recent call last):
  File "/home/iaroslav/scikit-optimize/benchmarks/min_gbrt.py", line 13, in <module>
    opt.tell(x, y)
  File "/home/iaroslav/scikit-optimize/skopt/optimizer/optimizer.py", line 271, in tell
    self.gains_ -= est.predict(np.vstack(self.next_xs_))
ValueError: non-broadcastable output operand with shape (3,) doesn't match the broadcast shape (3,3)

Process finished with exit code 1
```
If I try to run this with other surrogates, I do not experience this exception. Looks like a bug. 
The code was run on python 2.7.

iaroslav-ai:
Setting `acq_func='EI'` in Optimizer solves the issue. It seems that default `acq_func="gp_hedge"` does not work well with `GradientBoostingQuantileRegressor`.

iaroslav-ai:
Appears to be solved now. Closing.

PR #418, opened 2017-06-29
[MRG] Add support for random "base estimator" in Optimizer

### LINKED ISSUE #406: `dummy_minimize` with ask and tell?
Just realized this, does the current API allow one to use random search with the ask and tell API? (in a user-friendly way)

betatim:
Fixes #406 

This removes the need to special case dummy_minimize by adding
the ability to switch to supplying only random suggestions in
the Optimizer class.

Still uses `None`, needs changing to "dummy".

### DIFF
--- a/skopt/optimizer/dummy.py
+++ b/skopt/optimizer/dummy.py
@@ -84,7 +84,7 @@ def dummy_minimize(func, dimensions, n_calls=100, x0=None, y0=None,
     else:
         n_random_calls = n_calls
 
-    return base_minimize(func, dimensions, base_estimator=None,
+    return base_minimize(func, dimensions, base_estimator="dummy",
                          n_calls=n_calls, n_random_starts=n_random_calls,
                          x0=x0, y0=y0, random_state=random_state,
                          verbose=verbose,
--- a/skopt/optimizer/dummy.py
+++ b/skopt/optimizer/dummy.py
@@ -79,8 +79,8 @@ def dummy_minimize(func, dimensions, n_calls=100, x0=None, y0=None,
     """
     # all our calls want random suggestions, except if we need to evaluate
     # some initial points
-    if x0 is not None:
-        n_random_calls = n_calls - (len(x0) if not y0 else 0)
+    if x0 is not None and y0 is None:
+        n_random_calls = n_calls - len(x0)
     else:
         n_random_calls = n_calls
 
--- a/skopt/optimizer/dummy.py
+++ b/skopt/optimizer/dummy.py
@@ -1,18 +1,6 @@
 """Random search."""
 
-import copy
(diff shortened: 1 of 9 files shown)


glouppe:
+40 / -100 lines of code :+1: 

Looks good to me otherwise.

betatim:
> Looks good to me otherwise.

What do you mean otherwise? Deleting code is the best :)

skopt/optimizer/dummy.py
```diff
@@ -89,73 +77,15 @@ def dummy_minimize(func, dimensions, n_calls=100, x0=None, y0=None,
...
-        y.append(y_i)
-        result = create_result(X[:i + 1], y, space, rng, specs)
-        if eval_callbacks(callbacks, result):
-            break
-
-    y = np.array(y)
-    return create_result(X, y, space, rng, specs)
+    # all our calls want random suggestions, except if we need to evaluate
+    # some initial points
+    if x0 is not None:
```
MechCoder: nitpick: maybe you can write this as

```python
if x0 is not None and y0 is None:
    n_random_calls = n_calls - len(x0)
```

I looked at just the if statement and was wondering why it would be true, before I read the next line.

MechCoder:
Can you please add a test that tests the behaviour of using the `Optimizer` instance with `dummy` as `base_estimator` mainly the attributes of the `results` that the `tell` method returns and the output of `ask`?

betatim:
As part of this I ended up changing the interface of the gradient boosted quantile regressor. It was odd compared to the others because it returned the prediction for each quantile. This is in 	79b0e0a. What should we do in terms of API deprecation etc? I would like to just do it and assume that no one had been digging that deep into our internals ... but ...

We should probably also add some tests that nail down the API we expect of our regressors.

iaroslav-ai:
Changing the interface of GBQR might be relevant for #348 , but will take a more detailed look later

glouppe:
> What should we do in terms of API deprecation etc? I would like to just do it and assume that no one had been digging that deep into our internals ... but ...

Looks reasonable, not sure we have many power users just yet :)

betatim:
This also fixes #417

betatim:
> Changing the interface of GBQR might be relevant for #348 , but will take a more detailed look later 

Jupp this is also fixed in this PR

MechCoder:
Merging, thanks!

MechCoder:
hmm. I think it might be better to set `self.base_estimator_` to `"Dummy"` instead of None and return "Dummy" instead of None from `cook_estimator`. wdyt?

betatim:
Not sure. I thought that it would be OK to use `None` as value for "we don't have a base_estimator" instead of adding our own word/value for it.

Issue #402, opened 2017-06-15
Obtaining same functionality of gp_minimize with Optimizer


kittythecat:
gp_minimize works well for my problem set, but I'd really like to use the ask() and tell() features of Optimizer. However, I've been unable to reproduce gp_minimize using Optimizer. Is there an example that we can post somewhere?

Thank you

betatim:
Currently there is no example of how to reproduce the setup of any of the `*_minimize` functions.

There are two things to look at: how the space is transformed and how the kernel is setup. At the moment it is probably easiest to look at the [code](https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/gp.py#L206-L243).

Some related issues #338 #390 #400

Young-Todd:
@kittythecat, have you had any luck with this?

MechCoder:
#408 is a step in that direction.

rene-rex:
I am in the same situation as @kittythecat, so +1 from me!

betatim:
You should [now](https://github.com/scikit-optimize/scikit-optimize/pull/408) be able to now do something like:
```python
opt = Optimizer([(-2.0, 2.0)], "GP")
```
by providing a string as the optimizer you will get a surrogate model that is configured like the ones used in `*_minimize`. You can also look at `skopt.utils.cook_estimator()` to learn how they are configured.

betatim:
Feel free to keep discussing in this thread if this doesn't work (or open a new one if there is something not directly related to this).

Closing the issue to help us keep track of what needs doing for the v0.4 release.

MechCoder:
Hmm can we still keep this open?

It would be great to add a test to check that both indeed produce the same results.

rene-rex:
As far as I understand the code, `gp_minimize` is also doing some space transformations which are not handled by the new `cook_estimator` method. Should this be done by the user before constructing the optimizer? Maybe an option is to pull out the code from `gp_minimize` and offer a `skopt.utils.transform_space`?

MechCoder:
That's right. Would you like to propose a pull request?

yngtodd:
[...]

PR #420, opened 2017-07-03
[MRG+1] Refactoring dimension transformations in `gp_minimize` to be used by `Optimizer`.

yngtodd:
This is a small step towards obtaining the same results with `gp_minimize` and the `Optimizer` class noted in https://github.com/scikit-optimize/scikit-optimize/issues/402. 

The code extracts the code responsible for transforming the `dimensions` in `gp_minimize`, creating a new function, ~~`transform_gpdims`~~  ~~`transform_dims`~~ `normalize_dimensions`, which can then be used by both `gp_minimize` and `Optimizer`.

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -131,7 +131,7 @@ class Optimizer(object):
         to sample points, bounds, and type of parameters.
 
     """
-    def __init__(self, dimensions, base_estimator="gp",
+    def __init__(self, dimensions, base_estimator="GP",
                  n_random_starts=None, n_initial_points=10,
                  acq_func="gp_hedge",
                  acq_optimizer="auto",
@@ -174,7 +174,7 @@ class Optimizer(object):
         self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
         self.n_jobs = n_jobs
 
-        if base_estimator == "gp":
+        if base_estimator == "GP":
             dimensions = normalize_dimensions(dimensions)
         self.space = Space(dimensions)
         self.models = []
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -22,6 +22,7 @@ from ..utils import create_result
 from ..utils import has_gradients
 from ..utils import is_listlike
 from ..utils import is_2Dlistlike
+from ..utils import normalize_dimensions
 
 
 class Optimizer(object):
@@ -131,7 +132,7 @@ class Optimizer(object):
         to sample points, bounds, and type of parameters.
 
     """
-    def __init__(self, dimensions, base_estimator="gp",
+    def __init__(self, dimensions, base_estimator="GP",
                  n_random_starts=None, n_initial_points=10,
                  acq_func="gp_hedge",
                  acq_optimizer="auto",
@@ -165,6 +166,8 @@ class Optimizer(object):
         n_jobs = acq_optimizer_kwargs.get("n_jobs", 1)
         self.acq_optimizer_kwargs = acq_optimizer_kwargs
 
+        if base_estimator == "GP":
+            dimensions = normalize_dimensions(dimensions)
         self.space = Space(dimensions)
         self.models = []
         self.Xi = []
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -288,7 +288,7 @@ class Optimizer(object):
                https://hal.archives-ouvertes.fr/hal-00732512/document
 
                With this strategy a copy of optimizer is created, which is
-               then asked for a point, and the point is told to the copy of
+               then ed for a point, and the point is told to the copy of
                optimizer with some fake objective (lie), the next point is
                asked from copy, it is also told to the copy with fake
                objective and so on. The type of lie defines different
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -132,7 +132,7 @@ class Optimizer(object):
         to sample points, bounds, and type of parameters.
 
     """
-    def __init__(self, dimensions, base_estimator="GP",
+    def __init__(self, dimensions, base_estimator="gp",
                  n_random_starts=None, n_initial_points=10,
                  acq_func="gp_hedge",
                  acq_optimizer="auto",
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -24,7 +24,6 @@ from ..utils import is_listlike
 from ..utils import is_2Dlistlike
 from ..utils import normalize_dimensions
 
-
 class Optimizer(object):
     """Run bayesian optimisation loop.
 
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -24,6 +24,7 @@ from ..utils import is_listlike
 from ..utils import is_2Dlistlike
 from ..utils import normalize_dimensions
 
+
 class Optimizer(object):
     """Run bayesian optimisation loop.
 
@@ -173,7 +174,7 @@ class Optimizer(object):
 
         if base_estimator == "GP":
             dimensions = normalize_dimensions(dimensions)
-            
+
         self.space = Space(dimensions)
         self.models = []
         self.Xi = []
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -131,7 +131,7 @@ class Optimizer(object):
         to sample points, bounds, and type of parameters.
 
     """
-    def __init__(self, dimensions, base_estimator="GP",
+    def __init__(self, dimensions, base_estimator="gp",
                  n_random_starts=None, n_initial_points=10,
                  acq_func="gp_hedge",
                  acq_optimizer="auto",
@@ -197,7 +197,6 @@ class Optimizer(object):
 
         self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
         self.n_jobs = n_jobs
-
 
         # The cache of responses of `ask` method for n_points not None.
         # This ensures that multiple calls to `ask` with n_points set
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -165,6 +165,15 @@ class Optimizer(object):
         n_jobs = acq_optimizer_kwargs.get("n_jobs", 1)
         self.acq_optimizer_kwargs = acq_optimizer_kwargs
 
+        if n_random_starts is not None:
+            warnings.warn(("n_random_starts will be removed in favour of "
+                           "n_initial_points."),
+                          DeprecationWarning)
+            n_initial_points = n_random_starts
+
+        self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
+        self.n_jobs = n_jobs
+
         if base_estimator == "GP":
             dimensions = normalize_dimensions(dimensions)
         self.space = Space(dimensions)
@@ -179,15 +188,6 @@ class Optimizer(object):
                 self._cat_inds.append(ind)
             else:
                 self._non_cat_inds.append(ind)
-
-        if n_random_starts is not None:
-            warnings.warn(("n_random_starts will be removed in favour of "
-                           "n_initial_points."),
-                          DeprecationWarning)
-            n_initial_points = n_random_starts
-
-        self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
-        self.n_jobs = n_jobs
 
         # The cache of responses of `ask` method for n_points not None.
         # This ensures that multiple calls to `ask` with n_points set
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -174,7 +174,7 @@ class Optimizer(object):
         self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
         self.n_jobs = n_jobs
 
-        if base_estimator == "GP":
+        if base_estimator == "gp":
             dimensions = normalize_dimensions(dimensions)
         self.space = Space(dimensions)
         self.models = []
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -189,6 +189,16 @@ class Optimizer(object):
             else:
                 self._non_cat_inds.append(ind)
 
+        if n_random_starts is not None:
+            warnings.warn(("n_random_starts will be removed in favour of "
+                           "n_initial_points."),
+                          DeprecationWarning)
+            n_initial_points = n_random_starts
+
+        self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
+        self.n_jobs = n_jobs
+
+
         # The cache of responses of `ask` method for n_points not None.
         # This ensures that multiple calls to `ask` with n_points set
         # return same sets of points.
@@ -287,7 +297,7 @@ class Optimizer(object):
                https://hal.archives-ouvertes.fr/hal-00732512/document
 
                With this strategy a copy of optimizer is created, which is
-               then ed for a point, and the point is told to the copy of
+               then asked for a point, and the point is told to the copy of
                optimizer with some fake objective (lie), the next point is
                asked from copy, it is also told to the copy with fake
                objective and so on. The type of lie defines different
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -456,7 +456,7 @@ def normalize_dimensions(dimensions):
     if isinstance(base_estimator, str):
         base_estimator = base_estimator.upper()
         if base_estimator not in ["GP", "ET", "RF", "GBRT"]:
-            raise ValueError("Valid strings for the base_estimator parameter "
+            raise ValueError("Valid strings for the base_estimator parameter " +
                              "are: 'RF', 'ET', 'GBRT', or 'GP'.")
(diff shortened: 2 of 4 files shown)


skopt/utils.py
```diff
@@ -292,3 +295,44 @@ def cook_estimator(base_estimator, space=None, **kwargs):
 
     base_estimator.set_params(**kwargs)
     return base_estimator
+
+def transform_gpdims(dimensions):
```
MechCoder: What do you think of making this as `transform_dims` with an `estimator` argument?, then we can relegate the special casing of `estimator` to `transform_dims`

yngtodd: I like that idea.

MechCoder: no you are right. I just thought that abstracting away the logic would be a good idea, in case we would like to make some changes in the future.

yngtodd: I am with you. Just wanted to make sure that I wasn't missing any other places that could use the dimension transformations. 

I have updated the code. Now `transform_dims` has an argument `estimator="GP"`. I figured a default setting to the Gaussian process would work best.

skopt/optimizer/optimizer.py
```diff
@@ -176,6 +178,7 @@ def __init__(self, dimensions, base_estimator="gp",
         self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
         self.n_jobs = n_jobs
 
+
```
betatim: spurious blank line

skopt/utils.py
```diff
@@ -292,3 +295,50 @@ def cook_estimator(base_estimator, space=None, **kwargs):
 
     base_estimator.set_params(**kwargs)
     return base_estimator
+
+def transform_dims(dimensions, estimator="GP"):
```
betatim: can we name it `transform_dimensions` or maybe even `normalize_dimensions`? Right now the name doesn't really tell you what the function does.

skopt/utils.py
```diff
@@ -292,3 +295,50 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+    """
+    Transforms space for Gaussian processes.
```
betatim: We should definitely explain what the function does here instead of referring back to some historical setup in the code. I'd suggest something like:
```
Normalize all dimensions.

This is particularly useful for gaussian process based regressors and is used internally by `gp_minimize.`
```

skopt/utils.py
```diff
@@ -292,3 +295,50 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
+          dimensions),
+        - a `(upper_bound, lower_bound, "prior")` tuple (for `Real`
+          dimensions),
+        - as a list of categories (for `Categorical` dimensions), or
+        - an instance of a `Dimension` object (`Real`, `Integer` or
+          `Categorical`).
+
+         NOTE: The upper and lower bounds are inclusive for `Integer`
+         dimensions.
```
betatim: needs a blank line

skopt/utils.py
```diff
@@ -292,3 +295,50 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+    * `estimator` [str, ]:
+        optimization estimator to be used.
+        each estimator may use different transformations for dimensions.
+    """
```
betatim: Maybe we can add an example here to show how to use it together with `Optimizer`. Would that be useful for people who don't know the insides of `gp_minimize`?

skopt/utils.py
```diff
@@ -292,3 +295,50 @@ def cook_estimator(base_estimator, space=None, **kwargs):
...
+    if estimator == "GP":
```
betatim: I think we should allow the same types for `base_estimator` as `Optimizer` does. This means we need to check strings as well as the type of instances of regressors.

betatim: The alternative might be that we assume the user knows what they are doing when they call this and we just always noramlise the dimensions according to the below recipe. Leaving it to the user to decide if they want to call this or not. Think this is my favourite option at the moment.

skopt/utils.py
yngtodd: `normalize_dimensions` is more clear, I think. Would it be possible that in the future the function could handle some special case for any of the estimators not involving normalizing the dimensions? If so, then `transform_dimensions` may be better.

Either way, I agree it could be more clear.

skopt/utils.py
```diff
@@ -388,3 +387,61 @@ def point_aslist(search_space, point_as_dict):
...
+
+    * `estimator` ["GP", "RF", "ET", "GBRT", or sklearn regressor, default="GP"]:
+        optimization estimator to be used.
+        each estimator may use different transformations for dimensions.
+    """
+    if isinstance(estimator, str):
+        estimator = estimator.upper()
+        if estimator not in ["GP", "ET", "RF", "GBRT"]:
+            raise ValueError("Valid strings for estimator parameter "
+                             "are: 'RF', 'ET', 'GBRT', or 'GP', not %s" % estimator)
```
yngtodd: I am stuck on this one. Travis is upset, and I am getting to following error:

pytest_error

Is my `ValueError` conflicting with the `@pytest.mark.fast_test`?

betatim:
Instead of merging master into your branch we tend to rebase the branch instead. This keeps the history more linear for later when you are digging around. This looks like a nice guide: https://anavarre.net/how-to-rebase-a-github-pull-request/ (ignore the last bit about squashing your changes). If you are new to this make a copy of your repo before starting just in case :) and let me know if you get stuck.

skopt/utils.py
betatim: This test checks that calling `Optimizer` with an invalid value for `base_estimator` raises an exception. It also checks that the message of the exception matches what it expects. This is what is failing at the moment. The expected string does not contain `GBRT` but the actual message does.

I think this is a mistake in the test as GBRT should be in the string. So go ahead and update the test.

skopt/optimizer/optimizer.py
```diff
@@ -124,7 +125,7 @@ class Optimizer(object):
         to sample points, bounds, and type of parameters.
 
     """
-    def __init__(self, dimensions, base_estimator="gp",
+    def __init__(self, dimensions, base_estimator="GP",
```
betatim: you don't need to change this

betatim:
I think the way to go is to make `normalize_dimensions` a function in utils that people can use to assemble their own optimisation pipeline. This means I'd remove the checks for what type of estimator is passed (the user needs to know if they want to call `normalize_dimensions` or not) in and also remove the call to `normalize_dimensions` made inside `Optimizer`.

It does mean you need to know that you need to normalise the dimensions in order to reproduce `gp_minimize` when building your own optimisation pipeline. Don't think this is a big burden because building your own pipeline is something for power users. If we place the call inside `Optimizer` (or check types inside `normalize_dimensions`) then power users need to hack their way around that, which iMHO is even uglier.

What do you guys think?

skopt/optimizer/optimizer.py
```diff
@@ -154,6 +155,7 @@ def __init__(self, dimensions, base_estimator="gp",
         n_jobs = acq_optimizer_kwargs.get("n_jobs", 1)
         self.acq_optimizer_kwargs = acq_optimizer_kwargs
 
+        dimensions = normalize_dimensions(dimensions, estimator=base_estimator)
```
betatim: If we go ahead with my idea from https://github.com/scikit-optimize/scikit-optimize/pull/420#issuecomment-313610646 this is the line to delete.

skopt/utils.py
```diff
@@ -388,3 +387,61 @@ def point_aslist(search_space, point_as_dict):
     ]
     return point_as_list
 
+def normalize_dimensions(dimensions, estimator="GP"):
```
betatim: If we go ahead with my idea from https://github.com/scikit-optimize/scikit-optimize/pull/420#issuecomment-313610646 we should remove the `estimator` argument.

skopt/utils.py
```diff
@@ -15,12 +15,14 @@
...
 
+from .space import check_dimension
+from .space import Categorical
+
 __all__ = (
     "load",
     "dump",
 )
 
-
```
betatim: Please don't delete these blank lines. We follow https://www.python.org/dev/peps/pep-0008/ for code formatting (`pip install pep8` and then `pep8 <filename>` to get advice about formatting). There should be two blank lines between functions, but only one blank line between methods of a class.

skopt/utils.py
yngtodd: Do you mean at the bottom after the note?

yngtodd:
> If we place the call inside Optimizer (or check types inside normalize_dimensions) then power users need to hack their way around that, which iMHO is even uglier.

That is true, and removing it would reduce the lines of code. 

> It does mean you need to know that you need to normalise the dimensions in order to reproduce gp_minimize when building your own optimisation pipeline.

This could be an easy thing to overlook when setting up a pipeline. I can imagine running a bunch of experiments, switching to a different `base_estimator` and forgetting that the Gaussian process needs its dimensions normalized.

betatim:
> This could be an easy thing to overlook when setting up a pipeline. I can imagine running a bunch of experiments, switching to a different base_estimator and forgetting that the Gaussian process needs its dimensions normalized.

I think I'd prefer this over having to try and undo things that `Optimizer` does automatically.

Maybe an alternative is for GP regressors (and others that want normalised dimensions) to output a warning if they detect that the space they are operating on isn't normalised? This would give people a hint that something isn't right, experts could chose to ignore it, and it reminds someone who forgot to add it.

MechCoder:
I strongly believe that if we provide a "magic" default base GP estimator instead of the one that scikit-learn defaults to, then it makes sense to provide "magic" transformations to the dimensions by default that performs better instead of the user having the hassle to set things up.

yngtodd:
> Maybe an alternative is for GP regressors (and others that want normalised dimensions) to output a warning if they detect that the space they are operating on isn't normalised? 

That sounds like it would be good for the forgetful ones. I guess that would be most helpful for non-experts, and that wouldn't be the intended audience with the Optimizer here. 

> I strongly believe that if we provide a "magic" default base GP estimator instead of the one that scikit-learn defaults to, then it makes sense to provide "magic" transformations to the dimensions by default that performs better instead of the user having the hassle to set things up.

This is a good point. One alternative would be to swap the default estimator away from the GP. 

As it stands, when the `Optimizer` is initialized, it is only a matter of setting the `default_estimator` argument to determine whether or not the dimensions will be normalized. If the `default_estimator="GP"`, then the dimensions will be normalized and not otherwise. That does not seem like too much fiddling with setting the `Optimizer` for the programer, but it does introduce some overhead for the program since `normalize_dimensions` is called regardless of which `defualt_estimator` is used.

betatim:
Maybe estimator's can provide a method `prepare_space` that does what that estimator needs doing to the space? IMHO `Optimizer` is for people who know what they are doing, so they need to explicitly do things (IMHO also call `cook_estimator` themselves and pass that into `Optimizer`).

The problem with calling things from inside `Optimizer` is that we need a way to disable or override it. Which probably means increasing the number of arguments or adding some heuristic to detect that the user meant what they said (and `Optimizer` should not modify it).

Maybe more people want access to the ask-and-tell interface and aren't actually interested in being experts. In that case we should maybe provide a `PreConfiguredOptimizer` (better name needed) that does all the things `*_minimize` does and have `Optimizer` as the low level tool that does nothing for you. (We should probably move this discussion to a new issue.)

betatim:
Is it strictly required to normalise the dimensions for it to work with GP based regressors? Or if you don't do it will it work but give worse results?

Maybe as a compromise we can move the check of whether or not to call `normalize_dimensions` into `Optimizer` and call it for GP based regressors but not others. It feels like we should try and limit how much the assumptions for one type of regressor spread around the code base.

yngtodd:
> IMHO Optimizer is for people who know what they are doing, so they need to explicitly do things (IMHO also call cook_estimator themselves and pass that into Optimizer)... It feels like we should try and limit how much the assumptions for one type of regressor spread around the code base.

Would this mean stripping the Optimizer of all its default arguments (`base_estimator="gp"`, `acq_func= "gp_hedge" `, `acq_optimizer= "auto"`)? You could see the Optimizer as it is now as the `PreconfiguredOptimizer`. I could give it a shot at writing a more stripped down version Optimizer that makes less assumptions about how the Optimizer would be initially setup. I could open that as another pull request. 

>Is it strictly required to normalise the dimensions for it to work with GP based regressors? Or if you don't do it will it work but give worse results?

Looking back at https://github.com/scikit-optimize/scikit-optimize/issues/402, If you run the Optimizer without transforming the dimensions, the mu_Optimizer is way off from the true, unknown function. Check out below:

sans_transforms

MechCoder:
> Is it strictly required to normalise the dimensions for it to work with GP based regressors? Or if you don't do it will it work but give worse results?

It will work but give worse results, this is because the `ConstantKernel` term might be dominated by the feature that has a higher range. (Not completely sure that the length scale of each feature takes care of that)

@yngtodd Tim and I had a discussion offline and arrived at the consensus that ideally setting `base_estimator="GP"` (the string GP and not the estimator GaussianProcessRegressor) should default to `gp_minimize`. I'm fine with moving the checks inside Optimizer. WDYT?

yngtodd:
That sounds good to me. 

To check that I am following, do you mean checking if the `base_estimator="GP"` and if the dimensions have been normalized? If so, would should it throw an exception? Would those checks be added to `_check_arguments`? In that case, we could then remove the call to `normalize_dimensions` and leave that up to the user before initializing the `Optimizer`

Am I on the right track there?

MechCoder:
Umm, no.

I just meant normalizing internally if `estimator` is set to `GP`. The current behaviour except that:

1. The line

```python
if base_estimator == "GP" or base_estimator == GaussianProcessRegressor:
```

is replaced by 

```python
if base_estimator == "GP":
```

and

2. Moving this if statement internally inside `Optimizer` as it was done before I suggested to change it.

Do I make sense?

skopt/utils.py
```diff
@@ -388,3 +391,62 @@ def point_aslist(search_space, point_as_dict):
...
+    """
+    if isinstance(base_estimator, str):
+        base_estimator = base_estimator.upper()
+        if base_estimator not in ["GP", "ET", "RF", "GBRT"]:
+            raise ValueError("Valid strings for the base_estimator parameter " +
+                             "are: 'RF', 'ET', 'GBRT', or 'GP'.")
+    elif not is_regressor(base_estimator):
+        raise ValueError("base_estimator has to be a regressor.")
+
+    if base_estimator == "GP" or base_estimator == GaussianProcessRegressor:
```
yngtodd: @MechCoder we would then remove the `base_estimator="GP"` argument, the checks for expected strings associated with that argument, and the check for the regressor, right? 

Basically, we would keep lines 432-449?

skopt/utils.py
```diff
@@ -388,3 +391,62 @@ def point_aslist(search_space, point_as_dict):
     ]
     return point_as_list
 
+
+def normalize_dimensions(dimensions, base_estimator="GP"):
+    """
```
yngtodd: @MechCoder we would then remove the `base_estimator="GP"` argument, the checks for expected strings associated with that argument, and the check for the regressor, right? 

Basically, we would keep lines 432-449?

skopt/utils.py
betatim: TBH I've forgotten/can't reconstruct what I was thinking :-/

betatim:
Starting to look good!

(needs a rebase)

skopt/optimizer/optimizer.py
```diff
@@ -154,6 +155,8 @@ def __init__(self, dimensions, base_estimator="gp",
         n_jobs = acq_optimizer_kwargs.get("n_jobs", 1)
         self.acq_optimizer_kwargs = acq_optimizer_kwargs
 
+        if base_estimator == "gp":
```
MechCoder:
```python
base_estimator.lower
```

MechCoder:
lgtm pending nitpick.

MechCoder:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/400

skopt/optimizer/optimizer.py
yngtodd: Meaning that "gp" is preferred over "GP"?

betatim: Don't think there is a preference, just that it has to work with both cases :)

skopt/utils.py
```diff
@@ -357,7 +360,6 @@ def point_asdict(search_space, point_as_list):
     """Convert the list representation of a point from a search space
     to the dictionary representation, where keys are dimension names
     and values are corresponding to the values of dimensions in the list.
-
```
betatim: don't delete me

MechCoder:
Cool. Merging! Thanks @yngtodd and congrats on your first contribution!

yngtodd:
Thanks @MechCoder! That is amazing.

skopt/optimizer/optimizer.py
```diff
@@ -165,6 +166,15 @@ def __init__(self, dimensions, base_estimator="gp",
         n_jobs = acq_optimizer_kwargs.get("n_jobs", 1)
         self.acq_optimizer_kwargs = acq_optimizer_kwargs
 
+        if n_random_starts is not None:
+            warnings.warn(("n_random_starts will be removed in favour of "
+                           "n_initial_points."),
+                          DeprecationWarning)
+            n_initial_points = n_random_starts
+
+        if base_estimator == "GP":
```
betatim: Need to add a `.upper()` here

PR #421, opened 2017-07-05
[MRG+1] update version to 0.4

### LINKED ISSUE #308: Using GaussianProcessRegressor fails for default values
```python

dimensions = [(-5.0, 10.0), (0.0, 15.0)]
gpr = GaussianProcessRegressor()
optimizer = Optimizer(
    dimensions=dimensions,
    base_estimator=gpr,
    n_random_starts=1,
    acq_func="EI",
    random_state=0)

optimizer.tell(optimizer.ask(), branin(x_cand))

AttributeError: 'Product' object has no attribute 'gradient_x'
```

fredcallaway:
Fixes #308. I went with 0.4 because that's the highest version in the change log.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ except ImportError:
     from distutils.core import setup
 
 setup(name='scikit-optimize',
-      version='0.3',
+      version='0.4',
       description='Sequential model-based optimization toolbox.',
       long_description=open('README.rst').read(),
       url='https://scikit-optimize.github.io/',
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ except ImportError:
     from distutils.core import setup
 
 setup(name='scikit-optimize',
-      version='0.4',
+      version='0.4-dev',
       description='Sequential model-based optimization toolbox.',
       long_description=open('README.rst').read(),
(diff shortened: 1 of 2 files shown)


betatim:
What do others think? It solves the problem of pip thinking it already has the version of `master` installed. I am 👍  for this.

Maybe we can add a `(unreleased)` to line 4 of the changelog? If only as a way for us to keep track of which is the latest released version.

betatim:
An alternative would be that in the code we add a third level to the version (`0.3.x`) while we work towards the release of v0.4. My assumption has always been that we only have two digits in the version because the last one is set to zero for releases.

glouppe:
> An alternative would be that in the code we add a third level to the version (0.3.x) while we work towards the release of v0.4

Or `0.4-dev`? Then release `0.4` when `0.4-dev` has matured enough?

MechCoder:
> Or 0.4-dev? Then release 0.4 when 0.4-dev has matured enough?

+1

betatim:
Thanks! And congratulations on your first contribution to scikit-optimize!

fredcallaway:
Haha, hopefully the next one will be a bit more substantive.

PR #423, opened 2017-07-06
[MRG] PEP8 all the things

betatim:
Should change anything or add new features ... just reduce the number of complaints from the linter.

### DIFF
--- a/skopt/tests/test_gp_opt.py
+++ b/skopt/tests/test_gp_opt.py
@@ -47,7 +47,8 @@ def test_gp_minimize_bench3(search, acq):
 
 
 @pytest.mark.fast_test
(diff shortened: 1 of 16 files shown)


skopt/tests/test_gp_opt.py
```diff
@@ -67,6 +68,5 @@ def test_n_jobs():
 @pytest.mark.fast_test
 def test_gpr_default():
     """Smoke test that gp_minimize does not fail for default values."""
-    gpr = GaussianProcessRegressor()
```
betatim: Is the fact that this regressor wasn't used a bug?

glouppe:
Nothing critical in here. Merging.

PR #430, opened 2017-07-11
[MRG] Quick fixup

betatim:
Fixes #420  to make parameter case insensitive.


betatim:
@MechCoder and @yngtodd could you have a quick look at this? Got lost in the last PR I think.

MechCoder:
THanks!

PR #432, opened 2017-07-12
[MRG+1] Implement greedy A-optimal acquisition function for pure exploration

kiudee:
This acquisition function aims at reducing the overall uncertainty of our objective function approximation.
This is useful if you want to accurately gauge the effect of every hyperparameter on the objective function, typically to set proper ranges for the subsequent optimization or to remove a parameter completely.

The `gaussian_a_opt` function uses the standard deviation provided by the base estimator and samples those points first where it is maximal.

Suggestions for improvement are welcome.

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -105,11 +105,11 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
         It should have a ``return_std`` parameter that returns the standard
         deviation.
 
-    * `kappa`: [float, default 1.96 or 'Aopt']:
+    * `kappa`: [float, default 1.96 or 'inf']:
         Controls how much of the variance in the predicted values should be
         taken into account. If set to be very high, then we are favouring
         exploration over exploitation and vice versa.
-        If set to 'Aopt', the acquisition function will only use the variance
+        If set to 'inf', the acquisition function will only use the variance
         which is useful in a pure exploration setting.
         Useless if ``method`` is set to "LCB".
 
@@ -134,13 +134,13 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
                 X, return_std=True, return_mean_grad=True,
                 return_std_grad=True)
 
-            if kappa == "Aopt":
+            if kappa == "inf":
                 return -std, -std_grad
             return mu - kappa * std, mu_grad - kappa * std_grad
 
         else:
             mu, std = model.predict(X, return_std=True)
-            if kappa == "Aopt":
+            if kappa == "inf":
                 return -std
             return mu - kappa * std
 
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -105,10 +105,12 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
         It should have a ``return_std`` parameter that returns the standard
         deviation.
 
-    * `kappa`: [float, default 1.96]:
+    * `kappa`: [float, default 1.96 or 'Aopt']:
         Controls how much of the variance in the predicted values should be
         taken into account. If set to be very high, then we are favouring
         exploration over exploitation and vice versa.
+        If set to 'Aopt', the acquisition function will only use the variance
+        which is useful in a pure exploration setting.
         Useless if ``method`` is set to "LCB".
 
     * `return_grad`: [boolean, optional]:
@@ -132,10 +134,14 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
                 X, return_std=True, return_mean_grad=True,
                 return_std_grad=True)
 
+            if kappa == "Aopt":
+                return -std, -std_grad
             return mu - kappa * std, mu_grad - kappa * std_grad
 
         else:
             mu, std = model.predict(X, return_std=True)
+            if kappa == "Aopt":
+                return -std
             return mu - kappa * std
 
 
--- a/skopt/tests/test_acquisition.py
+++ b/skopt/tests/test_acquisition.py
@@ -71,6 +71,14 @@ def test_acquisition_pi_correctness():
 
 
 @pytest.mark.fast_test
+def test_acquisition_a_opt_correctness():
+    # check that it works with a vector as well
+    X = 10 * np.ones((4, 2))
+    aopt = gaussian_lcb(X, ConstSurrogate(), kappa='Aopt')
+    assert_array_almost_equal(aopt, [-1.0] * 4)
+
+
+@pytest.mark.fast_test
 def test_acquisition_lcb_correctness():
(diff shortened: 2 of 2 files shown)


iaroslav-ai:
Looks interesting!

Could you elaborate a bit more on particular use cases for the function, eg a bit more description of practical use cases? Also could you provide some references to the literature where such thing is used?  Would be good so that people can take a look at it a bit more in detail.

One idea I would have in my mind is to possibly use this instead of random initialization for the optimizers, so that initial points generated are distributed "more evenly" across search space.

kiudee:
The general setting is called [active learning](https://en.wikipedia.org/wiki/Active_learning_(machine_learning)) in which you want to learn the target function which as few evaluations as possible.

"A-optimality" was established in [optimal design](https://en.wikipedia.org/wiki/Optimal_design#Minimizing_the_variance_of_estimators) . The goal is to specify design points in advance which reduce the average variance of the parameter estimates. See [1] for a good treatment of the different optimality criteria when applied in Bayesian optimization. This reference could also be useful if we want to implement more criteria like the mutual information.

For initialization we could calculate a fixed set of `n_random_starts` points to implement an optimal design.
I would advise against using the surrogate model for that purpose.
For quasi-random initialization I would recommend a sequence of points satisfying low-discrepancy (see [2] for a recent paper on quasi-monte carlo integration). This captures your intuition of "more evenly" exploring the search space.
The library [Spearmint](https://github.com/JasperSnoek/spearmint) uses a [Sobol sequence](https://en.wikipedia.org/wiki/Sobol_sequence) for initialization. I would recommend choosing a random start value of the sequence, otherwise it will always start with the exact same points.

[1] Krause, Andreas, Ajit Singh, and Carlos Guestrin. "[Near-optimal sensor placements in Gaussian processes: Theory, efficient algorithms and empirical studies.](http://www.jmlr.org/papers/v9/krause08a.html)" Journal of Machine Learning Research 9.Feb (2008): 235-284.
[2] Dick, Josef, Frances Y. Kuo, and Ian H. Sloan. "[High-dimensional integration: the quasi-Monte Carlo way.](https://www.cambridge.org/core/journals/acta-numerica/article/highdimensional-integration-the-quasimonte-carlo-way/03F126DDF465F915B22D5D709CD28946)" Acta Numerica 22 (2013): 133-288.

betatim:
Naive question: how is this acquisition function different from evaluating the objective using a Sobol (or your favourite quasi random) sequence? Is it because with a Sobol sequence you explore the space "evenly" and here you pick points that have large uncertainty? Is there a simple example where the two don't lead to "the same" thing? (heteroscedasticity?)

MechCoder:
Hmm, I think you ca achieve the same by setting `Kappa` to a very-high value in `LCB`. Is it not?

kiudee:
@betatim I will play around with a few GPs to come up with an example where the behavior is different. In any case the Sobol sequence is not adaptive, ie it will not change if the user provides an initial set of points for which the objective value is already known.

@MechCoder Yes, indeed I was doing exactly this as a workaround before deciding to implement the acquisition function. In my opinion it is cleaner this way, since the effect of the mean is completely removed.

MechCoder:
In that case, I would prefer having a special value for `Kappa` that will set exploitation to zero (and will have no controversy in getting merged) instead of having yet another acquisition function.

kiudee:
Ok, that sounds like a good compromise. I can make the change next week
since I'm going on vacation today.

kiudee:
I made the change, by letting the user provide a special string `'Aopt'` as the parameter `kappa` in LCB.

Somehow Github did not like that I rebased the commits and force-pushed. Any ideas on how to fix the pull request without recreating it?

glouppe:
Looks good to me. +1 for merge

skopt/acquisition.py
```diff
@@ -105,10 +105,12 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
         It should have a ``return_std`` parameter that returns the standard
         deviation.
 
-    * `kappa`: [float, default 1.96]:
+    * `kappa`: [float, default 1.96 or 'Aopt']:
         Controls how much of the variance in the predicted values should be
         taken into account. If set to be very high, then we are favouring
         exploration over exploitation and vice versa.
+        If set to 'Aopt', the acquisition function will only use the variance
```
MechCoder: Sorry for being a prick but is `Aopt` the best name?
`

kiudee: I agree, since we do not have any other acquisition functions approximating optimal designs, we could call it something like `'var'`, `'variance'`, `'var_only'` or `'explore_only'`. I am open to suggestions.

glouppe: `"variance"` is fine with me.

skopt/acquisition.py
```diff
@@ -105,10 +105,12 @@ def gaussian_lcb(X, model, kappa=1.96, return_grad=False):
         It should have a ``return_std`` parameter that returns the standard
         deviation.
 
-    * `kappa`: [float, default 1.96]:
+    * `kappa`: [float, default 1.96 or 'variance']:
         Controls how much of the variance in the predicted values should be
         taken into account. If set to be very high, then we are favouring
         exploration over exploitation and vice versa.
+        If set to 'variance', the acquisition function will only use the variance
```
MechCoder: Sorry again, but this should be `std'?

kiudee:
After looking at the scikit-optimize documentation I would propose calling
it uncertainty which is the term used in the introduction of Bayesian
optimization. Though technically true that we pick the points maximizing
standard deviation and equivalently variance, I would say it is more
consistent to use uncertainty. Thoughts?

skopt/acquisition.py
iaroslav-ai: Do you talk about the name of acquisition function? Some might have weird associations with 'std' as abbreviation :sweat_smile:

MechCoder:
So the confusion on my side is because `kappa` denotes the value by which the std is multiplied and not the acquisition function itself.

I would be fine with allowing `kappa="inf"` or/and `kappa=np.inf` with a note that says this sets off exploitation. WDYT?

glouppe:
No strong opinion from my side. Either way is fine for me.

(sent from my phone)

kiudee:
I would be fine calling it `'inf'` and explaining it in the docstring.

Gilles Louppe <notifications@github.com> schrieb am Do., 27. Juli 2017 um
07:53 Uhr:

> No strong opinion from my side. Either way is fine for me.
>
> (sent from my phone)
>

glouppe:
Good to go for now when Travis is happy.

kiudee:
The Travis build canceled due to
`The job exceeded the maximum time limit for jobs, and has been terminated.`

MechCoder:
Thanks!

PR #437, opened 2017-07-14
[MRG+2] Allowing numpy int and float typed tuples in dimensions

yngtodd:
Allows `check_dimension` to handle tuples of numpy integers and floats. 

In cases where a tuple passed to `check_dimension` contains numpy int and float types, the function throws a confusing error where it looks like you are not allowed to pass a tuple of integers: 

screen shot 2017-07-14 at 10 34 27 am

I have been running into this when working on distributed hyperparameter optimization. It becomes an issue when you are defining dimensions by creating tuples from numpy ndarrays.


betatim:
Does using `numbers.Integral` or `numbers.Real` (https://docs.python.org/3.6/library/numbers.html) work? If we don't have to define our own collection of types that would be great.

yngtodd:
Oh, nice. Yes, that seems to work. Will update.

betatim:
Could you also add a test for `check_dimensions` to prevent this bug from coming back in the future please.

yngtodd:
Using `numbers.Integral` and `numbers.Real` is causing one test case to fail in `test_space_consistency`. 

```
# Failing test case
s1 = Space([Real(0.0, 1.0)])
s6 = Space([(0, 1.0)])

assert_equal(s1, s6)
```
The test makes sure that the dimension created by the tuple takes the greater precision offered by the float. 

When using `numbers.Integral` and `numbers.Real` there are two order options:
1. Check for real valued tuples first. 
* Since the integers are a subset of the reals, this turns all tuples of integers into `Real` dimensions (this is certainly not a good idea)
2. Check for integer valued tuples first.
* This causes the above test case to fail. The resulting dimension is of type `Integer`.

MechCoder:
I cannot reproduce this on master with Python 3 and NumPy 1.12.1

```python
space = Space([(np.float(0.0), 1.0)])
space.dimensions
[Real(low=0.0, high=1.0, prior=uniform, transform=identity)]

space = Space([(np.int(0.0), 1.0)])
space.dimensions
[Real(low=0, high=1.0, prior=uniform, transform=identity)]

space = Space([(np.int(0.0), np.int(1.0))])
space.dimensions
[Integer(low=0, high=1)]
```

yngtodd:
Just double checked you are right. It can be reproduced with the following:

```
import numpy as np
from skopt.space import *

arr = np.arange(1, 5)
tup = (min(arr), max(arr))

type(tup) # tuple
type(tup[0]) # np.int64

space = Space(tup) # raise ValueError("Dimension has to be a list or tuple.")
 check_dimension(tup) # ValueError: Invalid dimension (1, 4). Read the documentation for supported types.
```
Creating the tuple with `(min(arr), max(arr))` is a bit awkward. I am not sure how to better reproduce this.

MechCoder:
I see, this (https://github.com/numpy/numpy/issues/2951) seems to be the "problem". Checking for `numbers.Integral / numbers.Real` seems to be the way to go.

```python
a = np.int64(3)
isinstance(a, int)
False
a = np.int(3)
isinstance(a, int)
True
```

You might want to do change the test to reflect the behvaiour that fails in master, that is use `np.int64` instead of `np.int`

skopt/space/space.py
```diff
@@ -69,10 +69,10 @@ def check_dimension(dimension, transform=None):
     if len(dimension) == 2:
         if any([isinstance(d, str) for d in dimension]):
             return Categorical(dimension, transform=transform)
-        elif any([isinstance(dim, float) for dim in dimension]):
```
MechCoder: If I read correctly both the orderings are the same?

yngtodd: In the old version it checks for floats before integers. With the new it checks for integers first. If not, any intended `Integer` dimension will actually be transformed into a `Real` dimension.

MechCoder: Sorry, I'm dense.

Can you give me an example?

yngtodd: Since it's a little bit lengthy to modify the `check_dimension` check out the gist here:

https://gist.github.com/yngtodd/c8123a49d55a4ab0c505b3d99a472706

MechCoder: Oh yeah, duh forgot that that is `numbers.Real` and not `float`

MechCoder:
lgtm

iaroslav-ai:
lgtm too!

MechCoder:
Thanks! and nice catch

yngtodd:
Great! Thank you!

PR #443, opened 2017-07-20
[MRG+1] DeadlineCallback, early stopping for a fixed budget

yngtodd:
Creates a new `DeadlineCallback` class, an early stopping callback for fixed time budgets. When running optimizations on our clusters, we have fixed wall times for the jobs to be completed within. After pushing the clock a couple of times, I decided to write this class. 

After specifying the total amount of time you would like the optimization to run, the callback keeps track of the time taken for each iteration (exactly as in `TimerCallback`). It then decrements the iteration time from the total time, saving this as `time_remaining`. The early stopping occurs when the `time_remaining` is less than or equal to longest single iteration time. 

> Note: This code has some repetition from the `TimerCallback`. I originally tried to use multiple inheritance with both the `TimerCallback` and the `EarlyStopper` class, but ran into issues working around both parent classes' `__call__` methods. I also tried using composition as well a combination of inheritance and composition without much luck. If you guys have any thoughts on using multiple inheritance here, I am all ears!

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -142,7 +142,6 @@ class TimerCallback(object):
         elapsed_time = time() - self._time
         self.iter_time.append(elapsed_time)
         self._time = time()
-        print("I am from TimerCallback!")
 
 
 class EarlyStopper(object):
@@ -157,7 +156,6 @@ class EarlyStopper(object):
         * `result` [`OptimizeResult`, scipy object]:
             The optimization as a OptimizeResult object.
         """
-        print("You will have a criterion!")
         return self._criterion(result)
 
     def _criterion(self, result):
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -249,8 +249,8 @@ class DeadlineStopper(EarlyStopper):
         return self._criterion(result)
 
     def _criterion(self, result):
-        if len(result.x_iters) >= 1:
+        if result.x_iters:
             time_remaining = self.total_time - np.sum(self.iter_time)
             return time_remaining <= np.max(self.iter_time)
         else:
-            None
+            return None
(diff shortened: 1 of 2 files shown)


glouppe:
Looks like something useful! Could you add tests?

betatim:
Ready for review? (change title to [MRG] if yes)

skopt/callbacks.py
```diff
@@ -220,3 +220,37 @@ def _criterion(self, result):
...
+        self.total_time = total_time
+
+    def __call__(self, result):
+        elapsed_time = time() - self._time
+        self.iter_time.append(elapsed_time)
+        self._time = time()
+        return self._criterion(result)
+
+    def _criterion(self, result):
+        if len(result.x_iters) >= 1:
```
betatim: should be `if result.x_iters` to test if the list contains something

skopt/callbacks.py
```diff
@@ -220,3 +220,37 @@ def _criterion(self, result):
...
+            time_remaining = self.total_time - np.sum(self.iter_time)
+            return time_remaining <= np.max(self.iter_time)
+        else:
+            None
```
betatim: missing `return`?

skopt/callbacks.py
```diff
@@ -220,3 +220,37 @@ def _criterion(self, result):
...
+        finish within.
+    """
+    def __init__(self, total_time):
+        super(DeadlineStopper, self).__init__()
+        self._time = time()
+        self.iter_time = []
+        self.total_time = total_time
+
+    def __call__(self, result):
+        elapsed_time = time() - self._time
```
betatim: You could move all this logic to `_cirterion` if you wanted to and then not have a `__call__`.

betatim:
Thanks for putting up with my nitpicks :)

I restarted travis as it failed because of a timeout. Let's see if we can get a greenlight!

glouppe:
Looks good to me. +1

betatim:
🎆

yngtodd:
Great, thanks everyone!

PR #444, opened 2017-07-20
[MRG] Fix bug in has_gradients

betatim:
Fix bug in `has_gradients` and add tests for it at the same time.

`has_gradients` was returning `not has_gradients` as far as I could tell.

One thing I'd like to double check, the special case for `space.is_categorical` when constructing the GP kernel: Hamming kernels are only for spaces which have exclusively categorical dimensions. Is that correct?

### DIFF
--- a/skopt/__init__.py
+++ b/skopt/__init__.py
@@ -59,6 +59,7 @@ from .optimizer import gbrt_minimize
 from .optimizer import gp_minimize
 from .optimizer import Optimizer
 from .searchcv import BayesSearchCV
+from .space import Space
 from .utils import dump
 from .utils import expected_minimum
 from .utils import load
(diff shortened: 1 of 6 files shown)


betatim:
Surprised by the many test failures this has caused :-/ @MechCoder did you write the original version of `has_gradients`? I think for all the tree based ones the answer is: no gradients and for the GP based ones it depends on which kernel you use right?

MechCoder:
In addition, it should be `if not has_gradients(self.base_estimator)` here (https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/optimizer.py#L224)

(Two wrongs make a right)

MechCoder:
Thanks!

PR #445, opened 2017-07-21
[WIP] Parallel Bayesian Optimization

kejiashi:
Implementation of the parallel Bayesian optimization algorithm proposed by:
> Snoek et al. 2012, Practical Bayesian Optimization for Machine Learning

Project Tracking:
- [x] Locate positions to add new functions
- [ ] Add fantasizing procedures before calculating EI
- [ ] |— 1. Edit gaussian_ei
- [ ] |— 2. Edit fant_predict
- [ ] |— 3. ...
- [ ] Test fantasizing behavior
- [ ] Build/check connections to other base_estimators
- [ ] Edit docstrings
- [ ] Connect to external sampler
- [ ] (?) Add internal sampler file
- [ ] Test overall behavior
- [ ] Edit docstrings


kejiashi:
The computation of the acquisition function in this algorithm is quite different from existing constant liar. Therefore the original `prediction` function inside the gp file has to change in order to meet this need.
My first question is: is it ok to write a new function inside the gp class (I'm naming it `fant_predict`) or it's better if we just write a separate function in the `Optimizer`?
Second: is it better if we still add this feature to the original `gaussian_ei` function or create a new maybe `gaussian_ei_fant` function?

kejiashi:
Never mind the previous questions. Just realized the right places to implement should be:

- Add fantasizing behavior inside the `ask` function instead of `gaussian_ei` or `predict`
- Then go on reset `y_opt` and `X` to new values

kejiashi:
I have a doubt in the `prediction` function within the `GaussianProcessRegressor` class. 

The original Snoek et al. 2012 paper searches for the global optimal vector `(x_n+1, x_n+2, x_n+3, ...)` as a whole: stack all pending evaluation points on top of all the evaluated points, calculate Cholesky decomposition and variance based on the new stacked data, truncate the new variance for further fantasizing behaviors...

As our `ask-and-tell` API, we take a "pseudo-sequential greedy" strategy (as stated in the `Kriging is well-suited...` paper), which sequentially lie and tell over all points. This makes it less intensive in the case where we have large `q` compared to the approach where we compute it as a whole. However, in the `prediction` we restrict `n_samples` to `1`. I think to implement this new feature, we have to add support for `n_samples > 1`.

So I'm just wondering if there're some other concerns about implementing support for multi-points when writing this restriction, since I could not track the development of this function down from closed PRs. :)

kejiashi:
Close to fix the branching issue. Will open a new one soon.

PR #451, opened 2017-07-26
[MRG] Shaving off some seconds from the tests

### LINKED ISSUE #449: Speed up tests, part deux
We get intermittent travis failures because the tests take too long to run 😢 

Short of just disabling all tests marked as "slow", what else can we do? Reduce number of iterations in minimizer loops?

betatim:
Fixes #449 

Trying to cut down on the combinatorics mostly.

### DIFF
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -157,8 +157,8 @@ def test_minimizer_api_random_only(minimizer):
 def test_fixed_random_states(minimizer):
     # check that two runs produce exactly same results, if not there is a
     # random state somewhere that is not reproducible
-    n_calls = 7
-    n_random_starts = 4
+    n_calls = 4
+    n_random_starts = 2
 
     space = [(-5.0, 10.0), (0.0, 15.0)]
     result1 = minimizer(branin, space, n_calls=n_calls,
@@ -177,8 +177,8 @@ def test_fixed_random_states(minimizer):
 def test_minimizer_with_space(minimizer):
     # check we can pass a Space instance as dimensions argument and get same
     # result
-    n_calls = 7
-    n_random_starts = 4
+    n_calls = 4
+    n_random_starts = 2
 
     space = Space([(-5.0, 10.0), (0.0, 15.0)])
     space_result = minimizer(branin, space, n_calls=n_calls,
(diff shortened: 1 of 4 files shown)


betatim:
This mostly cuts down the `n_calls` parameters to speed things up.

A few tests I also switched from all `*_minimize` functions times all acquisition functions to just all minimize functions as the parametrisation. I think this should be fine but if someone else could take a look that would be great.

skopt/tests/test_common.py
```diff
@@ -227,7 +227,7 @@ def test_init_points_and_models(n_random_starts, optimizer_func):
     # starts (no y0 in this case)
     space = [(-5.0, 10.0), (0.0, 15.0)]
     x0 = [[1, 2], [3, 4], [5, 6]]
-    n_calls = 8
+    n_calls = 7
```
MechCoder: I think you can still reduce this to `n_calls=6`?

skopt/tests/test_common.py
```diff
@@ -244,7 +244,7 @@ def test_init_points_and_models(n_random_starts, optimizer_func):
 def test_init_vals(n_random_starts, optimizer_func):
     space = [(-5.0, 10.0), (0.0, 15.0)]
     x0 = [[1, 2], [3, 4], [5, 6]]
-    n_calls = 10
+    n_calls = len(x0) + n_random_starts
```
MechCoder: I would suggest this to reduce this rather to check the case where this exceeds len(x0) + n_random_starts

MechCoder:
Just a couple of minor comments. We should probably look at the difference in the speed of the tests before merging from now on

skopt/tests/test_common.py
betatim: So you mean `n_calls = len(x0) + n_random_starts + 1`?

skopt/tests/test_common.py
betatim: At the maximum we have `n_random_starts=4` and three initial points so we need seven calls. I think. At least we get an error if I reduce it more :-/

glouppe:
Let's merge this! Travis really is too slow :(

betatim:
image
 Let's enjoy it while it lasts.

MechCoder:
lgtm

Issue #433, opened 2017-07-12
Picking initial points with latin hypercubes


betatim:
Inspired by the comments in #432: currently we pick the initial points at random, and discussed using a Sobol sequence (or other quasi random sequence) instead.

Finding a good implementation of a QR sequence generator in many dimensions isn't soo easy.

However I just realised that we pick the number of initial points "up front", which means we can design a "optimal" grid (latin hypercube?) and evaluate the objective using that. Sobol or random only wins over this static allocation if you do not know how many points you will sample. This would be great because I think latin hypercubes should be much easier to code up than a Sobol sequence.

Would be good to hear some opinions on this.

kiudee:
With latin hypercube designs we should be careful (even a straight line of
points is a valid latin hypercube design) in how we construct the grid.
The R package 'lhs' implements a few methods to construct latin hypercube
designs satisfying different criteria like mean distance of each point to
all other points.

I agree with the point that since we have a fixed set of points, we should
try to optimize their position instead of using quasi random sequences.

betatim:
I thought for a 1D problem it would be fine to use an evenly spaced line of numbers?

Then in N-dimensions the property you want is that none of the points overlap so that when you project onto a particular axis you get `x` evaluations (if you have `x` points in total). (I am getting close to the edge of my experience here though)

kiudee:
Yes, in 1D this already suffices. In more dimensions you will increasingly find large volumes of space which are not covered if the latin hypercube is chosen randomly.

For 2D consider this example:
![LHS](https://user-images.githubusercontent.com/564473/28155405-de659b58-67ae-11e7-9e6a-048071e035ac.png)
None of the points is overlapping and the 1D marginals are covered evenly. Yet, large areas are not covered and the GP subsequently will have a large uncertainty in these areas.

glouppe:
Do you have a benchmark (or reference) showing that this is better than
uniformly random points? I remember @MechCoder playing a similar idea
(pseudo-random initialization points), which we eventually abandoned
because we  couldnt show this was better.

iaroslav-ai:
I agree with @glouppe . We should replace things only if we can show that they improve on existing stuff.

kiudee:
I will start by comparing different initializations on the branin function,
to see whether there is a difference already for low-dimensional functions.

Iaroslav Shcherbatyi <notifications@github.com> schrieb am Do., 13. Juli
2017 um 11:12 Uhr:

[...]

Issue #415, opened 2017-06-28
Support noisy EI (NEI)


kiudee:
Using the current best objective value or the GP mean/quantile estimated at that point is problematic when using EI to optimize noisy target functions. Often this leads to premature exploitation.

Letham et al. propose the _noisy expected improvement_ function which integrates the EI over the GP posterior:
[Constrained Bayesian Optimization with Noisy Experiments](https://arxiv.org/abs/1706.07094v1)

The authors compare their acquisition function to Spearmint EI and PESC (predictive entropy search) and obtain very convincing results:
![NEI results](https://user-images.githubusercontent.com/564473/27628151-67b421e4-5bee-11e7-8233-f0b206b51946.png)

The acquisition function can easily be approximated by sampling noiseless GPs from the GP posterior and averaging their EI. The authors additionally use low-discrepancy sequences to improve the integration (QMC sampling).

betatim:
Would you be interested in contributing a benchmark and NEI implementation?

stefanocereda:
I'm very interested in this since I'm working on a project where constraints are needed. I'll try to see if I can come up with an implementation.

However, from my understanding, the constraints considered in the paper are different from the ones of #355. In fact, 355 talks about constraints on the variables (e.g. x1+x2 < 10) whereas the constraints of the paper are about objective functions (e.g. optimize compiler parameters so as to minimize execution time but without having compilation time exceed 10 minutes)

betatim:
I think we could handle both cases (invalid parameter combination and compile time too long) by returning a "very large value" for the objective function. This might not be the absolute best strategy but could be a good baseline to start with. Mostly because it requires very little if any changes to `skopt`. Once there is a benchmark for this we could discuss other more specialised/sophisticated methods and see how much they buy us in terms of performance and how much it costs in terms of maintenance/extra code. WDYT?

stefanocereda:
I thinks that a first implementation without constraints is the best solution, even just to see if this function is actually good in real scenarios.

I'll be able to work on this just on my free time, but I think that this first implementation should not take much time. The pseudocode of the tell function should be something like:
- for each previously observed point sample from the GP the mean and the variance
- build the vector of means and the covariance matrix (putting the vector of variances on diagonal)
- using np.random.multivariate_normal generate N "variants" of the observed data (better version: do as in the paper and use sobol sequences)
- for each variant of data build a gp
- the EI of a point is the average of all the computed GPs

Am I missing something?

[...]

PR #452, opened 2017-07-27
[WIP] implementation of noisy_EI

stefanocereda:
Dummy implementation of noisy EI as discussed in https://github.com/scikit-optimize/scikit-optimize/issues/415.
This is still a WIP, I'm opening just to discuss.

Still need to:
	- implement sobol sequences for better sampling
	- documentation
	- can we cache "sampled" GPs?


kiudee:
Thank you for the pull request.

For reference, here is the version of the nEI, I have been experimenting with:
```python
from SALib.sample.sobol_sequence import sample
def gaussian_nei(X, model, n_samples=10, y_opt=0.0, xi=0.01, return_grad=False):
    n_train = len(model.X_train_)
    mu, cov = model.predict(model.X_train_, return_cov=True)
    L = np.linalg.cholesky(cov + np.eye(n_train) * 1e-10)

    u = sample(n_samples, n_train)

    GPs = norm.ppf(u).dot(L.T) + mu

    # Fit `n_samples` new noiseless GPs for which we will average the
    # expected improvement:
    new_model = clone(model)
    new_model.noise = 1e-10
    values = 0.0
    if return_grad:
        grad = np.zeros(X.shape[1])
    for i in range(n_samples):
        new_model.fit(model.X_train_, GPs[i])

        v = gaussian_ei(X, new_model, y_opt=y_opt, xi=xi,
                        return_grad=return_grad)
        if return_grad:
            v, g = v
        values += v / n_samples
        if return_grad:
            grad += g / n_samples

    if return_grad:
        return values, grad

    return values
```

I have been comparing all of the implementations of low-discrepancy sequences for python and have to say that the only viable one is the one in [SALib](https://github.com/SALib/SALib) since all the other ones only support <= 40 dimensions.

stefanocereda:
I'm now running some benchmark tests, when they complete I will also include your version to test the sobol sequence.

Just a couple of questions:
 - you directly ask the model to provide the covariance matrix, whereas the paper just use the std dev of the sampled points. Do you think we should run some tests to compare the two approaches?
 - by putting the code for sobol sampling + gp fitting in the aquisition function code I think we would end up recomputing the same things multiple times. For example when using lbfgs the acquisition function isa called many times, whereas if we do this in the tell function we just do that once. What do you think?

kiudee:
Yes, doing the fitting outside of the optimization is a very good idea since the fitting is very slow compared to the rest of the code.

test_nei.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
kiudee: Can you merge these tests into `test_acquisition.py`?

stefanocereda: I commited that file by mistake, it's just the one I'm using to see if this is actually working. I will remove it in the next commit as I don't think it's relevant as a test. It just takes some hours and produce plots.

glouppe:
Thanks for working on this! This looks very promising. Let us know when you want us to do a first round of reviews.

stefanocereda:
immagine

I have made some tests on branin+noise. Those are the best _so_far values averaged across 20 repetitions of epochs 11-20. It actually seems like a good acquisition function.

I now want to understand if we really need to resample and recompute GPs at every tell() or if we can cache the old ones and just add new points, as this would greatly speed up things.

Any ideas about the noisyEI_N_variants parameter?

kiudee:
>you directly ask the model to provide the covariance matrix, whereas the paper just use the std dev of the sampled points. Do you think we should run some tests to compare the two approaches?

>EDIT: I'm clearly wrong on the std.dev / covariance. Sorry

No problem, the authors used the slightly misleading notation `diag(Σf, ...)` for the block diagonal matrix.

stefanocereda:
I've tested this code on a real problem and it seems to give good results. Please let me know if the code is ok to you.

kiudee:
Looks good so far. I am going to try your implementation later this day.

Regarding Sobol sampling, we should wait for the implementation of #433 for an official interface.

skopt/optimizer/optimizer.py
```diff
@@ -153,9 +156,13 @@ def __init__(self, dimensions, base_estimator="gp",
         else:
             self.cand_acq_funcs_ = [self.acq_func]
 
+        if self.acq_func == "noisyEI":
+            self.noisyEI_N_variants = noisyEI_N_variants
```
kiudee: `noisyEI_N_variants` is not defined here

stefanocereda: Sooner or later I will learn to use git propery......
Those 2 lines can be removed, I'm setting it on line 165

skopt/optimizer/optimizer.py
```diff
@@ -446,6 +450,31 @@ def tell(self, x, y, fit=True):
...
+                    noisy_est = clone(self.base_estimator_)
+                    with warnings.catch_warnings():
+                        warnings.simplefilter("ignore")
+                        noisy_est.fit(self.space.transform(self.Xi),
+                                      yis[variant])
+                    noisy_estimators.append(noisy_est)
+
+                # use this list as estimator, it will be handled by the
+                # acquisition function
+                est = noisy_estimators
```
kiudee: This has the disadvantage that the result object returned after optimization only containts the sampled noiseless GPs and not the original noisy GP.
In my opinion result.models should always only contain the original GPs. Maybe we need to introduce a new variable for the variants.

glouppe:
Could we fork out the changes regarding random sampling with a Sobol sequence? (in order to fix #433 independently of adding a new acquisition function)

MechCoder:
@stefanocereda Thanks for your work! I had some questions for my understanding.

1. In your graph does epoch 1 correspond to after the initial random points? Can you provide the same 10 points to all the four variants as start points so they have the same information when they start?
2. Can you also provide a plot with minimum obtained till epoch `x` with error bars? 
3. Also, as the number of models increase, the EI estimate should be better? What is your intuition on the graph with 10 models lying above the ones with 1 and 5?

stefanocereda:
Yes epoch 1 is after random, basically it is epoch 11. I will provide the requested graphs as soon as I can. I'm still experimenting with the number of models, but I think that fewer models are better in the first stages, whereas they can be increased later on.

For the sobol sampling should I go back to basic random sampling? I can just leave the sobol sampling commented so it can be added back when a Sobol library is decided for the whole project.

For the return value I made it that way because I need all the sampled estimators, but if you prefer I can change it so we return the original GP.

I'm sorry for the slow pace but I'm quite full right now.

skopt/optimizer/optimizer.py
```diff
@@ -446,6 +450,31 @@ def tell(self, x, y, fit=True):
                 warnings.simplefilter("ignore")
                 est.fit(self.space.transform(self.Xi), self.yi)
 
+            # if we are using noisy expected improvement generate N variants
+            # of the observed data and compute a GP for each variant.
+            # The variable names reflect the notation used in the original 
+            # paper (Algorithm 1)
+            if self.acq_func == "noisyEI":
+                mu, sigma = est.predict(self.Xi, return_cov=True)
```
MechCoder: It might also be a good idea to check if we achieve similar performances just by sampling randomly, you can use the `sample_y` function to sample from the posterior directly.

MechCoder:
> I'm sorry for the slow pace but I'm quite full right now.

That's all right. All of us are ;)

skopt/optimizer/optimizer.py
stefanocereda: Also need to do self.space.transform(self.Xi) in the est.predict

stefanocereda:
I've run some more tests on this as the graphs above were not really statistically sound.
I've run 100 epochs of optimization and saved the y obtained at each step. This was repeated 50 times and averaged.
In the first graph I've used branin+noise and called gp_minimize with noise='gaussian'
In the second graph I've specified noise=1.0

image
image

I do not think this result is promising...I've saved all the results, so I can produce other graphs if needed

Issue #413, opened 2017-06-23
Do a speed profile


MechCoder:
We should do a speed profile, to identify the time taken in each part so as to see if we can speed up some obvious parts.

yngtodd:
I was interested in this too when looking at `gp_minimize` for hyperparameter optimization. 

Here are some [first steps](https://github.com/yngtodd/skopt_notes/tree/master/speed_test) using [rkern's line_profiler](https://github.com/rkern/line_profiler).

In that directory you will find 

* [`speed_test.py`](https://github.com/yngtodd/skopt_notes/blob/master/speed_test/speed_test.py): the example hyperparameter optimization found on [scikit-optimize.github.io](https://scikit-optimize.github.io/notebooks/hyperparameter-optimization.html). 

* [`profile_results.txt`](https://github.com/yngtodd/skopt_notes/blob/master/speed_test/profile_results.txt): the readout from the profiler. This gives the line-by-line breakdown for the time taken by each function used by `speed_test.py`.

MechCoder:
Thanks for your benchmarks!

It seems odd that this line (https://github.com/yngtodd/skopt_notes/blob/master/speed_test/profile_results.txt#L754) takes 23% of time, i.e 51 seconds in the method `tell`. Maybe that can be reduced?

betatim:
What does `_gaussian_acquisition` use `X` for? Maybe we can use a `linspace` instead (which might be quicker). Then we only need to sample such a large number of points when using "sampling" as optimizer for the acquisition function. For BFGS we can sample only `n_restarts_optimizer` points which is usually much smaller than `n_points`.

MechCoder:
The start points of lbfgs are obtained from `X` corresponding to the optimal acquisition function values from the random grid.

Apart from that, I'm unsure whether a uniform grid is better than a random grid

betatim:
The starting points for BFGS are taken from `X`, but there are usually many less starting points then the size of `X`. So we could not taken them from `X` and only generate the much smaller number needed.

What I was wondering is why `X` is passed to `_gaussian_acquisition`.

To illustrate what I mean: #

PR #453, opened 2017-07-28
[MRG] Reduce number of points sampled when using LBFGS

betatim:
When using LBFGS as minimizer for the acquisition function we do not
need to sample as many points as when we optimise by random sampling.
This saves CPU time.

This illustrates https://github.com/scikit-optimize/scikit-optimize/issues/413#issuecomment-318578862

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -450,8 +450,12 @@ class Optimizer(object):
                 self.gains_ -= est.predict(np.vstack(self.next_xs_))
             self.models.append(est)
 
-            X = self.space.transform(self.space.rvs(
-                n_samples=self.n_points, random_state=self.rng))
+            if self.acq_optimizer == "sampling":
+                X = self.space.transform(self.space.rvs(
+                    n_samples=self.n_points, random_state=self.rng))
+            elif self.acq_optimizer == "lbfgs":
+                X = self.space.transform(self.space.rvs(
+                    n_samples=self.n_restarts_optimizer, random_state=self.rng))
             self.next_xs_ = []
             for cand_acq_func in self.cand_acq_funcs_:
                 values = _gaussian_acquisition(
@@ -467,7 +471,7 @@ class Optimizer(object):
                 # minimization starts from `n_restarts_optimizer` different
                 # points and the best minimum is used
                 elif self.acq_optimizer == "lbfgs":
-                    x0 = X[np.argsort(values)[:self.n_restarts_optimizer]]
+                    x0 = X[np.argsort(values)]
 
                     with warnings.catch_warnings():
                         warnings.simplefilter("ignore")
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -450,12 +450,11 @@ class Optimizer(object):
                 self.gains_ -= est.predict(np.vstack(self.next_xs_))
             self.models.append(est)
 
-            if self.acq_optimizer == "sampling":
-                X = self.space.transform(self.space.rvs(
-                    n_samples=self.n_points, random_state=self.rng))
-            elif self.acq_optimizer == "lbfgs":
-                X = self.space.transform(self.space.rvs(
-                    n_samples=self.n_restarts_optimizer, random_state=self.rng))
+            # even with BFGS as optimizer we want to sample a large number
+            # of points and then pick the best ones as starting points
+            X = self.space.transform(self.space.rvs(
+                n_samples=self.n_points, random_state=self.rng))
+
             self.next_xs_ = []
             for cand_acq_func in self.cand_acq_funcs_:
                 values = _gaussian_acquisition(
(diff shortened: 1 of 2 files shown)


glouppe:
Good catch. +1

skopt/optimizer/optimizer.py
```diff
@@ -467,7 +471,7 @@ def tell(self, x, y, fit=True):
                 # minimization starts from `n_restarts_optimizer` different
                 # points and the best minimum is used
                 elif self.acq_optimizer == "lbfgs":
-                    x0 = X[np.argsort(values)[:self.n_restarts_optimizer]]
+                    x0 = X[np.argsort(values)]
```
betatim: Maybe someone with more randomness intuition can help out here: is this semantically the same? Picking (say) top five from a random list or sampling five points at random. Do the two cases have the same chances of finding a good point? I think by sampling many you have a higher chance of finding good ones ...

I *think* most papers actually talk about "pick five points at random and use those" and not "sample many and then use the top five".

skopt/tests/test_common.py
```diff
@@ -422,7 +422,7 @@ def test_early_stopping_delta_x(minimizer):
     res = minimizer(bench1,
                     callback=DeltaXStopper(0.1),
                     dimensions=[(-1., 1.)],
-                    x0=[[0.1], [-0.1], [0.9]],
+                    x0=[[-0.1], [0.9], [0.1]],
```
betatim: This test ended up failing because we get other(?) random numbers and don't have two successive values of `x` be close enough. The first "random" point we draw is at `x=0.01` so by swapping the order of the initial points we hit the stopping criteria right away.

betatim:
Ready for a review and thoughts on the inline comment I left. Then ready for merge (from my side).

skopt/optimizer/optimizer.py
glouppe: Hmm, actually, I remember now. We did this on purpose. The idea was to do a very coarse optimization with many random points. Then start LBFGS from the best points. This was to mimick the implementation of Spearmint. 

In the end, I would rather leave this unchanged.

betatim: Ok. I will add a comment to jog our collective memory for next time.

I wonder if we always need this many points to kickstart LBFGS. Feels like we should be able to scale this by the number of dimensions instead of using a flat number. (In 3D after testing 10k points do we really still need to LBFGS??)

The other tradeoff we could think about is "we have a budget of (say) 10k evals of the acquisition function, would we find a better point by spending 10k/`N` evals inside `N` LBFGS loops instead of searching at random?"

iaroslav-ai: hmm maybe something like N_random_points = 10^min(4, dimensions)? But then I guess one should make a distinction for different surrogates, as for example for tree based ones you would want to use all 10^4 as you do not apply LBFGS afterwards. 

With the second point, do you mean that you would rather repeat more often LBFGS optimization instead of random sampling? Intuitively it seems that it depends on whether acquisition function has a lot of local sub optimal minima or flat areas where solver can get stuck, and hence random sampling should help avoid such areas by providing initialization close to optimal points (but this is somewhat handwaving argument).

MechCoder: Look at https://github.com/scikit-optimize/scikit-optimize/pull/284 for the benchmarks that lead to this change.

betatim: I read the two comments (https://github.com/scikit-optimize/scikit-optimize/pull/284#issuecomment-270078807 and following) as: "running BFGS after sampling doesn't really improve things much over purely sampling" and "BFGS with only `n_restarts_optimizer` random points isn't very good at all"

@iaroslav-ai yes for trees we can't use BFGS, so all we do is sample to try and find the optimum. I was wondering for the case where you can use BFGS if we would do better by distribution our budget of 10000 evaluations of the acquisition function differently between random sampling, different starting points, and iterations of each BFGS run

iaroslav-ai: yup fine by me.

glouppe:
What do we decide to do with this?

I am rather status quo. I am okay with losing some seconds if that better guarantees to sample the most promising next point.

betatim:
Think I agree, will change the PR to add a comment to the code so we remember next time we benchmark it and wonder why we did this.

I'm still confused by the benchmark results linked to in this comment https://github.com/scikit-optimize/scikit-optimize/pull/453#discussion_r130243315 My main question: running BFGS after random sampling doesn't improve the minimum??? Am I interpreting this the wrong way?

MechCoder:
Why do you think that way? "-3.2273 < -3.1416 < -2.9502" no? Or are you looking at some other benchmark?

betatim:
Yes, but the difference is so small :-/ maybe the problem is "too easy" as for branin there is even less difference. Basically, I find it counter-intuitive that the difference is so small. Convinced that we should be doing running BFGS after sampling :)

Merge once travis is happy?

MechCoder:
oh I see, by the way the true global minimum is at -3.322. We are pretty close to it in our current benchmarks. I think is a good idea to run the benchmarks from here on, once we make major changes to the codebase.

skopt/tests/test_common.py
MechCoder: This can be changed back?

glouppe: Strange that this changed anything, since those are the same points, but in a different order. 

I am merging in spite of this, that's not very important.

betatim:
Can you add the magic line for just the hart6 benchmark to `CONTRIBUTING.md`?

Issue #403, opened 2017-06-16
Parallel optimization as in original spearmint paper?


amueller:
Does scikit-optimize do parallel optimization as in [the spearmint paper](https://papers.nips.cc/paper/4522-practical-bayesian-optimization-of-machine-learning-algorithms.pdf) if not, would you be interested in it?
I have a student that wants to work on auto-ml over the summer and we're brainstorming ideas. Pointers welcome.

iaroslav-ai:
Currently "constant liar" parallelization approach is used, see eg [here](https://arxiv.org/pdf/1602.05149.pdf) (search for 'constant liar'), which is a bit different from the one in the reference. We used this approach due to its simplicity and as it can work with any of the surrogate models used to model the objective and uncertainty in skopt. We also currently support only batch parallelism. 
From my point of view, having a more advanced parallel optimization would be great, assuming it  outperforms current approach. I did not check the literature on this tbh, but for example approach in the referenced paper seems to eventually outperform constant liar. So to summarize yes on my side I think it would be great to have more better parallelization :)

iaroslav-ai:
P.S. some other ideas that might also be interesting for brainstorming: introduce knowledge transfer, which can be used eg to speed up search for good machine learning parameters, as in [here](http://papers.nips.cc/paper/5086-multi-task-bayesian-optimization.pdf) for example. Also [scalable black box optimization](http://proceedings.mlr.press/v37/snoek15.html) with neural networks seems pretty cool, maybe something that can be used for some wild tasks like joint selection of features, neural network architecture and learning parameters.  These are some things I would also be very happy personally to see in scikit-optimize :)

betatim:
I think it would be a nice addition. I'd also be OK with being a mentor to help them find their way around the code base.

betatim:
(#352 is another issue on a method of parallel optimisation (qEI))

Young-Todd:
@amueller, I am a student researcher at Oak Ridge National Lab and have been working on distributing the optimization methods in scikit-optimize on our clusters. I have a working approach that distributes smaller subregions of the hyperparameter search space for the sequential models to search over. 

https://github.com/scikit-optimize/scikit-optimize/issues/401 is related to that project where I am looking for the best way to store the many results of all of the distributed runs. 

I would be more than happy to talk with you more about what I have been working on if you are interested.

MechCoder:
[...]

PR #458, opened 2017-07-31
[WIP] Updated benchmarks and benchmark infrastructure 

iaroslav-ai:
This is relevant to any issue or PR where an improvement over existing algorithms in the repository is proposed, eg #433, #452, #403, #453  and others. Set of good benchmarks allows to compare different versions of skopt, and to establish whether there is an improvement (possibly on specific class of problems) with proposed changes.  

Running benchmarks for many repetitions is computationally demanding and can take hours for sufficient number of iterations, even for fast objectives. Hence the evaluation procedure in PR is parallelized and can be run on multiple machines with dask. Evaluation with dask was tested successfully with 32 machine cluster on EC2 spot fleet.

It is important to run benchmarks for sufficient number of iterations in order to average out noise for reliable comparisons. To access the quality of estimation of algorithm performance, 95% confidence interval is computed using [bootstrap](https://github.com/facebookincubator/bootstrapped). This allows to estimate whether difference between two algorithms is likely to not be just because of noise. Bootstrapping is used because distribution of minimum values found by algorithms appears to be non - Gaussian. 

Performance of algorithms should be evaluated on benchmarks that generalize well to practical applications of black box optimization, in order to ensure that performance metrics are relevant. Thus instructions are given on how to come up with new benchmarks. 

A bit of technical details: 

Every instance of benchmark optimization problem is represented as a class with `space` field and `__call__` that implements evaluation of some input point from space. 

All modules with definitions of benchmarks are located  in `tracks` module. Every module should have `problems` variable that lists all benchmarks in the module. 

Example evaluation with `hart6` and `branin`:

```python
from evaluation import parallel_evaluate, plot_results, calculate_metrics
from tracks.classic import Hart6, Branin
from skopt import dummy_minimize, forest_minimize

r = parallel_evaluate(
    solvers=[dummy_minimize, forest_minimize],
    task_subset=[Hart6, Branin],  # specify None to run on all problems
    n_reps=50,
    eval_kwargs={'n_calls': 50},
    joblib_kwargs={'n_jobs': -1, 'verbose': 10}
)

p = calculate_metrics(r)
p.to_csv('data.csv')
plot_results(r)
```

Status:
- [x] Parallel evaluation procedure
- [x] Computation of mean of results, as well as confidence intervals using bootstrap into pandas dataframe
- [x] Visualization of results using skopt's plot_convergence function
- [x] EC2 spot fleet dask cluster creation, management and destruction
- [x] Add 2 new non - trivial tasks
- [x] Recalculate results with GP, Forest and Dummy
- [x] Add results of benchmarks to the readme
- [x] Add initial instructions on benchmarks
- [ ] Add docstrings
- [ ] Iterate the benchmarks instructions
- [ ] Iterate the structure of the repository
- [ ] pep8 all things
- [ ] update requirements


iaroslav-ai:
Let me know what you think of this or if you have any questions. 
You can find updated benchmark results in `README`.
I might not be active for the next few days as I will likely have no reliable access to internet.

amueller:
Some thoughts on benchmarking: Have you thought about adding wrappers to benchmark against other implementations like hyperopt, SMAC and spearmint?
It would also be great to add a problem which requires non-stationary kernels.
Finally, it seems there are some other good synthetic examples out there. [Preferential Bayesian Optimization](https://arxiv.org/abs/1704.03651) uses Forrester, Six Hump Camel, Gold Stein and Levy.
You can find a whole [repository here](https://www.sfu.ca/~ssurjano/optimization.html). There's a paper that looks interesting [here](http://www.redcedartech.com/pdfs/SHERPA_Benchmark_0110.pdf).

I had also found some higher-dim problems, which I think would be important to add. Not sure if they are in the repo I cited above, I can't find the page any more :-/

amueller:
Can you post the results of your cloud benchmark?

betatim:
This is another large set of benchmark functions: https://github.com/sigopt/evalset

(Point of order: should we make a new issue to discuss benchmarks we should add and discuss here a change of infrastructure only?)

This is huge! I will need some time as well to read the PR, so no worries about being away.

One thing after a quick look during a train ride: we need a way to run the benchmarks locally as well/without a cluster within a reasonable amount of time. Maybe it is as easy as marking things as fast/slow. Thinking is that we want everyone who proposes a change that touches on performance to run these so at least the basics need to run "quickly" (and we shouldn't exclude people who don't have AWS tokens to burn/a cluster) otherwise no one will do it :-/

glouppe:
Shouldn't we take this PR as the opportunity to put benchmarks in a separate repo, as discussed in #i-dont-remember-which-issue?

amueller:
There's also a bunch of benchmark funcitons in HPOlib https://automl.github.io/HPOlib2/stable/synthetic_benchmarks.html

iaroslav-ai:
Thanks everyone for nice references!

@amueller yes, I plan to make wrappers to other software. It is quite easy, a wrapper should be simply a function that looks like `gp_minimize` and return similar result.

The results for benchmarks are [here](https://github.com/iaroslav-ai/scikit-optimize/blob/updated_benchmarks/benchmarks/README.md). So far looks good.

iaroslav-ai:
@betatim @amueller the synthetic benchmarks look interesting and I think I will include (subset) of them for the time being. 

However, I have in general 2 issues with these synthetic benchmarks:

1. They are not directly derived from practical problems, hence not guaranteed to represent "set of practical problems" accurately

2. Most of these objectives are noiseless, while most of practical problems (at least ones I have) are actually noisy.

My vision for how to come up with benches that satisfy 1 and 2 is to learn VAE or GAN or other generative model from the data of evaluation of expensive, "practical" optimization problems, to model accurately these objectives and their noise, while speeding up their evaluation.  

This is something I attempted in #433 with mapping params of DNN to score with VAE. So far kinda works, but still researchy.

iaroslav-ai:
@betatim I think this is a good idea to make a separate issue for this a tiny bit later, in order to not blow up this PR too much.

iaroslav-ai:
@betatim wrt to fast benchmarks this is a good point that you are making. We might cluster benchmarks into groups which represent some specific practical application domains, and from every cluster select 1 or 2 which appear to be "representable" of the whole cluster. That is, such subset for which if you improve then you improve on other benchmarks as well. These should probably be selected from the main set by seeing which ones are "representable" **iaroslav-ai's handwaving intensifies**

It is also good to select subset even with fast benchmarks, as repeating eg gp_minimize some 100 times takes quite some time on a single machine.

iaroslav-ai:
@glouppe eventually we will, I just work here for now as there seem no need so far to move to other repo.

amueller:
One thing I'd like to see from synthetic benchmarks is functions that require non-stationary kernels (e.g. require searching on a log-scale) and seeing how algorithms cope with irrelevant dimensions.

iaroslav-ai:
Hey everyone,
My schedule is likely to go bananas over next days, so do not be surprised of me being unresponsive. 
I moved this pr to temporary repo here: [https://github.com/iaroslav-ai/scikit-optimize-benchmarks](https://github.com/iaroslav-ai/scikit-optimize-benchmarks) .

I would suggest we move it to scikit-optimize/benchmarks eventually.  

In addition to this PR, repo contains ~50 artificial benchmarks from evalset, as well as wrappers to Spearmint, SMAC3, Hyperopt, GPyOpt. These wrappers have api similar to skopt, eg you can 

```python
spearmint_minimize(fnc, [Real(0, 1), Categorical(['a', 'b', 'c'])], n_calls=64)
```

There is also install script for ubuntu and the like to install everything. 

I will try to run the benchmarks over the next days, probably when the aws will finally register my personal account :-\ or maybe I will just run it on scaleway.

iaroslav-ai:
Comparison of `SMAC3`, `hyperopt`, `GpyOpt` and `scikit-optimize` on sigopt's evalset 
 and some of our benchmarks are available here:

[https://github.com/iaroslav-ai/scikit-optimize-benchmarks/blob/master/README.md](https://github.com/iaroslav-ai/scikit-optimize-benchmarks/blob/master/README.md)

I do not include `spearmint` as it is unfortunately too slow to evaluate in reasonable time on 60+ different optimization problems as it seems.

kejiashi:
I found this project closely related with this PR.

http://coco.gforge.inria.fr

> COCO (COmparing Continuous Optimisers) is a platform for systematic and sound comparisons of real-parameter global optimisers. COCO provides benchmark function testbeds, experimentation templates which are easy to parallelize, and tools for processing and visualizing data generated by one or several optimizers. The COCO platform has been used for the Black-Box-Optimization-Benchmarking (BBOB) workshops that took place during the GECCO conference in 2009, 2010, 2012, 2013, and in 2015-2017. It was also used at the IEEE Congress on Evolutionary Computation (CEC'2015) in Sendai, Japan.

This sounds like what we need. :)

iaroslav-ai:
COCO looks good. Offloading the maintenance of benchmark infrastructure is also good :) What I wander from the name is whether it can work with categorical / integer parameters. It would also be nice to do evaluations distributed on multiple machines (this specially appeals to impatient peoples like me :D). Will take a closer look.

betatim:
As I am just re-discovering this PR: should we keep discussing here or move to https://github.com/iaroslav-ai/scikit-optimize-benchmarks?

iaroslav-ai:
Lets keep it here - [https://github.com/iaroslav-ai/scikit-optimize-benchmarks](https://github.com/iaroslav-ai/scikit-optimize-benchmarks) is a temporary location for this repo. I think eventually this should be moved somewhere to scikit-optimize, but before that I will look into COCO to see if we can use it, remove the current cluster management code and add some more comments.

iaroslav-ai:
As @kejiashi suggested, https://github.com/numbbo/coco seems pretty good. If there is to be any benchmarking introduced, this should be used as a first step. More complicated things can be introduced later. I leave this open as a reminder for benchmarking.

Issue #456, opened 2017-07-31
Mutual information for Gaussian process acquisition functions?


yngtodd:
I know that there are quite a few acquisition functions already supported, but what does everyone think about including a mutual information acquisition function?

I was reading through [[1]](http://proceedings.mlr.press/v32/contal14.pdf) and thought that it looked promising. The authors showed that using mutual information can consistently reduce the upper bounds for cumulative regret compared with GP-UCB and GP-EI. 

I know that @kiudee mentioned this in https://github.com/scikit-optimize/scikit-optimize/pull/432. What do you guys think?

glouppe:
I think this would be a good thing to have. Adding one more acquisition function shouldn't be something that is too intrusive.

PR welcome :)

Issue #118, opened 2016-07-19
References


betatim:
Start a collection of references (in the `__doc__` of each `*_minimize()`?) to papers which were used as basis or implemented directly.

PR #460, opened 2017-08-02
[WIP] Mutual Information Acquisition Function

yngtodd:
Adds a new acquisition function: Mutual Information. 

This conversation started in https://github.com/scikit-optimize/scikit-optimize/issues/456. I was interested in seeing how mutual information could speed up searches in complicated spaces. 

I first came across the idea of using mutual information following [[1]](http://www.jmlr.org/papers/v9/krause08a.html), but implementing Krause's approach seemed like it would not be straightforward. I then found there was a more recent approach brought to ICML in 2014 [[2]](http://proceedings.mlr.press/v32/contal14.pdf). This follows the second paper. 

To Do:
- [ ] Find a good setting for `alpha`

Open Questions:
* Is `gamma` being updated properly at each iteration?
* How to best handle kwargs for the acquisition function?

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -32,6 +32,8 @@ def _gaussian_acquisition(X, model, y_opt=None, acq_func="LCB",
         acq_func_kwargs = dict()
     xi = acq_func_kwargs.get("xi", 0.01)
     kappa = acq_func_kwargs.get("kappa", 1.96)
+    alpha = acq_func_kwargs.get("alpha", 0.5)
+    gamma = acq_func_kwargs.get("gamma", 0.0)
 
     # Evaluate acquisition function
     per_second = acq_func.endswith("ps")
@@ -40,6 +42,13 @@ def _gaussian_acquisition(X, model, y_opt=None, acq_func="LCB",
 
     if acq_func == "LCB":
         func_and_grad = gaussian_lcb(X, model, kappa, return_grad)
+        if return_grad:
+            acq_vals, acq_grad = func_and_grad
+        else:
+            acq_vals = func_and_grad
+
+    elif acq_func == "MI":
+        func_and_grad = gaussian_mi(X, model, y_opt, alpha, gamma, return_grad)
         if return_grad:
             acq_vals, acq_grad = func_and_grad
         else:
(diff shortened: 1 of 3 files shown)


yngtodd:
I have a short [notebook](https://github.com/yngtodd/skopt_notes/blob/master/acquisition_funcs/mutual_information_dev.ipynb) following the development of the acquisition function. 

It looks like using mutual information has some promise. Check out the convergence plot on the Branin function at the end of the notebook.

kiudee:
For the derivative at time t I derived the following:

Derivative
Let me know whether that works.

yngtodd:
You are awesome! 

Let me take a look at it. Do you think that would be all that is necessary when using gradients, or would we possibly need to use some sort of linear combination with the vanilla mutual information?

skopt/acquisition.py
```diff
@@ -296,3 +305,54 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+        mu, std, mu_grad, std_grad = model.predict(
+            X, return_std=True, return_mean_grad=True,
+            return_std_grad=True)
+    else:
+        mu, std = model.predict(X, return_std=True)
+
+    values = np.zeros_like(mu)
+    phi = np.sqrt(alpha) * (np.sqrt(std + gamma) - np.sqrt(gamma))
+    values = mu + phi
+    gamma = np.max(gamma + std)
```
kiudee: This is not correct. The algorithm first selects the x for which μ + ϕ is maximal and then adds the variance at that position.

skopt/acquisition.py
```diff
@@ -296,3 +305,54 @@ def gaussian_ei(X, model, y_opt=0.0, xi=0.01, return_grad=False):
...
+    """
+    if return_grad:
+        mu, std, mu_grad, std_grad = model.predict(
+            X, return_std=True, return_mean_grad=True,
+            return_std_grad=True)
+    else:
+        mu, std = model.predict(X, return_std=True)
+
+    values = np.zeros_like(mu)
+    phi = np.sqrt(alpha) * (np.sqrt(std + gamma) - np.sqrt(gamma))
```
kiudee: In the algorithm γ is the sum of variances (instead of the standard deviation σ they use σ²).

skopt/acquisition.py
yngtodd: So then would it be `values = np.max(mu) + phi`?

skopt/acquisition.py
yngtodd: Ah, right. Then in line 350, we would have 

`phi = np.sqrt(alpha) * (np.sqrt(std**2 + gamma) - np.sqrt(gamma))`

and in line 352:

`gamma = np.max(gamma + std**2)`

skopt/acquisition.py
kiudee: Now that I think about it: Currently the state of γ is not preserved and the argument is not used.
* Somewhere we need to store γ (either implement a class, or we need to do it in the optimizer)
* We have to be very careful, when to increment γ, because the function might be called several times with different points in order to reach the optimum. We can only increment it *after* one complete optimization run. This we need to implement in the optimizer.

Thoughts?

yngtodd: That was an open question of mine. 

Perhaps `gamma` could be returned from the function and set as a kwarg in the `Optimizer`?

kiudee: I agree, that is one of the cleanest solutions, since we have to update the value anyway in the optimizer.

kiudee:
I would also recommend citing the reference to the paper in the docstring (similiar to how it’s done in [PyMC3](https://github.com/pymc-devs/pymc3/blob/master/pymc3/distributions/multivariate.py#L882); there is also issue #118 in which it is requested).

skopt/acquisition.py
yngtodd: I am not quite sure where it would be best in the `Optimizer` to reset that `gamma`. Any ideas?

yngtodd: I am a bit stumped on this one. 

It seems like the most straight forward approach would be if the acquisition function was a method inside the `Optimizer`, and therefore could set a class variable `self.gamma = gamma` from within the method. However, this would either require pulling the acquisition functions into the `Optimizer` or creating a new class and using inheritance. Both would be a fairly major revisions of the code base. 

Not quite sure how to proceed with the kwarg approach. Any thoughts?

skopt/optimizer/base.py
```diff
@@ -22,7 +22,7 @@ def base_minimize(func, dimensions, base_estimator,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96, n_jobs=1):
+                  xi=0.01, kappa=1.96, alpha=1.6, gamma=0.0, n_jobs=1):
```
stefanocereda: The value of alpha shouldn't be checked to be greater than zero?
Just a quick comment: I think that the default value for alpha should be something like 6. In the paper they define alpha as log(2/delta) and then  perform all the experiments with delta=10^-6 (leading to an alpha of nearly 6).
Then they show that this is quite robust, and plot the regret for various value of delta (from 10^-1 to 10^-9), but I think that if they showed the results only for 10^-6 it should be the one with best results :)

yngtodd: gp_mi_calibrate_a

Thanks, I missed that they set <a href="https://www.codecogs.com/eqnedit.php?latex=\delta=10^{-6}" target="_blank"></a>. Like you said, the regret seems robust to `alpha`, but I wonder why they didn't set <a href="https://www.codecogs.com/eqnedit.php?latex=\delta=10^{-1&space;}" target="_blank"></a>. At least for the experiment shown in the Figure 4 above, it seems to be best. Am I overlooking their justification for that somewhere?

PR #465, opened 2017-08-03
[MRG+1] Optimizer() and gp_minimize should be equivalent

betatim:
Follow up for #402 

This PR adds a test that checks if indeed `Optimizer()` and `gp_minimize()` give the same results when used with default settings.

For that to be true I had to modify `gp_minimize` to use the normalised space. @MechCoder should take a look to double check.

Thoughts on the points only being within 1e-5 of each other instead of being identical? The random points are exactly identical but then it seems thigns diverge a bit once we start using the GP :-/

### DIFF
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -205,8 +205,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
     """
     # Check params
     rng = check_random_state(random_state)
-    transformed_dims = normalize_dimensions(dimensions)
-    space = Space(transformed_dims)
+    space = normalize_dimensions(dimensions)
     base_estimator = cook_estimator("GP", space=space, random_state=rng,
                                     noise=noise)
 
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -3,8 +3,6 @@
 from sklearn.utils import check_random_state
 
 from .base import base_minimize
-from ..space import check_dimension
-from ..space import Categorical
 from ..space import Space
 from ..utils import cook_estimator
 from ..utils import normalize_dimensions
@@ -213,7 +211,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
                                     noise=noise)
 
     return base_minimize(
-        func, dimensions, base_estimator=base_estimator,
+        func, space, base_estimator=base_estimator,
         acq_func=acq_func,
         xi=xi, kappa=kappa, acq_optimizer=acq_optimizer, n_calls=n_calls,
         n_points=n_points, n_random_starts=n_random_starts,
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -14,6 +14,7 @@ from sklearn.utils import check_random_state
 
 from ..acquisition import _gaussian_acquisition
 from ..acquisition import gaussian_acquisition_1D
+from ..learning import GaussianProcessRegressor
 from ..space import Categorical
 from ..space import Space
 from ..utils import check_x_in_space
@@ -172,7 +173,10 @@ class Optimizer(object):
                           DeprecationWarning)
             n_initial_points = n_random_starts
 
-        if isinstance(base_estimator, str) and base_estimator.upper() == "GP":
+        self._check_arguments(base_estimator, n_initial_points, acq_optimizer,
+                              dimensions)
+
+        if isinstance(self.base_estimator_, GaussianProcessRegressor):
             dimensions = normalize_dimensions(dimensions)
 
         self.space = Space(dimensions)
@@ -188,7 +192,6 @@ class Optimizer(object):
             else:
                 self._non_cat_inds.append(ind)
 
-        self._check_arguments(base_estimator, n_initial_points, acq_optimizer)
         self.n_jobs = n_jobs
 
         # The cache of responses of `ask` method for n_points not None.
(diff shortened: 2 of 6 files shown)


skopt/optimizer/gp.py
```diff
@@ -213,7 +211,7 @@ def gp_minimize(func, dimensions, base_estimator=None,
                                     noise=noise)
 
     return base_minimize(
-        func, dimensions, base_estimator=base_estimator,
+        func, space, base_estimator=base_estimator,
```
MechCoder: Nice catch, but with the recent changes by @yngtodd this reduces from a bug-fix to a micro-optimization

MechCoder:
I don't have any quick intuition on why both may diverge but one way is to check out the attributes of the models returned from both the Optimizer and the minimize functions?

skopt/tests/test_optimizer.py
```diff
@@ -194,3 +196,21 @@ def test_optimizer_base_estimator_string_smoke(base_estimator):
     opt = Optimizer([(-2.0, 2.0)], base_estimator=base_estimator,
                     n_initial_points=1, acq_func="EI")
     opt.run(func=lambda x: x[0]**2, n_iter=3)
+
+
+def test_defaults_are_equivalent():
```
betatim: I think this is a good test to add to get some testing of consistency. WDYT @MechCoder

skopt/optimizer/gp.py
betatim: Can you point me to the change?

betatim: I'm asking because the test I added in this PR fails without this change.

iaroslav-ai:
Hmm something is wrong with setting the random state - `test_minimizer_with_space` fails. Apparently after `n_random_starts` is used up, something does not has its random state set. Does `Space` has a random state itself maybe?

betatim:
I get this when running `test_minimizer_with_space`:
```
>       assert_array_almost_equal(space_result.x_iters, result.x_iters)
E       AssertionError: 
E       Arrays are not almost equal to 6 decimals
E       
E       (mismatch 50.0%)
E        x: array([[  1.25533 ,  10.804867],
E              [ -4.998284,   4.534989],
E              [  1.262632,  10.822849],
E              [  1.31651 ,   7.240632]])
E        y: array([[  1.25533 ,  10.804867],
E              [ -4.998284,   4.534989],
E              [  1.235615,  10.760034],
E              [  0.65846 ,   9.441846]])
```

so the deviation is after you sampled the random points. The first two points are from `Space` and then it is the model that picks the points. Maybe there is a shared random number generator somewhere :-/

betatim:
Only happens for `gp_minimize` for me.

skopt/optimizer/gp.py
```diff
@@ -207,13 +205,12 @@ def gp_minimize(func, dimensions, base_estimator=None,
     """
     # Check params
     rng = check_random_state(random_state)
-    transformed_dims = normalize_dimensions(dimensions)
-    space = Space(transformed_dims)
+    space = normalize_dimensions(dimensions)
```
betatim: `normalize_dimensions` now returns a `Space` instead of a list of dimensions.

skopt/optimizer/optimizer.py
```diff
@@ -172,7 +173,10 @@ def __init__(self, dimensions, base_estimator="gp",
                           DeprecationWarning)
             n_initial_points = n_random_starts
 
-        if isinstance(base_estimator, str) and base_estimator.upper() == "GP":
+        self._check_arguments(base_estimator, n_initial_points, acq_optimizer,
+                              dimensions)
+
+        if isinstance(self.base_estimator_, GaussianProcessRegressor):
```
betatim: switched the order of `_check_arguments` and this test so that we can handle `"GP"` and the user passing a GPRegressor instance

skopt/space/space.py
```diff
@@ -22,10 +22,13 @@ def __repr__(self):
...
 def check_dimension(dimension, transform=None):
-    """
+    """Turn a provided dimension description into a dimension object.
+
     Checks that the provided dimension falls into one of the
     supported types. For a list of supported types, look at
-    the documentation of `dimension` below.
+    the documentation of ``dimension`` below.
+
+    If ``dimension`` is already a ``Dimension`` instance, return it.
```
betatim: Don't think we should (ab)use `check_dimension` for modifying a `Dimension` so added this sentence.

skopt/utils.py
```diff
@@ -441,8 +442,7 @@ def point_aslist(search_space, point_as_dict):
 
 
 def normalize_dimensions(dimensions):
```
betatim: changes here are to help `normalize_dimensions` to also work when you pass a list of `Dimension` objects or a `Space`. Felt like this was unclear/complex before because `check_dimension` would simply return the value you pass unchanged if it was already a `Dimension` instance.

betatim:
IMHO the way to review this is: do you think that `test_defaults_are_equivalent` is a good test/behaviour. If yes most of the other changes follow from it and havinging `test_minimizer_with_space` in `test_common.py` pass.

skopt/tests/test_optimizer.py
iaroslav-ai: I think this is nice. Consider using a function that takes as input categorical, numerical and real values, in order to test consistency with a variety of dimension types, in order to possibly catch more bugs. For example:
```python
space = [('1', '2', '3'), (1, 3), (0.5, 1.5)]
def fnc(x):
    return float(x[0]) + float(x[1]) + float(x[2])
```

iaroslav-ai: Another suggestion: consider generating random seed at the beginning of the test, and then feeding  it as seed. This would ensure that for different seeds optimizers act same. More like a paranoia, but you never know.

skopt/utils.py
```diff
@@ -453,9 +453,9 @@ def normalize_dimensions(dimensions):
         List of search space dimensions.
         Each search dimension can be defined either as
 
-        - a `(upper_bound, lower_bound)` tuple (for `Real` or `Integer`
```
iaroslav-ai: yeah lol :D

iaroslav-ai:
Other than suggestions that I have for `test_defaults_are_equivalent` I do not have any other issues. LGTM

skopt/tests/test_optimizer.py
betatim: I'd delay this to an issue: the code for checking that the points sampled are the same gets a bit ugly if you have strings (I know lazy Tim) and we should think about how to run this test with less than 15 iterations when we add different seeds as starting point.

Issue #461, opened 2017-08-02
Test failures on master


amueller:
```
surrogate = 'gp', n_jobs = 1

    @pytest.mark.parametrize("surrogate", ['gp', None])
    @pytest.mark.parametrize("n_jobs", [1, -1])  # test sequential and parallel
    def test_searchcv_runs(surrogate, n_jobs):
        """
        Test whether the cross validation search wrapper around sklearn
        models runs properly with available surrogates and with single
        or multiple workers.
    
        Parameters
        ----------
    
        * `surrogate` [str or None]:
            A class of the scikit-optimize surrogate used. None means
            to use default surrogate.
    
        * `n_jobs` [int]:
            Number of parallel processes to use for computations.
    
        """
    
        X, y = load_iris(True)
        X_train, X_test, y_train, y_test = train_test_split(
            X, y, train_size=0.75, random_state=0
        )
    
        # None search space is only supported when only `step` function is used
        assert_raises(ValueError, BayesSearchCV(SVC(), None).fit, (X, y))
    
        # check if invalid dimensions are raising errors
        with pytest.raises(ValueError):
            BayesSearchCV(SVC(), {'C': '1 ... 100.0'})
    
        with pytest.raises(TypeError):
            BayesSearchCV(SVC(), ['C', (1.0, 1)])
    
        # create an instance of a surrogate if it is not a string
        if surrogate is not None:
            optimizer_kwargs = {'base_estimator': surrogate}
        else:
            optimizer_kwargs = None
    
        opt = BayesSearchCV(
            SVC(),
            {
                'C': Real(1e-6, 1e+6, prior='log-uniform'),
                'gamma': Real(1e-6, 1e+1, prior='log-uniform'),
                'degree': Integer(1, 8),
                'kernel': Categorical(['linear', 'poly', 'rbf']),
            },
            n_jobs=n_jobs, n_iter=11,
            optimizer_kwargs=optimizer_kwargs
        )
    
>       opt.fit(X_train, y_train)

skopt/tests/test_searchcv.py:70: 
[...]

PR #473, opened 2017-08-08
Fix test_utils.py with scikit-learn master

lesteve:
Need to seek at the beginning of the file before loading the pickle
from the file.

This is probably a change in the joblib we bundle with scikit-learn between scikit-learn 0.18 and scikit-learn master.

Note that I think this is fine and has the same behaviour as pickle:

```py
import tempfile
import pickle

with tempfile.TemporaryFile() as f:
    pickle.dump(object(), f)
    pickle.load(f)
```

```
---------------------------------------------------------------------------
EOFError                                  Traceback (most recent call last)
<ipython-input-3-5857a6e21a79> in <module>()
      4 with tempfile.TemporaryFile() as f:
      5     pickle.dump(object(), f)
----> 6     pickle.load(f)
      7 

EOFError: Ran out of input
```

Fixes part of #461.

### DIFF
--- a/skopt/tests/test_utils.py
+++ b/skopt/tests/test_utils.py
@@ -44,6 +44,7 @@ def test_dump_and_load():
     # Test normal dumping and loading
     with tempfile.TemporaryFile() as f:
         dump(res, f)
+        f.seek(0)
         res_loaded = load(f)
     check_optimization_results_equality(res, res_loaded)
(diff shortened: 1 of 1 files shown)


ogrisel:
LGTM.

glouppe:
Thanks a lot for the fix!

Issue #476, opened 2017-08-12
BayesSearchCV only searches through first parameter space


savvastj:
I tried running a couple of toy examples, including the advanced example from the documentation, and `BayesSearchCV` seems to only search over the first parameter space in the list of tuples that I pass to `search_spaces`. 

[Here](https://github.com/savvastj/ml_playground/blob/master/try_skopt_bayescv.ipynb) is a notebook where I try out `RandomForestRegressor`, `Ridge`, and `Lasso`, but `cv_results` only returns information regarding the Random Forest's parameter space.  (Another strange behavior is that even though I set 50 evaluations for Random Forest, 56 were returned.)

And [here](https://github.com/savvastj/ml_playground/blob/master/adv_bayescv_skopt_example.ipynb) is a notebook containing the advanced example from the documentation.  The `cv_results` only contain evaluations for `LinearSVC`, but nothing for the `DecisionTreeClassifier` or `SVC`.

betatim:
Thanks for filing the issue and all the details!

Do you also want to create a PR to fix this?

betatim:
I think the bug is [here](https://github.com/scikit-optimize/scikit-optimize/blob/39966fd750c62e236e2a629f26ad5e7f6a548452/skopt/searchcv.py#L390) where we check if the names are a list or not. By default a `range()` is passed in which doesn't pass the check (ranges aren't lists :( )

iaroslav-ai:
Hmm I think this can be solved [here](https://github.com/scikit-optimize/scikit-optimize/blob/39966fd750c62e236e2a629f26ad5e7f6a548452/skopt/searchcv.py#L280) 

I think this needs to be list(range ...)

Some time ago I looked at it, but then forgot about it :(

iaroslav-ai:
fixed with #478

PR #478, opened 2017-08-12
[MRG] Fix the space enumeration of the BayesSearch, add corresponding test

iaroslav-ai:
#476

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -282,7 +282,7 @@ class BayesSearchCV(sklearn.model_selection._search.BaseSearchCV):
             # account for the case when search space is a dict
             if isinstance(search_spaces, dict):
                 search_spaces = [search_spaces]
-            self.add_spaces(range(len(search_spaces)), search_spaces)
+            self.add_spaces(list(range(len(search_spaces))), search_spaces)
 
         self.n_iter = n_iter
         self.random_state = random_state
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -282,7 +282,7 @@ class BayesSearchCV(sklearn.model_selection._search.BaseSearchCV):
             # account for the case when search space is a dict
             if isinstance(search_spaces, dict):
                 search_spaces = [search_spaces]
-            self.add_spaces(range(len(search_spaces)), search_spaces)
+            self.add_spaces(list(range(len(search_spaces))), search_spaces)
 
         self.n_iter = n_iter
         self.random_state = random_state
@@ -393,7 +393,7 @@ class BayesSearchCV(sklearn.model_selection._search.BaseSearchCV):
 
         if not isinstance(search_spaces, list):
             search_spaces = [search_spaces]
-        if isinstance(names, Iterable):
+        if isinstance(names, list):
             names = list(names) # to avoid emptying generator
         else:
             names = [names]
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -1,4 +1,4 @@
-from collections import defaultdict, Sized
+from collections import defaultdict, Sized, Iterable
 from functools import partial, reduce
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


iaroslav-ai:
If no issues arise any time soon, I will merge it.

betatim:
Travis is unhappy :-/

betatim:
Should we change the if statement to check for `Iterable` instead of list? At least `list` and `tuple` should be handled the same.

skopt/tests/test_searchcv.py
```diff
@@ -85,19 +88,42 @@ def test_searchcv_runs_multiple_subspaces():
...
-                'C': Real(1e-6, 1e+6, prior='log-uniform'),
-                'gamma': Real(1e-6, 1e+1, prior='log-uniform'),
-            }, 10),
-            {
-                'degree': Integer(1, 8),
-                'kernel': Categorical(['linear', 'poly', 'rbf']),
-            }
-        ],
+        pipe,
+        [(lin_search, 11), (dtc_search, 11), svc_search],
```
betatim: can we reduce this to being one evaluation per subspace? Save on runtime.

skopt/tests/test_searchcv.py
```diff
@@ -85,19 +88,42 @@ def test_searchcv_runs_multiple_subspaces():
...
         n_iter=10
     )
 
     opt.fit(X_train, y_train)
+
+    # test if all subspaces are explored
+    total_evaluations = len(opt.cv_results_['mean_test_score'])
+    assert_equal(
```
betatim: don't need `assert_equal` anymore with pytest. `assert x == y, 'my message here'` should work.

iaroslav-ai:
Yup, new sklearn got released 23 hours ago :) Maybe lets resolve #479 first, and then I rebase onto master.

skopt/tests/test_searchcv.py
iaroslav-ai: yup nice

skopt/tests/test_searchcv.py
iaroslav-ai: yes

iaroslav-ai:
Used `Iterable` as you suggest, more general

betatim:
> Used `Iterable` as you suggest, more general

still need to deal with `range()` though as I think we use `names` twice which means the second time `range()` will be empty (because it is a generator).

iaroslav-ai:
Hmm seems to pass the tests fine anyhows 
But then maybe I just add conversion to list?

skopt/searchcv.py
```diff
@@ -393,7 +393,9 @@ def add_spaces(self, names, search_spaces):
 
         if not isinstance(search_spaces, list):
             search_spaces = [search_spaces]
-        if not isinstance(names, list):
+        if isinstance(names, list):
```
betatim: not sure this is what we really want, "if it is a list, then make a list"?

The more I think about this, can't we make our life easier by requiring that this method has only one argument `spaces` which has to be a list of `({'param1': ..., 'param2': ..}, 2)` or `{'param1': ..., 'param2': ..}` items. Internal to the method we then make up names.

Can you remind me why we need to name the spaces at all? Couldn't we switch `self.search_spaces_` to a list instead? We can check if a search space is already in the list with `space in self.search_spaces_`.

To have named spaces (and dimensions) I would go the route of adding it to `skopt.Space` etc and then taking it from there. Most `BayesSearchCV` users probably come from `sklearn.GridSearchCV` where they can't name their spaces either.

iaroslav-ai: I thought originally of a use case where you can resume the calculations if necessary, or switch between calculations using `step` function, which does a single step in particular subspace which is specified as an argument to the function. This way one can have some custom stopping criterions.

Thus far I am also more leaning to reduce the complexity of the class - we can revert it when necessary.

I would suggest to do the following:

1. Merge this as is and release 0.4 soon
2. Reduce the class after the 0.4 release

There are many issues right now that people have simply because pip version of package is outdated, so I think that would be nice to fix soon.

betatim: Ok, let's do this.

betatim:
Will merge once circleci is happy

PR #483, opened 2017-08-17
Bump version to 0.4

betatim:
Preparation for releasing v0.4. Once this get's merge I'll tag the new version and we should have a new release.

What about #465? I think the test I added should pass? Should we wait for that PR?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,6 +35,8 @@ Important links
 Install
 -------
 
+The latest released version of scikit-optimize is v0.4, which you can install
+with:
 ::
 
     pip install numpy # install numpy explicitly first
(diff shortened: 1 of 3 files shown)


betatim:
What did I forget to do for the release?

iaroslav-ai:
#465 would be awesome to add if you think it can be added soon - but I think not critical.

betatim:
If this builds I will tag 0.4rc1, see if the "push directly to PyPI" setup works, if yes then tag 0.4 straight away unless we find some weird last minute bugs.

iaroslav-ai:
Awesome! :D

PR #485, opened 2017-08-18
Switch from `ax` to `ax_`. Fix Numpy/matplotlib bug.

Deepblue129:
`ax` is `numpy.ndarray` while `ax_` is matplotlib 

**Got this error using the library:**
  File "bin/tune.py", line 209, in <module>
    main(base_config, tune_config)
  File "/root/pytorch-seq2seq/seq2seq/config/configurable.py", line 250, in configurable
    return func(*args, **merged)
  File "bin/tune.py", line 199, in main
    plot_result(optimizer, result)
  File "bin/tune.py", line 89, in plot_result
    ax.figure.savefig(path)
AttributeError: 'numpy.ndarray' object has no attribute 'figure'

Was running `plot_evaluations` and `plot_objective`.


Deepblue129:
Never mind!

Issue #490, opened 2017-08-29
Plot Objective Figure Size (PR)


arose13:
Could I make a PR where the user can easily change the figure size of this function? Maybe with something like plot_objective(cell_size=int) since depending on the matplotlib theme used 2 as a figure size multiplier is extremely small.

Thank you

betatim:
This would be a great PR. Not sure what the best way to implement this is though. Should we stop doing any explicit figure size settings in the function and ask the user to pass in a figure/axis if they want a different size?

However having a rectangular plot is probably not what anyone wants for a figure with a matrix of plots on it.

Do you have an idea how other libraries deal with: use a non default figure size and let the user override it?

arose13:
Yeah absolutely. I was going to implement it the same way [Seaborn](https://github.com/mwaskom/seaborn) allows changes in the size of their [pair plots](http://seaborn.pydata.org/generated/seaborn.pairplot.html) by using the arg `size`. 

Since these facet style plots are complicated to deal with in the first place on the user end I thought this may actually be the most elegant solution.

betatim:
Sounds like a good idea to me.

arose13:
I've submitted the pull request. Tell me if there are any issues

glouppe:
Thanks for your contribution :)

PR #491, opened 2017-08-30
Size argument added to control the size of facets in objective plotting

arose13:
This was for issue [490](https://github.com/scikit-optimize/scikit-optimize/issues/490) about figure sizes


glouppe:
Thanks!

Issue #240, opened 2016-09-29
Can't pass training data or estimator into objective function (need to make global)


ratrock:
Hi, maybe I'm doing this wrong, but I was trying to implement a version of what's in the hyperparameter optimization example, but not in an interactive notebook.  If I set up the training data and search space in main() or another function, the only way for the objective function that I pass to the minimizer to access those variables is if I make them global.  Is that right?

Would it make sense for there be another way to pass additional variables/kwargs to the objective function in the minimizer call?  But then I guess the minimizer might then have to return those objects, too, so maybe not.  

Is there any way to avoid globals like I did below?

~~~
def objective(params):
    max_depth, learning_rate, max_features, min_samples_split, min_samples_leaf = params

    reg.set_params(max_depth=max_depth,
                   learning_rate=learning_rate,
                   max_features=max_features,
                   min_samples_split=min_samples_split, 
                   min_samples_leaf=min_samples_leaf)

    return -np.mean(cross_val_score(reg, X_train, y_train, cv=5, n_jobs=-1,
                                    scoring="neg_mean_absolute_error"))
~~~
~~~
def optimize_ml(X, y):
    global X_train
    global y_train
    global reg

    X_train, y_train, X_test, y_test = train_test_split(X, y, test_size=.25)
    reg = GradientBoostingRegressor(n_estimators=50, random_state=0)

    space  = [(1, 5),                           # max_depth
          (10**-5, 10**-1, "log-uniform"),  # learning_rate
          (1, X.shape[1]),                  # max_features
          (2, 30),                          # min_samples_split
          (1, 30)]                          # min_samples_leaf

    x0 = [3, 0.01, 6, 2, 1]
    res_gp = gp_minimize(objective, space, x0=x0, n_calls=50, random_state=0)
   ~~~

MechCoder:
Not currently, but following SciPy's API we can most certainly allow optional arguments to be passed into the minimizer for it to be passed into the objective function.

For example, args=() in http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_l_bfgs_b.html#scipy.optimize.fmin_l_bfgs_b

betatim:
👍 for `args=()`.

In the mean time you might be able work around this with something like:
```
from functools import partial

def objective(X_train, y_train, args):
    ...
def optimize_ml(X, y):
    X_train, y_train, X_test, y_test = train_test_split(X, y, test_size=.25)
    ...
[...]

PR #500, opened 2017-09-01
[WIP] Add support for the `args=()` argument to the *_minimize API

Anaaa:
Closes https://github.com/scikit-optimize/scikit-optimize/issues/240

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -9,8 +9,6 @@ import inspect
 import numbers
 from collections import Iterable
 
-import numpy as np
-
 from ..callbacks import check_callback
 from ..callbacks import VerboseCallback
 from .optimizer import Optimizer
@@ -22,7 +20,7 @@ def base_minimize(func, dimensions, base_estimator,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96, n_jobs=1, args=()):
+                  xi=0.01, kappa=1.96, n_jobs=1, args=None):
     """
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


skopt/optimizer/base.py
```diff
@@ -170,6 +172,11 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
+    if args:
+        opt_func = lambda x_param: func(x_param, args)
```
betatim: How about using [`partial` from `functools`](https://docs.python.org/3/library/functools.html#functools.partial) instead of our own `lambda` based thing?

Should this be `func(x_param, *args)` instead of `func(x_param, args)`? Also maybe we can just do `func = ...` instead of using `opt_func`?

skopt/optimizer/base.py
```diff
@@ -170,6 +172,11 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
+    if args:
```
betatim: Should test `if args is not None`

betatim:
Do you want to add a test that checks the `args` argument is correctly passed to the objective `func` for each kind of `*_minimize` function? After that this should be ready to be merged I think.

skopt/optimizer/base.py
```diff
@@ -22,7 +21,7 @@ def base_minimize(func, dimensions, base_estimator,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
                   callback=None, n_points=10000, n_restarts_optimizer=5,
-                  xi=0.01, kappa=1.96, n_jobs=1):
+                  xi=0.01, kappa=1.96, n_jobs=1, args=None):
```
MechCoder: Even though the scipy optimize API uses args, it might be better to use `func_kwargs` as a dict  to conveniently pass them into the black-box function such that the order does not matter and API consistency when we refactor into `optimizer_kwargs`, `acq_func_kwargs` and such.

betatim: So leave `gp_minimize` and co using `args=()` but switch `base_minimize` to pass it on as a dictionary?

We had discussed keyword args or not at the sprint and decided that following the `scipy.optmize` API would be better. I don't have a strong opinion beyond keeping the API as consistent as possible.

MechCoder: sounds ok. @Anaaa are you still there? Can you please add a test?

PR #501, opened 2017-09-01
Check point dimension on tell

### LINKED ISSUE #487: Optimizer.tell doesn't check length of point
First, let me say how awesome this library is and thanks for all the hard work that has been put into it.

I was playing around recently and made a typo that really tripped me up. The setup here is that there is some amount of validation that happens to points passed in to `Optimizer.tell`:

```python
from skopt import Optimizer
opt = Optimizer([('a', 'b', 'c'), (1, 10)])
opt.tell(['d', 2], 1.0)  # expect ValueError since 'd' not in space
[output trimmed...]
ValueError: Point (['d', 2]) is not within the bounds of the space ([('a', 'b', 'c'), (1, 10)]).
```
That behaves just like I would expect. But if I accidentally omit one of the dimensions of the point, then this happens:

```python
>>> opt.tell(['a', ], 1.0)  # I expect an error here too, but it works fine...
          fun: 1.0
    func_vals: array([ 1.,  1.])
       models: []
 random_state: <mtrand.RandomState object at 0x0101010102>
        space: Space([Categorical(categories=('a', 'b', 'c'), prior=None),
       Integer(low=1, high=10)])
        specs: None
            x: ['a', 2]
      x_iters: [['a', 2], ['a']]
>>> # I can still add more points
>>> for i in range(7):
...     opt.tell(opt.ask(), 1.0)
... 
[output trimmed...]
>>> # no errors until I make it to `opt._n_initial_points` and a model is built
>>> opt.tell(opt.ask(), 1.0)
[output trimmed...]
IndexError: list index out of range
```

It looks like this happens because of using `zip` in `Space.__contains__`, which is used by `utils.check_x_in_space` which is called at the start of `Optimizer.tell`. Since `zip` will work even if the inputs don't have the same length, maybe it should be  preceded by checking that `len(point) == len(self.dimensions)`?

Thanks!

mikesmyk:
Fixes #487 
Added check for dimension of point and space.
Check is added in utils.check_x_in_space(), could also be added in __contains__ function of space, but then Exception message in check_x_in_space should be adapted.

### DIFF
--- a/skopt/tests/test_optimizer.py
+++ b/skopt/tests/test_optimizer.py
@@ -101,6 +101,47 @@ def test_bounds_checking_2D_multiple_points():
 
 
 @pytest.mark.fast_test
+def test_dimension_checking_1D():
+    low = -2
+    high = 2
+    opt = Optimizer([(low, high)])
+    with pytest.raises(ValueError) as e:
+        # within bounds but one dimension too high
+        opt.tell([low+1, low+1], 2.)
+    assert "Dimensions of point " in str(e.value)
+
+
+@pytest.mark.fast_test
+def test_dimension_checking_2D():
+    low = -2
+    high = 2
+    opt = Optimizer([(low, high), (low, high)])
+    # within bounds but one dimension too little
+    with pytest.raises(ValueError) as e:
+        opt.tell([low+1, ], 2.)
+    assert "Dimensions of point " in str(e.value)
+    # wthin bounds but one dimension to much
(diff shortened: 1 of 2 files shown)


glouppe:
Looks good to me. Thanks! What do others think?

betatim:
Thanks for this @mikesmyk! Congratulations on your first contribution to scikit-optimize *and* your first ever PR on GitHub! 🎆🍻

PR #502, opened 2017-09-02
[MRG+1] Fix random seed of regressors

### LINKED ISSUE #499: Assertion Error while testing test_optimizer.py
Hi, 

when i ran the Test's i got an Assertion Error. The Error is at the end of this text.  I have the following python version
Python 3.6.1 :: Anaconda 4.4.0 (32-bit).

and i've installed the following packages:

alabaster==0.7.10
anaconda-client==1.6.3
anaconda-navigator==1.6.2
anaconda-project==0.6.0
asn1crypto==0.22.0
astroid==1.4.9
astropy==1.3.2
Babel==2.4.0
backports.shutil-get-terminal-size==1.0.0
beautifulsoup4==4.6.0
bitarray==0.8.1
blaze==0.10.1
bleach==1.5.0
bokeh==0.12.5
boto==2.46.1
Bottleneck==1.2.1
cffi==1.10.0
chardet==3.0.3
click==6.7
cloudpickle==0.2.2
clyent==1.2.2
colorama==0.3.9
conda==4.3.25
contextlib2==0.5.5
cryptography==1.8.1
cycler==0.10.0
Cython==0.25.2
cytoolz==0.8.2
dask==0.14.3
datashape==0.5.4
decorator==4.0.11
distributed==1.16.3
docutils==0.13.1
entrypoints==0.2.2
et-xmlfile==1.0.1
fastcache==1.0.2
Flask==0.12.2
Flask-Cors==3.0.2
gevent==1.2.1
greenlet==0.4.12
h5py==2.7.0
HeapDict==1.0.0
html5lib==0.999
idna==2.5
imagesize==0.7.1
ipykernel==4.6.1
ipython==5.3.0
ipython-genutils==0.2.0
ipywidgets==6.0.0
isort==4.2.5
itsdangerous==0.24
jdcal==1.3
jedi==0.10.2
Jinja2==2.9.6
jsonschema==2.6.0
jupyter==1.0.0
jupyter-client==5.0.1
jupyter-console==5.1.0
jupyter-core==4.3.0
lazy-object-proxy==1.2.2
llvmlite==0.18.0
locket==0.2.0
lxml==3.7.3
MarkupSafe==0.23
matplotlib==2.0.2
mistune==0.7.4
mpmath==0.19
msgpack-python==0.4.8
multipledispatch==0.4.9
navigator-updater==0.1.0
nbconvert==5.1.1
nbformat==4.3.0
networkx==1.11
nltk==3.2.3
nose==1.3.7
notebook==5.0.0
numba==0.33.0
numexpr==2.6.2
numpy==1.12.1
numpydoc==0.6.0
odo==0.5.0
olefile==0.44
openpyxl==2.4.7
packaging==16.8
pandas==0.20.1
pandocfilters==1.4.1
partd==0.3.8
pathlib2==2.2.1
patsy==0.4.1
pep8==1.7.0
pexpect==4.2.1
pickleshare==0.7.4
Pillow==4.1.1
ply==3.10
prompt-toolkit==1.0.14
psutil==5.2.2
ptyprocess==0.5.1
py==1.4.33
pycosat==0.6.2
pycparser==2.17
pycrypto==2.6.1
pycurl==7.43.0
pyflakes==1.5.0
Pygments==2.2.0
pylint==1.6.4
pyodbc==4.0.16
pyOpenSSL==17.0.0
pyparsing==2.1.4
pytest==3.0.7
python-dateutil==2.6.0
pytz==2017.2
PyWavelets==0.5.2
PyYAML==3.12
pyzmq==16.0.2
QtAwesome==0.4.4
qtconsole==4.3.0
QtPy==1.2.1
requests==2.14.2
rope-py3k==0.9.4.post1
scikit-garden==0.1.3
scikit-image==0.13.0
scikit-learn==0.18.1
-e git+git@github.com:BlackTeaAndCoffee/scikit-optimize.git@0d52aca9ad354a1f9cad942d4a41730edffbb271#egg=scikit_optimize
scipy==0.19.0
seaborn==0.7.1
simplegeneric==0.8.1
singledispatch==3.4.0.3
six==1.10.0
snowballstemmer==1.2.1
sortedcollections==0.5.3
sortedcontainers==1.5.7
Sphinx==1.5.6
spyder==3.1.4
SQLAlchemy==1.1.9
statsmodels==0.8.0
sympy==1.0
tables==3.3.0
tblib==1.3.2
terminado==0.6
testpath==0.3
toolz==0.8.2
tornado==4.5.1
traitlets==4.3.2
unicodecsv==0.14.1
virtualenv==15.1.0
wcwidth==0.1.7
Werkzeug==0.12.2
widgetsnbextension==2.0.0
wrapt==1.10.10
xlrd==1.0.0
XlsxWriter==0.9.6
xlwt==1.2.0
zict==0.1.2

```
________________________________________________________________ test_defaults_are_equivalent _________________________________________________________________________________

    def test_defaults_are_equivalent():
        # check that the defaults of Optimizer reproduce the defaults of
        # gp_minimize
        space = [(-5., 10.), (0., 15.)]
        opt = Optimizer(space, random_state=1)
    
        for n in range(15):
            x = opt.ask()
            res_opt = opt.tell(x, branin(x))
    
        res_min = gp_minimize(branin, space, n_calls=15, random_state=1)
    
        assert res_min.space == res_opt.space
        # tolerate small differences in the points sampled
>       assert np.allclose(res_min.x_iters, res_opt.x_iters, atol=1e-5)
E       assert False
E        +  where False = <function allclose at 0xb64c8ecc>([[1.2553300705386121, 10.804867401632373], [-4.9982843777398269, 4.5349885894775976], [-2.798661637743304, 1.385078921...3349359, 5.1834109056457169], [0.95151211346005038, 8.0822510100503564], [1.2879177160494235, 10.278292505951395], ...], [[1.2553300705386121, 10.804867401632373], [-4.9982843777398269, 4.5349885894775976], [-2.798661637743304, 1.385078921...3349359, 5.1834109056457169], [0.95151211346005038, 8.0822510100503564], [1.2879177160494235, 10.278292505951395], ...], atol=1e-05)
E        +    where <function allclose at 0xb64c8ecc> = np.allclose
E        +    and   [[1.2553300705386121, 10.804867401632373], [-4.9982843777398269, 4.5349885894775976], [-2.798661637743304, 1.385078921...3349359, 5.1834109056457169], [0.95151211346005038, 8.0822510100503564], [1.2879177160494235, 10.278292505951395], ...] =           fun: 0.50341434699834409\n    func_vals: array([  56.52874005,  172.66531087,  102.58290349,   28.8680646 ,\n ...624], [-1.7196703721825872, 9.277871209886758], [-3.2884466845220692, 12.677325208182735], [-4.1841820464590658, 15.0]].x_iters
E        +    and   [[1.2553300705386121, 10.804867401632373], [-4.9982843777398269, 4.5349885894775976], [-2.798661637743304, 1.385078921...3349359, 5.1834109056457169], [0.95151211346005038, 8.0822510100503564], [1.2879177160494235, 10.278292505951395], ...] =           fun: 0.50341439890679318\n    func_vals: array([  56.52874005,  172.66531087,  102.58290349,   28.8680646 ,\n ...624], [-1.719670454526022, 9.2778712199133473], [-3.2884467112417353, 12.677325427507068], [-4.1840688238414359, 15.0]].x_iters

skopt/tests/test_optimizer.py:215: AssertionError
```

betatim:
Fixes #499 

Pass an integer derived from the random state to our regressors instead
of sharing the random state with them.

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -245,10 +245,8 @@ def base_minimize(func, dimensions, base_estimator,
     for n in range(n_calls):
         next_x = optimizer.ask()
 
-        # no need to fit a model on the last iteration
-        fit_model = n < n_calls - 1
         next_y = func(next_x)
-        result = optimizer.tell(next_x, next_y, fit=fit_model)
+        result = optimizer.tell(next_x, next_y)
         result.specs = specs
 
         if eval_callbacks(callbacks, result):
--- a/skopt/optimizer/gp.py
+++ b/skopt/optimizer/gp.py
@@ -1,9 +1,10 @@
 """Gaussian process-based minimization algorithms."""
 
+import numpy as np
+
 from sklearn.utils import check_random_state
 
 from .base import base_minimize
-from ..space import Space
 from ..utils import cook_estimator
 from ..utils import normalize_dimensions
 
@@ -206,8 +207,9 @@ def gp_minimize(func, dimensions, base_estimator=None,
     # Check params
     rng = check_random_state(random_state)
     space = normalize_dimensions(dimensions)
-    base_estimator = cook_estimator("GP", space=space, random_state=rng,
-                                    noise=noise)
+    base_estimator = cook_estimator(
+        "GP", space=space, random_state=rng.randint(0, np.iinfo(np.int32).max),
+        noise=noise)
 
     return base_minimize(
         func, space, base_estimator=base_estimator,
@@ -215,5 +217,5 @@ def gp_minimize(func, dimensions, base_estimator=None,
         xi=xi, kappa=kappa, acq_optimizer=acq_optimizer, n_calls=n_calls,
         n_points=n_points, n_random_starts=n_random_starts,
         n_restarts_optimizer=n_restarts_optimizer,
-        x0=x0, y0=y0, random_state=random_state, verbose=verbose,
+        x0=x0, y0=y0, random_state=rng, verbose=verbose,
         callback=callback, n_jobs=n_jobs)
--- a/skopt/space/transformers.py
+++ b/skopt/space/transformers.py
@@ -108,15 +108,15 @@ class Normalize(Transformer):
         to `True` when applying this transform to integers.
     """
     def __init__(self, low, high, is_int=False):
-        self.low = low
-        self.high = high
+        self.low = float(low)
+        self.high = float(high)
         self.is_int = is_int
 
     def transform(self, X):
         X = np.asarray(X)
-        if np.any(X > self.high):
+        if np.any(X > self.high + 1e-8):
             raise ValueError("All values should be less than %f" % self.high)
-        if np.any(X < self.low):
+        if np.any(X < self.low - 1e-8):
             raise ValueError("All values should be greater than %f" % self.low)
         return (X - self.low) / (self.high - self.low)
 
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -121,7 +121,7 @@ def test_minimizer_api_dummy_minimize(verbose, call):
 def test_minimizer_api(verbose, call, minimizer):
     n_calls = 7
     n_random_starts = 3
-    n_models = n_calls - n_random_starts
+    n_models = n_calls - n_random_starts + 1
 
     result = minimizer(branin, [(-5.0, 10.0), (0.0, 15.0)],
                        n_random_starts=n_random_starts,
@@ -213,7 +213,7 @@ def test_init_vals_and_models(n_random_starts, optimizer_func):
     res = optimizer(branin, space, x0=x0, y0=y0, random_state=0,
                     n_calls=n_calls)
 
-    assert_equal(len(res.models), n_calls - n_random_starts)
+    assert_equal(len(res.models), n_calls - n_random_starts + 1)
 
 
 @pytest.mark.slow_test
@@ -232,7 +232,7 @@ def test_init_points_and_models(n_random_starts, optimizer_func):
     optimizer = partial(optimizer_func, n_random_starts=n_random_starts)
     res = optimizer(branin, space, x0=x0, random_state=0,
                     n_calls=n_calls)
-    assert_equal(len(res.models), n_calls - len(x0) - n_random_starts)
+    assert_equal(len(res.models), n_calls - len(x0) - n_random_starts + 1)
 
 
 @pytest.mark.slow_test
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -422,7 +422,7 @@ def test_early_stopping_delta_x(minimizer):
     res = minimizer(bench1,
                     callback=DeltaXStopper(0.1),
                     dimensions=[(-1., 1.)],
-                    x0=[[-0.1], [0.9], [0.1]],
+                    x0=[[-0.1], [0.1], [0.9]],
                     n_calls=n_calls,
                     n_random_starts=0, random_state=1)
     assert len(res.x_iters) < n_calls
--- a/skopt/tests/test_common.py
+++ b/skopt/tests/test_common.py
@@ -418,11 +418,11 @@ def test_consistent_x_iter_dimensions(minimizer):
 @pytest.mark.parametrize("minimizer",
                          [gp_minimize, forest_minimize, gbrt_minimize])
 def test_early_stopping_delta_x(minimizer):
-    n_calls = 10
+    n_calls = 11
     res = minimizer(bench1,
                     callback=DeltaXStopper(0.1),
                     dimensions=[(-1., 1.)],
-                    x0=[[-0.1], [0.1], [0.9]],
+                    x0=[[-0.1], [0.1], [-0.9]],
                     n_calls=n_calls,
                     n_random_starts=0, random_state=1)
     assert len(res.x_iters) < n_calls
--- a/skopt/tests/test_forest_opt.py
+++ b/skopt/tests/test_forest_opt.py
@@ -46,7 +46,7 @@ def check_minimize(minimizer, func, y_opt, dimensions, margin,
 @pytest.mark.slow_test
 @pytest.mark.parametrize("name, minimizer", MINIMIZERS)
 def test_tree_based_minimize(name, minimizer):
-    check_minimize(minimizer, bench1, 0.,
+    check_minimize(minimizer, bench1, 0.05,
                    [(-2.0, 2.0)], 0.05, 25, 5)
 
     # XXX: We supply points at the edge of the search
@@ -56,9 +56,9 @@ def test_tree_based_minimize(name, minimizer):
     # the minimum, since for a decision tree any point greater than
     # max(sampled_points) would give a constant value.
     X0 = [[-5.6], [-5.8], [5.8], [5.6]]
-    check_minimize(minimizer, bench2, -5,
+    check_minimize(minimizer, bench2, -4.7,
                    [(-6.0, 6.0)], 0.1, 20, 10, X0)
-    check_minimize(minimizer, bench3, -0.9,
+    check_minimize(minimizer, bench3, -0.4,
                    [(-2.0, 2.0)], 0.05, 10, 5)
-    check_minimize(minimizer, bench4, 0.0,
+    check_minimize(minimizer, bench4, 1.,
                    [("-2", "-1", "0", "1", "2")], 0.05, 5, 1)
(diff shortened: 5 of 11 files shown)


skopt/optimizer/base.py
```diff
@@ -245,10 +245,8 @@ def base_minimize(func, dimensions, base_estimator,
     for n in range(n_calls):
         next_x = optimizer.ask()
 
-        # no need to fit a model on the last iteration
-        fit_model = n < n_calls - 1
         next_y = func(next_x)
-        result = optimizer.tell(next_x, next_y, fit=fit_model)
+        result = optimizer.tell(next_x, next_y)
```
betatim: What was the thinking (other than performance) for not fitting a final model? Changed this because `expected_minimum()` and other users would expect that `res.models[-1]` is a model that includes all points. Without this fix you get the following:

image

One of the green points isn't used for the model.

skopt/space/transformers.py
```diff
@@ -108,15 +108,15 @@ class Normalize(Transformer):
...
-        self.low = low
-        self.high = high
+        self.low = float(low)
+        self.high = float(high)
         self.is_int = is_int
 
     def transform(self, X):
         X = np.asarray(X)
-        if np.any(X > self.high):
+        if np.any(X > self.high + 1e-8):
```
betatim: Unsure if this is really the best way of doing this. Sometimes `scipy.optimize` passes in a value that is just ever so slightly larger than the boundary, so this stops the exception from being raised.

skopt/tests/test_common.py
```diff
@@ -418,11 +418,11 @@ def test_consistent_x_iter_dimensions(minimizer):
 @pytest.mark.parametrize("minimizer",
                          [gp_minimize, forest_minimize, gbrt_minimize])
 def test_early_stopping_delta_x(minimizer):
-    n_calls = 10
+    n_calls = 11
     res = minimizer(bench1,
                     callback=DeltaXStopper(0.1),
                     dimensions=[(-1., 1.)],
-                    x0=[[-0.1], [0.9], [0.1]],
+                    x0=[[-0.1], [0.1], [-0.9]],
```
betatim: Had to update this test after changing the seeding procedure as it relies on a particular sequence of numbers to trigger the stopping condition.

skopt/utils.py
```diff
@@ -222,6 +222,7 @@ def expected_minimum(res, n_random_starts=20, random_state=None):
     """
     def func(x):
         reg = res.models[-1]
+        x = res.space.transform(x.reshape(1, -1))
```
betatim: We forgot to transform the values that `sp_minimize` passes in. An alternative would be to run the minimizer on the transformed space/bounds (as we do internally). This is the 'minimal' fix though.

betatim: Opinions @glouppe or should we even move this to a separate PR?

skopt/optimizer/base.py
glouppe: i think it was only about performance.

skopt/utils.py
glouppe: It is good like this.

skopt/optimizer/base.py
betatim: Let's change it. It confused the heck out of me making demo notebooks :-/

skopt/space/transformers.py
iaroslav-ai: Yes I actually remember having some boundary checking issues some time ago, so I think this is a good idea.

skopt/optimizer/gp.py
```diff
@@ -206,14 +207,15 @@ def gp_minimize(func, dimensions, base_estimator=None,
     # Check params
     rng = check_random_state(random_state)
     space = normalize_dimensions(dimensions)
-    base_estimator = cook_estimator("GP", space=space, random_state=rng,
-                                    noise=noise)
+    base_estimator = cook_estimator(
+        "GP", space=space, random_state=rng.randint(0, np.iinfo(np.int32).max),
```
iaroslav-ai: I am wondering should we introduce a function in `skopt.utils` that does sampling of random integer based on random state variable? Might be less verbose then.

skopt/tests/test_forest_opt.py
```diff
@@ -56,9 +56,9 @@ def test_tree_based_minimize(name, minimizer):
     # the minimum, since for a decision tree any point greater than
     # max(sampled_points) would give a constant value.
     X0 = [[-5.6], [-5.8], [5.8], [5.6]]
-    check_minimize(minimizer, bench2, -5,
+    check_minimize(minimizer, bench2, -4.7,
```
iaroslav-ai: Was this breaking the test with the old numbers?

skopt/utils.py
```diff
@@ -327,7 +328,7 @@ def cook_estimator(base_estimator, space=None, **kwargs):
 
         base_estimator = GaussianProcessRegressor(
             kernel=cov_amplitude * other_kernel,
-            normalize_y=True, random_state=None, alpha=0.0, noise="gaussian",
+            normalize_y=True, alpha=0.0, noise="gaussian",
```
iaroslav-ai: Should we pass here also the sampled random integer as a random state?

skopt/tests/test_forest_opt.py
iaroslav-ai: (just want to know why these particular numbers used here and below)

skopt/space/transformers.py
betatim: Is there a better way to do this? `np.allclose` uses a combination of absolute and relative tolerances to handle large numbers as well. Or should we go with this for now and improve later?

skopt/utils.py
betatim: if you pass `random_state` to `cook_estimator` then it will be passed on to the regressor via `base_estimator.set_params(**kwargs)` a few lines later

skopt/tests/test_forest_opt.py
betatim: yes the tests were breaking. Makes me a bit uncomfortable because it isn't clear if these changes made the code more or less correct ... :-/

skopt/optimizer/gp.py
betatim: I'm -0 on that. We could add a `maxrange_randint()` or so but it would only be useful for the special case where you want the max range.

skopt/space/transformers.py
iaroslav-ai: Both are fine with me tbh. It seems like `np.allclose` might be a bit more advanced for use int these situations.

skopt/utils.py
iaroslav-ai: I see

skopt/tests/test_forest_opt.py
iaroslav-ai: Hm
Results with small number of n_calls for checking the procedure are normally quite noisy, so this might break simply by chance from time to time
There are a few options I see here:
1. Run the minimization procedure for some large number of repetitions, say 50, and then compare average of minimum objectives to some "normal" average min. This way results are more stable. To obtain such "normal" value, run the procedure with current minimizers, get avg min value, and eg add 0.1. Is computationally expensive - testing might time out.
2. Remove these tests. It might happen that some change leads to some large improvement in performance with large n_calls, but at a cost of a relatively small decrease in performance with smaller call budget, which would however breaks tests. 

I am leaning more towards option 1, but in principle fine with both options. For future: maybe there is an option to run slower tests somehow on a different infrastructure, where we could run such benchmarks regularly.

betatim: I'd like to keep the tests even if they aren't that meaningful. We haven't had to adjust them for quite a while and when we do they serve more as a reason to discuss things. Overall not optimal but IMHO better than having tests that take a long time to run.

If we could run (a subset) of the benchmarks you put together once a day that would be awesome. Travis has a cron feature but I wonder if the resources we get through that are enough or maybe we need to create a set of tests for "once a day travis" runs. New issue?

iaroslav-ai: Yeah it is also ok to leave it like that. A better benchmarking is indeed a different issue.

skopt/optimizer/gp.py
iaroslav-ai: Yeah then lets leave it as it is.

iaroslav-ai:
LGTM! :)

betatim:
Hopefully this helps with the CircleCI failures :)

iaroslav-ai:
:tada: !

skopt/utils.py
fredcallaway: I think this change warrants a new pip release.

PR #511, opened 2017-09-14
Fixed example in readme.

jusjusjus:
`+` makes definitely for a prettier function, and it's in line with the toy example in the Intro.


glouppe:
Good catch! Thanks for your contribution :)

PR #515, opened 2017-09-17
[MRG] Updates and examples related to BayesSearchCV

iaroslav-ai:
This PR contains additions related to BayesSearchCV. 
Firstly, now dimensions can be defined by sinlge element list, which is considered to be a dimension of Category type, which issimply used to fix value of a dimension. This is useful for definition of some subspaces, eg:

```python
search_space = {
        'model': [DecisionTreeClassifier()],
        'model__max_depth': (1, 32),
        'model__min_samples_split': (1e-3, 1.0, 'log-uniform'),
}
```

Secondly, I added possibility to monitor the progress of BayesSearchCV with event handler that is called on every step of subspaces exploration. For single job mode, this is called on every evaluation of model configuration, and for parallel mode this is called when n_jobs models are evaluated in parallel. Corresponding test is added. 

Example usage (copied from example notebook):

```python

from sklearn.datasets import load_iris
from sklearn.svm import SVC
from sklearn.model_selection import train_test_split

X, y = load_iris(True)

opt = BayesSearchCV(
    SVC(),
    search_spaces={
        'C': (0.0, 1.0)
    },
    n_iter=10
)

total_iters = opt.total_iterations()

# event handler
def on_step(iter):
    print("Iteration #%s / %s, best score: %s" % (iter, total_iters, opt.best_score_))

opt.on_step = on_step
    
opt.fit(X, y)

```

Finally, I also added example of how to use `error_score` parameter in order to handle possibly infeasible configurations of parameters.


betatim:
```
# event handler
def on_step(iter):
    print("Iteration #%s / %s, best score: %s" % (iter, total_iters, opt.best_score_))

opt.on_step = on_step
```
Some of scikit-learn's estimators have a `monitor` keyword argument for their `fit` method. Should we use that? Another alternative would be to support `callbacks` as a keyword argument to `BayesSearchCV` reusing the callbacks we already have got `*_minimize`? WDYT?

skopt/searchcv.py
```diff
@@ -675,6 +690,29 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
         if self.refit:
             self._fit_best_model(X, y)
 
+    def total_iterations(self):
```
betatim: Should we make this a property of `skopt.Space`? Somehow feels like that is where this should be. Need a good solution to adding up sub-spaces though.

glouppe:
> Another alternative would be to support callbacks as a keyword argument to BayesSearchCV reusing the callbacks we already have got *_minimize? WDYT?

+1 for that, though the `Optimizer` object has no callback API (this is the point)

skopt/searchcv.py
iaroslav-ai: Hm
What this does is it simply sums over all iterations that are set for the subspaces provided to the Optimizer. One can decide on different number of iterations for same subspace depending on how much patience and resources one has to wait for the algorithm to terminate. So far this seems rather like a property of optimizer than space. So I am -0.5 on moving this to space, especially if we do not have a way to define subspaces there yet

iaroslav-ai:
I like the first suggestion by @betatim with `monitor` keyword of `fit` method. The reason is that I often pickle classes for later usage, and if I provide lambda as a callback which is stored as a property of the class then all of a sudden I will have issues pickling model. Plus the callback function is only relevant during the fit stage. Thus I think the way to go is to have a callback property of the `fit` method that is not stored as property of the class. WDYT?

betatim:
Undecided :-/ Maybe the right thing to do is to encourage (and make it easy) for people to inherit from `BayesSearchCV` when they want to do "exotic" things like modify the search spaces as optimisation proceeds. Instead of trying to support a lot of flexibility by building stuff in. How do we find a method to decide what is "exotic" and what isn't? Right now I am thinking that we should be pretty conservative -> quickly label things as exotic.

Using the existing callbacks we have would get around the `lambda` pickling issue by encouraging you to not use lambdas :p

iaroslav-ai:
I like the idea of using existing `callbacks`

iaroslav-ai:
Also removed here unnecessary fits of the estimator.

iaroslav-ai:
Currently callbacks are an argument to the `fit` method of BayesSearchCV, as they are only necessary during fitting procedure :p
Suggestions welcome :)

iaroslav-ai:
Hm `skopt/tests/test_space.py` fails somehow :/

betatim:
The failing test in `test_space.py` is one that checks that a single element list/tuple raises an exception. So I think it is Ok to change that.

glouppe:
Changes here seem reasonable.

glouppe:
Can you make Travis happy?

skopt/searchcv.py
```diff
@@ -670,14 +677,38 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
-        optimizer.tell(params, [-score for score in local_results])
+        return optimizer.tell(params, [-score for score in local_results])
 
-        # fit the best model if necessary
-        if self.refit:
-            self._fit_best_model(X, y)
 
-    def fit(self, X, y=None, groups=None):
+
+    def total_iterations(self):
```
betatim: can this be a property please? `@property` should be enough

skopt/searchcv.py
```diff
@@ -272,7 +272,8 @@ def __init__(self, estimator, search_spaces=None, optimizer_kwargs=None,
                  n_iter=50, scoring=None, fit_params=None, n_jobs=1,
                  iid=True, refit=True, cv=None, verbose=0,
                  pre_dispatch='2*n_jobs', random_state=None,
-                 error_score='raise', return_train_score=True):
+                 error_score='raise', return_train_score=True,
+                 on_step=None):
```
betatim: do we still need this now that we have callback support? I am not a fan of the name and also it is now not clear anymore if you should use this or the callbacks or both or ..?

skopt/searchcv.py
```diff
@@ -670,14 +677,38 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
+            if isinstance(elem, tuple):
+                space, n_iter = elem
+            else:
+                n_iter = self.n_iter
+
+            total_iter += n_iter
+
+        return total_iter
+
+    def fit(self, X, y=None, groups=None, callback=None):
```
betatim: do we use callback or callbackS in the rest of the library?

betatim: callback. :)

skopt/searchcv.py
iaroslav-ai: Oops nope we do not need it

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Release history
 
+## Version 0.5
+
+Catchy phrase here :)
```
betatim: Ideas: "High five!" or "[Famous Five](https://en.wikipedia.org/wiki/The_Famous_Five_(novel_series))" or "Gimme five" or "[five by five](https://www.urbandictionary.com/define.php?term=five%20by%20five)" or "[Man alive](https://en.wikipedia.org/wiki/List_of_British_bingo_nicknames)"

iaroslav-ai:
I would say lets first merge this and then remove `step` and `add_spaces` in a different PR.

CHANGELOG.md
iaroslav-ai: I like High five! most - for others people might not get a reference :)

skopt/searchcv.py
iaroslav-ai: now a property

iaroslav-ai:
Anyone has any other comments? Otherwise merge

betatim:
Looks like we will be removing `add_spaces` and `step` in https://github.com/scikit-optimize/scikit-optimize/pull/551

skopt/tests/test_space.py
```diff
@@ -361,9 +361,14 @@ def test_valid_transformation():
...
                         space_check_dimension, "23")
+    """
+    Single value dimensions are ok - they fix a dimension value
+    This is useful for example when looking for subspaces in
+    BayesSearchCV, and you want to fix an estimator for subspace
+    
     assert_raises_regex(ValueError, "Invalid dimension",
                         space_check_dimension, (23,))
-
+    """
```
betatim: We should remove this instead of commenting it out by making it a string. Or even better move it to a test that checks that this doesn't raise an exception anymore :)

skopt/searchcv.py
```diff
@@ -620,6 +621,10 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
         n_jobs : int, default=1
             Number of parameters to evaluate in parallel.
 
+        refit : bool, default=True
+            Whether to refit the best found parameters after a step to the whole
```
betatim: -> "Fit estimator after each iteration with the best parameters found so far".

skopt/searchcv.py
```diff
@@ -306,6 +306,7 @@ def __init__(self, estimator, search_spaces=None, optimizer_kwargs=None,
         self.best_index_ = None
         self.multimetric_ = False
 
+
```
betatim: blank line, delete me

betatim:
screen shot 2017-11-04 at 17 32 52
In the example notebook something is missing after "during" on the second line of the text. Just not rendering for me or everyone?

"Progress monitoring using on_step callback" -> "Monitoring progress with a callback"? Feels like the name `on_step` is left over from before.

betatim:
Sorry for always finding new things to comment on :-/

glouppe:
would be good to have this  finished and merged :)

iaroslav-ai:
Yup, should be done soon :)

iaroslav-ai:
There has been quite some changes to BayesSearchCV which need to be integrated in PR

iaroslav-ai:
The continuation of this story is in #591 :)

iaroslav-ai:
P.S. suggestions by @betatim were integrated in #551

PR #508, opened 2017-09-10
Add property name to Dimension

mirca:
Implements #424 
cc @betatim

PR #518, opened 2017-09-18
[MRG] Beautifying plots: dim-labels, log-scale.

jusjusjus:
This request

* Uses space.dimensions[.].prior to conditionally plot in log-scale.
* Creates interface for the user to adjust labels.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -130,6 +130,13 @@ def _format_scatter_plot_axes(ax, space, ylabel, dim_labels=None):
                 else:
                     [l.set_rotation(45) for l in ax_.get_xticklabels()]
                     ax_.set_xlabel(dim_labels[j])
+                
+                xypriors = (space.dimensions[j].prior,
+                        space.dimensions[i].prior)
+                xysetters = (ax_.set_xscale, ax_.set_yscale)
+                for set_scale, prior in zip(xysetters, xypriors):
+                    if 'log-uniform' == prior:
+                        set_scale('log')
 
             else:
                 ax_.set_ylim(*diagonal_ylim)
@@ -142,8 +149,11 @@ def _format_scatter_plot_axes(ax, space, ylabel, dim_labels=None):
                 ax_.xaxis.set_label_position('top')
                 ax_.set_xlabel(dim_labels[j])
 
-            ax_.xaxis.set_major_locator(MaxNLocator(6, prune='both'))
-            ax_.yaxis.set_major_locator(MaxNLocator(6, prune='both'))
+                if 'log-uniform' == space.dimensions[i].prior:
+                    ax_.set_xscale('log')
+
+            # ax_.xaxis.set_major_locator(MaxNLocator(6, prune='both'))
+            # ax_.yaxis.set_major_locator(MaxNLocator(6, prune='both'))
 
     return ax
 
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -93,13 +93,16 @@ def plot_convergence(*args, **kwargs):
     return ax
 
 
-def _format_scatter_plot_axes(ax, space, ylabel):
+def _format_scatter_plot_axes(ax, space, ylabel, dim_labels=None):
     # Work out min, max of y axis for the diagonal so we can adjust
     # them all to the same value
     diagonal_ylim = (np.min([ax[i, i].get_ylim()[0]
                              for i in range(space.n_dims)]),
                      np.max([ax[i, i].get_ylim()[1]
                              for i in range(space.n_dims)]))
+
+    if dim_labels is None:
+        dim_labels = ["$X_{%i}$" % i for i in range(space.n_dims)]
 
     # Deal with formatting of the axes
     for i in range(space.n_dims):  # rows
@@ -118,7 +121,7 @@ def _format_scatter_plot_axes(ax, space, ylabel):
                 if j > 0:
                     ax_.set_yticklabels([])
                 else:
-                    ax_.set_ylabel("$X_{%i}$" % i)
+                    ax_.set_ylabel(dim_labels[i])
 
                 # for all rows except ...
                 if i < space.n_dims - 1:
@@ -126,7 +129,7 @@ def _format_scatter_plot_axes(ax, space, ylabel):
                 # ... the bottom row
                 else:
                     [l.set_rotation(45) for l in ax_.get_xticklabels()]
-                    ax_.set_xlabel("$X_{%i}$" % j)
+                    ax_.set_xlabel(dim_labels[j])
 
             else:
                 ax_.set_ylim(*diagonal_ylim)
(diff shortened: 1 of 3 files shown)


jusjusjus:
So I looked into the failed test around the parallel-optimization.ipynb, and I can't really figure out what fails here.  The notebook works fine on my pc.   Can someone please help me understand what's going on?    Thanks :)

betatim:
Most likely the failure on circleCI is unrelated to your changes. There have been a bunch of (random) failures that we are tracking down.

betatim:
Is this PR ready to be reviewed or are you still working on it? (If you want to use [WIP] for "work in progress" and [MRG] for "ready to merge from my side" at the start of the title of the PR to indicate status)

jusjusjus:
Thanks for the info.  I used the [WIP]/[MRG] codes now.  I'm sorry for the patchy syntax.  So what you get from this pull request is showcased as so:
```
from skopt.plots import plot_evaluations
from skopt.benchmarks import branin as branin
bounds = [
    (-5.0, 10.0), 
    (0.01, 15.0, 'log-uniform')
]
res = forest_minimize(branin, bounds, n_calls=25, base_estimator="ET",
                             random_state=42)
plot_evaluations(res, bins=10, dim_names=['Dim 1', 'Dim 2'])
plt.show()
```

skopt/space/space.py
```diff
@@ -97,6 +97,8 @@ def check_dimension(dimension, transform=None):
 class Dimension(object):
     """Base class for search space dimensions."""
 
+    prior = "uniform"
```
mirca: any reason behind hard coding this?

jusjusjus: That's a good questions: the problem is a missing implimentation of `Integer.prior`.  Now, I saw a couple of options:

* Impliment a stub in Integer.__init__.  The downside is that a new inherited class of `Dimension` would have the same problem if it doesn't impliment `prior`.

* Create a if/else in the plots.py code i added.  To my preference, this seemed ugly.

* Add `prior` to the base class (which is what I did in a minimally invasive way).  This could also be done via `super().__init__`, but would add a lot more code and inherited classes should remember to call `super().__init__` (see problem 1).

i the current solution is not good, but the best solution for this particular patch.

skopt/plots.py
```diff
@@ -239,8 +256,8 @@ def partial_dependence(space, model, i, j=None, sample_points=None,
         return xi, yi, np.array(zi).T
 
 
-def plot_objective(result, levels=10, n_points=40, n_samples=250, size=2,
-                   zscale='linear'):
+def plot_objective(result, levels=10, n_points=40, n_samples=250, size=2, zscale='linear',
+        objective_name="Partial dependence", dim_names=None):
```
betatim: Not a fan of `dim_names` as a name, how about just `dimensions`?

What is the use case for changing "Partial dependence" to something else?

---

formatting: can you indent this to line up with the opening bracket please?

skopt/plots.py
```diff
@@ -280,6 +297,13 @@ def plot_objective(result, levels=10, n_points=40, n_samples=250, size=2,
         Scale to use for the z axis of the contour plots. Either 'linear'
         or 'log'.
 
+    * `objective_name` [str, default='Partial dependence']
+        Label of the objective function to show on y-axis.
+
+    * `dim_names` [list of str, default=None]
+        Labels of the dimension variables. `None` defaults to `['X_0', 'X_1',
```
betatim: How about "Select a subset of dimensions from the search space and plot only those. By default all dimensions are plotted."?

skopt/plots.py
```diff
@@ -93,14 +93,17 @@ def plot_convergence(*args, **kwargs):
...
+def _format_scatter_plot_axes(ax, space, ylabel, dim_labels=None):
     # Work out min, max of y axis for the diagonal so we can adjust
     # them all to the same value
     diagonal_ylim = (np.min([ax[i, i].get_ylim()[0]
                              for i in range(space.n_dims)]),
                      np.max([ax[i, i].get_ylim()[1]
                              for i in range(space.n_dims)]))
 
+    if dim_labels is None:
+        dim_labels = ["$X_{%i}$" % i for i in range(space.n_dims)]
```
betatim: We could use `dimension.name` for those dimensions which have a name now that #508 has been merged. Default to X_i if there is no name. WDYT?

skopt/plots.py
```diff
@@ -137,10 +152,12 @@ def _format_scatter_plot_axes(ax, space, ylabel):
 
                 ax_.xaxis.tick_top()
                 ax_.xaxis.set_label_position('top')
-                ax_.set_xlabel("$X_{%i}$" % j)
+                ax_.set_xlabel(dim_labels[j])
 
-            ax_.xaxis.set_major_locator(MaxNLocator(6, prune='both'))
-            ax_.yaxis.set_major_locator(MaxNLocator(6, prune='both'))
+                if 'log-uniform' == space.dimensions[i].prior:
```
betatim: to keep with usual python style can you flip the order of the condition to `var == 'value'`, luckily `if var = value` is a syntax error in python

skopt/plots.py
```diff
@@ -368,7 +397,12 @@ def plot_evaluations(result, bins=20):
     for i in range(space.n_dims):
         for j in range(space.n_dims):
             if i == j:
-                ax[i, i].hist(samples[:, j], bins=bins,
+                if 'log-uniform' == space.dimensions[j].prior:
+                    b = space.bounds[j]
```
betatim: Maybe -> `low, high = space.bounds[j]`

(Part of Tim's campaign for nice variable names everywhere :) )

skopt/space/space.py
betatim: This is a bit tricky. I would probably prefer a default value of `None`, but this means we need a check in the plotting ... the reason to prefer `None` is that `prior` doesn't have to be a string, for example in `space.Categorical`. Not sure what the best options is right now.

betatim:
Something to think about: what happens if the space contains a `space.Categorical`? I have a feeling things will just fail right now. Something to fix in a new issue probably, but maybe we can take that into account when deciding things here.

Any ideas how we could add a test or two for this? Right now none of the plotting code has coverage which is sad, mostly because I didn't know how to test it when it first got written. Still not sure how to do that but maybe someone else here has some experience.

skopt/space/space.py
jusjusjus: sure, we could do `None`.  I believe though that `uniform` is more appropriate because assuming *no* prior just means that you accept a uniform prior across the specified interval/indices/categories.

jusjusjus:
I rebased to master and updated the code according to your suggestions.

Regarding `space.Categorical`, We could construct plottable dimensions along these lines:
`dims_to_plot = [not isinstance(dim, Categorical) for dim in space.dimensions]`, and go from there.  I think the property that we're missing is something like a `Dimensions.has_ordering`.  What do you think?

jusjusjus:
So, what's up?

glouppe:
Thanks @jusjusjus ! Could you rebase your branch? It seems to have some conflicts with master. 

Could you also provide some before/after screenshot so that it is easier for us to quickly give you feedback?

Besides this, it looks good to me and should be good to go after the rebase.

betatim:
Conflicts have magically fixed themselves? At least the button is green :) Merge?

jusjusjus:
Ok, let me run another rebase.

jusjusjus:
@glouppe , sorry to ask:  Screen shots of what?

PR #524, opened 2017-09-27
Fixing TyperError with Categorical of len gt 7

tgaudin:
Small bug fix to be able to print Categorical space with a length greather than 7.


betatim:
Thanks for spotting this. Could you add a test to help us increase coverage and stop the bug from coming back?

Otherwise this looks pretty uncontroversial!

iaroslav-ai:
I think this was solved somewhere else. If there is still a bug feel free to reopen!

betatim:
This was fixed in #610

PR #527, opened 2017-10-01
Fix sklearn model_selection import error

MechCoder:


betatim:
Thanks for this! Any idea why this was working on travis?

Issue #506, opened 2017-09-07
Automatic nonlinear transformation of input space


stefanocereda:
So far the input space is normalized into the unit hypercube in a linear way, or following a logarithm is specified by the user.
It would be nice to have an automatic decision on how to optimally wrap the search space. This is done, for example, in the paper "Input Warping for Bayesian Optimization of Non-Stationary Functions" (http://proceedings.mlr.press/v32/snoek14.html) and is (I think) the default in the Spearmint program.

It would be nice to add this functionality to scikit-optimize.
If we consider the code of Spearmint (https://github.com/HIPS/Spearmint/blob/master/spearmint/transformations/beta_warp.py) the implementation seems pretty straightforward: they just define an Alpha and Beta hyperparameters and use them to compute the forward and backward transformation.

What is blocking me is how to tune those Alpha and Beta. From the paper they say that they use a prior distribution for those parameters, and then they can be integrated out. I do not know how to implement this step in scikit-optimize. From my understanding this should be similar to what is done with the kernel's length scale, so if anybody could give me an hint it would be much appreciated.

stefanocereda:
I've found the code for the optimization of the kernel's hyperparameters: this is actually done in scikit-learn (https://github.com/scikit-learn/scikit-learn/blob/ef5cb84a/sklearn/gaussian_process/gpr.py#L203)

I'd like to use a similar approach to tune the hyperparameters of the input space normalization, but I would also like to do this in an extensible way, so that in the future we can optimize any kind of hyperparameter. This would greatly simplify the usage of scikit-optimize in real world scenarios.

Does anybody have some ideas on how it would be better to perform this?

MechCoder:
Look at the discussion here (https://github.com/scikit-optimize/scikit-optimize/issues/434)

TL;DR Including such a mapping means that there is no closed-form solution for the derivative of the log-likelihood with respect to the kernel hyperparameters. So, one way to do this is simply sample from posterior of the kernel hyperparameters P(theta | data) instead of finding the argmax of log(P(data | theta))
which is implemented in the original spearmint paper.

stefanocereda:
Quick update: I have a working implementation of this and preliminary tests show really great results. I'm actually doing this in a slightly different way wrt spearmint, the optimization is performed by gradient descent instead of mcmc.
I should refine a little bit the implementation before pushing it, but I really think that we should come up with a solid solution able to handle every kind of hyperparameter.

betatim:
[...]

PR #529, opened 2017-10-02
[WIP] input space warping

stefanocereda:
As discussed in #506 this is not for merge, but just to discuss how to better implement this feature

### DIFF
--- a/skopt/learning/__init__.py
+++ b/skopt/learning/__init__.py
@@ -4,10 +4,11 @@ from skgarden import RandomForestRegressor
 from skgarden import ExtraTreesRegressor
 
 from .gbrt import GradientBoostingQuantileRegressor
-from .gaussian_process import GaussianProcessRegressor
+from .gaussian_process import GaussianProcessRegressor, GaussianProcessRegressor_BetaWarped
 
 
 __all__ = ("RandomForestRegressor",
            "ExtraTreesRegressor",
            "GradientBoostingQuantileRegressor",
-           "GaussianProcessRegressor")
+           "GaussianProcessRegressor",
+           "GaussianProcessRegressor_BetaWarped")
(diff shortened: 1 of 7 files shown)


betatim:
Thanks for this @stefanocereda. Who do you think would be a good person to pull into this discussion (we should ping them)? Personally I don't know very much about the input space warping, so I can't give you any useful feedback :)

We discussed that having a `Transformer` would be a good way to fit in with what is currently being done in skopt and that that isn't very easy to do. I'm going to be the annoying person who pushes for us finding a way to implement this as a transformer so it can also be used by other surrogates 👮 

Possibly naive questions/statements to start the ball rolling:
* beta warping is similar to log transforming or normalizing a dimension?
* unlike normalizing a dimension beta warping has parameters that need adjusting during the fit
* there is no way for the surrogate model to reach the instance of `Space` that it is operating on, and so can't adjust the parameters of the beta warp.

---

One thing that might help steer the discussion is if you can make some inline comments in the diff with questions/comments/pointers. 800lines of diff look intimidating to anyone trying to review them, with some pointers it will look less scary :)

stefanocereda:
Ok, this should be compared against the gpr regressor provided by scikit-learn, I will try to provide a shorter pull-request with just the relevant code.

Yes the idea is to have a "tunable transformer", therefore I do not think it would be feasible to implement it as a transformer if that would prevent us from tuning it.

I will provide a shorter commit so that it becomes easier to reason about it.

stefanocereda:
I've refactored the commit. I've created a new file (gpr_betaWarp) which is a verbatim copy of gpr.py, except it includes a different GPR, you should not focus on this file.

The important changes are inside warped_gpr.py, which extends the gpr defined in sklearn. Since The modifications are in the middle of the methods, I had to repeat much code. The relevant changes are  highlighted with a # DIFFERENCE comment.

You can use this regressor with gp_minimize simply by adding a use_betaWarp=true parameter.

Let me know what you think!

stefanocereda:
Maybe it would be feasible to implement the transformer as a kernel? In this way it could be easy to access the hyperparameters and tune them, but I wouldn't know how to apply them

PR #535, opened 2017-10-13
[MRG] gp_minimize should not use default estimator when a custom one is provided

stefanocereda:
I think that gp_minimize should use the given optimizer, if it exists

### DIFF
--- a/skopt/tests/test_gp_opt.py
+++ b/skopt/tests/test_gp_opt.py
@@ -10,6 +10,8 @@ from skopt.benchmarks import bench2
 from skopt.benchmarks import bench3
 from skopt.benchmarks import bench4
 from skopt.benchmarks import branin
+from skopt.utils import cook_estimator
+from skopt.utils import normalize_dimensions
 
 
 def check_minimize(func, y_opt, bounds, acq_optimizer, acq_func,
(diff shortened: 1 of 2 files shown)


betatim:
Indeed! Could you add a test so this bug doesn't come back?

stefanocereda:
I've added a test, let'me know if it's ok as it's my first one

skopt/tests/test_gp_opt.py
```diff
@@ -71,3 +73,17 @@ def test_gpr_default():
     """Smoke test that gp_minimize does not fail for default values."""
     gp_minimize(branin, ((-5.0, 10.0), (0.0, 15.0)), n_random_starts=1,
                 n_calls=2)
+
+
+@pytest.mark.fast_test
+def test_use_given_estimator():
+    """ Check that gp_minimize use the given estimator. """
+    domain = [(1.0, 2.0), (3.0, 4.0)]
+    space = normalize_dimensions(domain)
```
betatim: do we need this line?

skopt/tests/test_gp_opt.py
```diff
@@ -71,3 +73,17 @@ def test_gpr_default():
     """Smoke test that gp_minimize does not fail for default values."""
     gp_minimize(branin, ((-5.0, 10.0), (0.0, 15.0)), n_random_starts=1,
                 n_calls=2)
+
+
+@pytest.mark.fast_test
+def test_use_given_estimator():
+    """ Check that gp_minimize use the given estimator. """
```
betatim: I'd change this to `# Test that gp_minimize does not use default estimator if one is passed in explicitly.`

skopt/tests/test_gp_opt.py
stefanocereda: I have to give a Space to cook_estimator so I think it's needed

betatim:
If you rebase your branch on the latest master the CircleCI failure should be fixed as well.

skopt/tests/test_gp_opt.py
betatim: `cook_estimator("GP", [(1.0, 2.0), (3.0, 4.0)], noise=1234)` also works no? To save the call to `normalize_dimensions`

betatim:
Thanks for catching this bug! :clinking_glasses:

PR #538, opened 2017-10-15
Add a RepeatedMinStopper callback and tests.

smutch:
This callback stops the optimisation when there has been no improvement
in the achieved minimum after `n_best` steps.

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -238,8 +238,6 @@ class RepeatedMinStopper(EarlyStopper):
         if result.fun < self.minimum:
             self.minimum = result.fun
             self.count = 0
-        elif result.fun > self.minimum:
-            self.count = 0
         else:
             self.count += 1
 
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -222,6 +222,30 @@ class DeltaYStopper(EarlyStopper):
             return None
 
 
+class RepeatedMinStopper(EarlyStopper):
+    """Stop the optimization when there is no improvement in the minimum.
+
+    Stop the optimization when there is no improvement in the minimum
+    achieved function evaluation after `n_best` iterations.
+    """
+    def __init__(self, n_best=50):
+        super(EarlyStopper, self).__init__()
+        self.n_best = n_best
+        self.count = 0
+        self.minimum = np.finfo(np.float).max
+
+    def _criterion(self, result):
+        if result.fun < self.minimum:
+            self.minimum = result.fun
+            self.count = 0
+        elif result.fun > self.minimum:
+            self.count = 0
+        else:
+            self.count += 1
+
+        return self.count >= self.n_best
+
+
 class DeadlineStopper(EarlyStopper):
     """
     Stop the optimization before running out of a fixed budget of time.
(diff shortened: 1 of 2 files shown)


skopt/callbacks.py
```diff
@@ -222,6 +222,30 @@ def _criterion(self, result):
...
+        self.minimum = np.finfo(np.float).max
+
+    def _criterion(self, result):
+        if result.fun < self.minimum:
+            self.minimum = result.fun
+            self.count = 0
+        elif result.fun > self.minimum:
+            self.count = 0
+        else:
+            self.count += 1
```
betatim: Is there a typo in how the if statements are written? It seems like the `else` clause would never get evaluated and so we never trigger the stop?? 😕

betatim:
Thanks for opening this PR. Like your idea that code will make it more concrete :)

I thought I had figured it all out 2minutes ago, but now I can't write it down :-/ so instead this is what I think `DeltaYStopper` does: it stops if there are `n` function evaluations all within `delta`. The idea is that once you have found "the true" minimum then you will repeatedly evaluate it and find values close to each other. It does not look the order ("last N evaluations") because the optimizer will sometimes jump to a random location to do "exploration" which would break the chain of results.

smutch:
Good spot on the typo there.  That's a little embarrassing!  I think that `elif` was unreachable since `self.minimum >= result.fun` at all times.  I've updated the PR.

I agree with your description of `DeltaYStopper`, which makes me think that my proposed `RepeatedMinStopper` is perhaps not suitable for the bayesian optimisation method...  The idea behind it was that if there are `n_best` (a large number) evaluations and the achieved best minimum never changed then chances are that you are already close to the true minimum and can probably stop.  If I understand right then `DeltaYStopper` effectively requires the user to have some prior knowledge of the objective distribution to assess what is a suitable delta-y is.

I think the issue is that I am using (abusing?) the methodology to just give me a quick way to get close to the global minimum for a 8 parameter model with a set of different objective functions, without the need for any manual intervention.  In hindsight - this is maybe therefore not a generally useful addition to the code?

(p.s. Apologies if I have misused some of the terminology here! 😬 )

betatim:
> The idea behind it was that if there are `n_best` (a large number) evaluations and the achieved best minimum never changed then chances are that you are already close to the true minimum and can probably stop.

Agreed. The thing I am struggling with is that I think `DeltaYStopper` should work for you, at least that is the use case I had when creating `DeltaYStopper`.

> If I understand right then `DeltaYStopper` effectively requires the user to have some prior knowledge of the objective distribution to assess what is a suitable delta-y is.

Correct. If you have an objective that has no noise you can set it very small, but as soon as you have a noisy objective function then you need to help skopt out by setting it by hand. I'm not sure there is a good way to estimate the noise of an unknown objective just from the observations you have??

Overall I feel like I am missing something here as I think `DeltaYStopper` should work for your use case :-/

smutch:
I'm really sorry for my delay in getting back to you regarding this.  I am just now returning to this topic...

I think I'll close this PR as, after our discussion, I'm pretty sure that this code does do something different to `DeltaYStopper`, however, I think it is "the wrong thing to do" and likely not a good idea for robust results.

I just want to close off by saying thanks again for the discussion and for the project!

glouppe:
Thanks @smutch ! However, feel free to contribute to the improvement of the callbacks in general. We value deeply real-world experience and any idea good idea or suggestion to make them better for practical use is welcome!

Issue #539, opened 2017-10-17
printing minimization result with categorical variable with more than 7 categories fails


andreh7:
I tried this (the stack trace below is with the current head 39528092868fcb24c598ff2dcd9f90b1b55a1d81 but I originally saw the problem with skopt 0.4):

```
import numpy as np
from skopt import dummy_minimize
from skopt.space import Categorical

def f(x):
    print "evaluating at",x
    return np.random.randn()

res = dummy_minimize(f,
                    [
                        Categorical([1,2,3,4,5,6,7,8])
                    ])

print res
```
which fails on the print statement with the following stack trace:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../anaconda/envs/test/lib/python2.7/site-packages/scipy/optimize/optimize.py", line 125, in __repr__
    for k, v in sorted(self.items())])
  File ".../anaconda/envs/test/lib/python2.7/site-packages/skopt/space/space.py", line 529, in __repr__
    return "Space([{}])".format(',\n       '.join(map(str, dims)))
  File ".../anaconda/envs/test/lib/python2.7/site-packages/skopt/space/space.py", line 434, in __repr__
    cats = self.categories[:3] + [_Ellipsis()] + self.categories[-3:]
TypeError: can only concatenate tuple (not "list") to tuple
```

when I remove the last element from the list (resulting in a list with just seven elements) given to `Categorical()`, the above code works without problems. 

Is this expected ?

andreh7:
I think I found the origin of the problem

* `self.categories` is initialized to be a tuple in class `Categorical` here: 
https://github.com/scikit-optimize/scikit-optimize/blob/39528092868fcb24c598ff2dcd9f90b1b55a1d81/skopt/space/space.py#L400
* `__repr__` of this class then tries to concatenate parts of `self.categories` with a list just containing an ellipsis in this part: https://github.com/scikit-optimize/scikit-optimize/blob/39528092868fcb24c598ff2dcd9f90b1b55a1d81/skopt/space/space.py#L432-L441

A quick solution would be to use `(_Ellipsis(),)` instead of `[_Ellipsis()]` . If there are no concerns with this solution, I can create a pull request.

betatim:
Looks like a good solution to me. PR welcome especially if you add a test to stop this from coming back 😄

andreh7:
I prepared a test and fix for printing the actual categorical values.

[...]

PR #543, opened 2017-10-20
[WIP] test and fix for printing Categorical variables

andreh7:
as reported and discussed in #539 .

The test implemented in the first commit 57b1336 fails (which is expected) but passes with the second commit (a78c6ba, actual fix for printing the categories themselves).

Please do not merge yet because there are still some remaining questions on how to deal with the `prior` argument.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -384,7 +384,7 @@ class Categorical(Dimension):
         * `categories` [list, shape=(n_categories,)]:
             Sequence of possible categories.
 
-        * `prior` [list, shape=(categories,), default=None]:
+        * `prior` [list, shape=(n_categories,), default=None]:
             Prior probabilities for each category. By default all categories
             are equally likely.
 
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -431,7 +431,7 @@ class Categorical(Dimension):
 
     def __repr__(self):
         if len(self.categories) > 7:
-            cats = self.categories[:3] + [_Ellipsis()] + self.categories[-3:]
+            cats = self.categories[:3] + (_Ellipsis(),) + self.categories[-3:]
         else:
             cats = self.categories
 
--- a/skopt/tests/test_space.py
+++ b/skopt/tests/test_space.py
@@ -381,7 +381,7 @@ def test_categorical_identity():
 def test_categorical_repr(capsys):
 
     # ensure printing of Categorical objects does not
-    # raise an exception. Hhowever we do not insist
+    # raise an exception. However we do not insist
     # on the exact output format.
 
     for categories in (
--- a/skopt/tests/test_space.py
+++ b/skopt/tests/test_space.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import pytest
 import numbers
 import numpy as np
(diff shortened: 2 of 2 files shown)


skopt/tests/test_space.py
```diff
@@ -375,6 +377,27 @@ def test_categorical_identity():
...
+        range(3), # short ( < 7 items) list of categories
+        range(8), # long ( >= 7 items) list of categories with ellipsis
+        ):
+
+        categories = list(categories)
+        cat = Categorical(categories, transform="identity")
+
+        out = StringIO.StringIO()
+
+        print(cat, file = out)
```
betatim: Instead of using this hand made solution check out https://docs.pytest.org/en/3.2.3/capture.html#accessing-captured-output-from-a-test-function for capturing stdout and then `assert`ing its contents. This should also fix the travis failures.

betatim:
Thanks for opening this PR. Should we move the discussion here from the issue or keep discussing there?

While reading the docstring for `Categorical` I also noticed that the prior documentation has the wrong shape: 
https://github.com/scikit-optimize/scikit-optimize/blob/39528092868fcb24c598ff2dcd9f90b1b55a1d81/skopt/space/space.py#L387 could you replace `categories` with `n_categories` on that line? Basically to make it consistent with the categories argument.

---

To indicate that this PR is still in progress we usually use `[WIP] <title>` for a PR. Take a look at our `CONTRIBUTING.md` for some of the other idiosyncrasies of our development process.

skopt/tests/test_space.py
andreh7: Thanks for pointing this out (I would not have thought that such a solution existed :-) ), a406a8a has this now . It does not check the actual output format but essentially only assures that no exception is thrown. 

Do you want the test to also to require the exact output string it produces at the moment ?

skopt/space/space.py
```diff
@@ -384,7 +384,7 @@ def __init__(self, categories, prior=None, transform=None, name=None):
         * `categories` [list, shape=(n_categories,)]:
             Sequence of possible categories.
 
-        * `prior` [list, shape=(categories,), default=None]:
+        * `prior` [list, shape=(n_categories,), default=None]:
```
betatim: 🎉

skopt/tests/test_space.py
```diff
@@ -375,6 +377,22 @@ def test_categorical_identity():
     assert_array_equal(transformed, samples)
     assert_array_equal(samples, cat.inverse_transform(transformed))
 
+@pytest.mark.fast_test
+def test_categorical_repr(capsys):
+
+    # ensure printing of Categorical objects does not
+    # raise an exception. Hhowever we do not insist
```
betatim: Hhowever -> However

skopt/tests/test_space.py
```diff
@@ -375,6 +377,22 @@ def test_categorical_identity():
...
+    # on the exact output format.
+
+    for categories in (
+        range(3), # short ( < 7 items) list of categories
+        range(8), # long ( >= 7 items) list of categories with ellipsis
+        ):
```
betatim: You can use something along the lines of `@pytest.mark.parametrize("categories", [range(5), range(9)])` to get rid of the for loop inside the test. I think there is an example of this decorator further down in this file.

skopt/tests/test_space.py
andreh7: oops :-) fixed in 6424c3e

skopt/tests/test_space.py
andreh7: Thanks for pointing this out. Changed now in 9804c0e .

I checked with `pytest -s` that it actually executes the test twice with the correct parameters.

betatim:
Superseded by #610

PR #544, opened 2017-10-22
[MRG] Fix not positive definitive matrix in parallel optimisation

betatim:
This does two things: tries to fix the Circlie CI error we keep seeing about "not pos def matrix" in the parallel evaluation example and makes the calls with a parallel strategy reproducible by fixing the seed.

### DIFF
--- a/examples/parallel-optimization.ipynb
+++ b/examples/parallel-optimization.ipynb
@@ -1,129 +1,132 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
-   "source": [
-    "# Parallel optimization\n",
-    "\n",
-    "Iaroslav Shcherbatyi, May 2017.\n",
-    "\n",
-    "Reviewed by Manoj Kumar and Tim Head."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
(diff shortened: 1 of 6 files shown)


iaroslav-ai:
Fine with me as a solution to current circleci failures

iaroslav-ai:
Ahaaaaaaaaa so the difference to sklearn implementation is in that there is a stabilization of matrix inverse using alpha > 0.0 
Consider adding this as a test (fails when alpha is set to 0.0):
```python
def test_gpr_handles_similar_points():
    """
    This tests whether our implementation of GPR 
    does not crash when the covariance matrix whose
    inverse is calculated during fitting of the 
    regressor is singular. 
    Singular covariance matrix often indicates
    that same or very close points are explored 
    during the optimization procedure.
    """

    import numpy as np
    from skopt.learning.gaussian_process.gpr import GaussianProcessRegressor
    
    X = np.random.rand(8, 3)
    y = np.random.rand(8)
    
    X[:3,:] = 0.0
    y[:3] = 1.0

    model = GaussianProcessRegressor()
    # this fails if singular matrix is not handled
    model.fit(X, y)
```

iaroslav-ai:
So LGTM

betatim:
Test added.

betatim:
Merging this.

iaroslav-ai:
Nice. That was a long standing issue

PR #549, opened 2017-11-01
[WIP] Add scikit-garden forest estimators to skopt

### LINKED ISSUE #548: Inlining scikit-garden code instead of dependency
We currently depend on `scikit-garden` which is a bit tricky to install as it requires `numpy` to be already installed before you install scikit-garden. Users sometimes trip over this, we have a special install instruction and making a [binder](//mybinder.org) that uses scikit-optimize is tricky (you can't just list scikit-garden in the `requirements.txt`).

What makes scikit-garden tricky to install is its use of cython which is only needed for the mondrian forest part (I think). My suggestion would be to move the code that we use back into scikit-optimize in the same place we had it before and stop depending on scikit-garden. An alternative would be for someone to actively maintain scikit-garden and invest the time in fixing issues like this.

We re-asses this when we make use of mondrian forests in scikit-optimize at which point we can also contribute to the maintenance burden. Right now it seems like we shoulder ourselves with a lot of tedious maintenance for no benefit because we don't use the mondrian forests.

What are your opinions around this? In particular @MechCoder as you are one of the main scikit-garden maintainers.

betatim:
Fixes #548 

Move the code back into scikit-optimize to reduce the number of
dependencies of this library. scikit-garden provides mondiran forests
which require cython and hence a compiler. This adds complexity that we
do not yet need as we don't use mondrian forests.


glouppe:
Merge?

iaroslav-ai:
still times out huh
LGTM anywasy

betatim:
The timeout is annoying, especially as it didn't used to. So the question is: how could this have made it slower??

PR #551, opened 2017-11-03
[MRG+1] Fix compatibility with sklearn

### LINKED ISSUE #550: CV fit method should return self just like GridSearchCV in Sci-kit learn
In BayesSearchCV class, fit method does Not return self. It's inconsistent with the way sci-kit learn is built.
https://github.com/scikit-optimize/scikit-optimize/blob/b3cc7f6897c604988ab0a9e983adcd3cb11c301e/skopt/searchcv.py#L680
Here is fit method of GridSearchCV in sklearn. 
https://github.com/scikit-learn/scikit-learn/blob/8e599c68fb06967313d0e76a053062158cd90312/sklearn/model_selection/_search.py#L749

fcharras:
Doing *anything* in the `__init__` method of a sklearn estimator will most of the time break his compatibility with sklearn's clone, get_params and set_params, or the ability to fit it several times. It is in fact the case here, it would break if used in a Pipeline or other meta-objects in sklearn.

This PR is a quick (for now, dirty) proposal to fix the compatibility issues.

I've basically moved every lines of code that are not `self.foo = foo`, where foo is an arg, inside the fit() and add_space() methods.

Fixes #550 #551

### DIFF
--- a/examples/sklearn-gridsearchcv-replacement.ipynb
+++ b/examples/sklearn-gridsearchcv-replacement.ipynb
@@ -2,7 +2,10 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "# Scikit-learn hyperparameter search wrapper\n",
     "\n",
@@ -11,7 +14,10 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "## Introduction\n",
     "\n",
@@ -31,7 +37,11 @@
   {
    "cell_type": "code",
    "execution_count": 11,
-   "metadata": {},
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
    "outputs": [
     {
      "name": "stdout",
@@ -71,7 +81,10 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "## Advanced example \n",
     "\n",
@@ -81,7 +94,11 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
    "outputs": [
     {
      "name": "stdout",
@@ -144,8 +161,8 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (Spyder)",
-   "language": "python3",
+   "display_name": "Python 3",
+   "language": "python",
    "name": "python3"
   },
   "language_info": {
@@ -158,9 +175,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.3"
+   "version": "3.5.2"
   }
  },
  "nbformat": 4,
- "nbformat_minor": 1
+ "nbformat_minor": 0
 }
--- a/examples/sklearn-gridsearchcv-replacement.ipynb
+++ b/examples/sklearn-gridsearchcv-replacement.ipynb
@@ -2,10 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "# Scikit-learn hyperparameter search wrapper\n",
     "\n",
@@ -14,10 +11,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## Introduction\n",
     "\n",
@@ -36,24 +30,27 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 11,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true
-   },
+   "execution_count": 6,
+   "metadata": {},
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/usr/lib/python3.6/site-packages/sklearn/model_selection/_split.py:2026: FutureWarning: From version 0.21, test_size will always complement train_size unless both are specified.\n",
+      "  FutureWarning)\n"
+     ]
+    },
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "val. score: 0.983667409057\n",
+      "val. score: 0.985894580549\n",
       "test score: 0.982222222222\n"
      ]
     }
    ],
    "source": [
-    "from skopt import BayesSearchCV\n",
     "from sklearn.datasets import load_digits\n",
     "from sklearn.svm import SVC\n",
     "from sklearn.model_selection import train_test_split\n",
@@ -81,10 +78,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## Advanced example \n",
     "\n",
@@ -93,19 +87,23 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true
-   },
+   "execution_count": 7,
+   "metadata": {},
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/usr/lib/python3.6/site-packages/sklearn/model_selection/_split.py:2026: FutureWarning: From version 0.21, test_size will always complement train_size unless both are specified.\n",
+      "  FutureWarning)\n"
+     ]
+    },
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "val. score: 0.985894580549\n",
-      "test score: 0.982222222222\n"
+      "val. score: 0.991833704529\n",
+      "test score: 0.993333333333\n"
      ]
     }
    ],
@@ -161,8 +159,8 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
+   "display_name": "Python 3 (Spyder)",
+   "language": "python3",
    "name": "python3"
   },
   "language_info": {
@@ -175,9 +173,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.5.2"
+   "version": "3.6.3"
   }
  },
  "nbformat": 4,
- "nbformat_minor": 0
+ "nbformat_minor": 1
 }
--- a/examples/sklearn-gridsearchcv-replacement.ipynb
+++ b/examples/sklearn-gridsearchcv-replacement.ipynb
@@ -2,10 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "# Scikit-learn hyperparameter search wrapper\n",
     "\n",
@@ -14,10 +11,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## Introduction\n",
     "\n",
@@ -37,11 +31,7 @@
   {
    "cell_type": "code",
    "execution_count": 11,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "outputs": [
     {
      "name": "stdout",
@@ -81,10 +71,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "source": [
     "## Advanced example \n",
     "\n",
@@ -94,11 +81,7 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true
-   },
+   "metadata": {},
    "outputs": [
     {
      "name": "stdout",
@@ -157,87 +140,12 @@
     "print(\"val. score: %s\" % opt.best_score_)\n",
     "print(\"test score: %s\" % opt.score(X_test, y_test))"
    ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "deletable": true,
-    "editable": true
-   },
-   "source": [
-    "## Iterative search utilizing `step` function\n",
-    "\n",
-    "A single call to `step` function of the `BayesSearchCV` instance allows to perform partial search for at most `n_jobs` points in parallel. This allows to use custom stopping criterions and to pickle a parameter search class instance to the file for recovery in case of failures of any sort or in case computations need to be transported to a different machine. \n",
-    "\n",
-    "An example usage is shown below. "
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {
-    "collapsed": false,
-    "deletable": true,
-    "editable": true,
-    "scrolled": true
-   },
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "0 0.0955555555556\n",
-      "1 0.0955555555556\n",
-      "2 0.94\n",
-      "3 0.94\n",
-      "4 0.94\n",
-      "5 0.94\n",
-      "6 0.975555555556\n",
-      "7 0.975555555556\n",
-      "8 0.975555555556\n",
-      "9 0.971111111111\n",
-      "10 0.991111111111\n",
-      "11 0.993333333333\n",
-      "12 0.993333333333\n",
-      "13 0.993333333333\n",
-      "14 0.993333333333\n",
-      "15 0.993333333333\n"
-     ]
-    }
-   ],
-   "source": [
-    "from skopt.space import Real, Categorical, Integer\n",
-    "from skopt import BayesSearchCV\n",
-    "\n",
-    "from sklearn.datasets import load_digits\n",
-    "from sklearn.svm import SVC\n",
-    "from sklearn.model_selection import train_test_split\n",
-    "\n",
-    "X, y = load_digits(10, True)\n",
-    "X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=0.75, random_state=0)\n",
-    "\n",
-    "opt = BayesSearchCV(SVC())\n",
-    "\n",
-    "# add instance of a search space \n",
-    "opt.add_spaces('space_1', {\n",
-    "    'C': Real(1e-6, 1e+6, prior='log-uniform'),\n",
-    "    'gamma': Real(1e-6, 1e+1, prior='log-uniform'),\n",
-    "})\n",
-    "\n",
-    "for i in range(16):\n",
-    "    opt.step(X_train, y_train, 'space_1')\n",
-    "    # save the model or use custom stopping criterion here\n",
-    "    # model is updated after every step\n",
-    "    # ...\n",
-    "    score = opt.score(X_test, y_test)\n",
-    "    print(i, score)"
-   ]
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
+   "display_name": "Python 3 (Spyder)",
+   "language": "python3",
    "name": "python3"
   },
   "language_info": {
@@ -250,9 +158,9 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.5.2"
+   "version": "3.6.3"
   }
  },
  "nbformat": 4,
- "nbformat_minor": 0
+ "nbformat_minor": 1
 }
--- a/examples/sklearn-gridsearchcv-replacement.ipynb
+++ b/examples/sklearn-gridsearchcv-replacement.ipynb
@@ -2,7 +2,10 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "# Scikit-learn hyperparameter search wrapper\n",
     "\n",
@@ -11,7 +14,10 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "## Introduction\n",
     "\n",
@@ -30,27 +36,24 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
-   "metadata": {},
+   "execution_count": 11,
+   "metadata": {
+    "collapsed": false,
+    "deletable": true,
+    "editable": true
+   },
    "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "/usr/lib/python3.6/site-packages/sklearn/model_selection/_split.py:2026: FutureWarning: From version 0.21, test_size will always complement train_size unless both are specified.\n",
-      "  FutureWarning)\n"
-     ]
-    },
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "val. score: 0.985894580549\n",
+      "val. score: 0.983667409057\n",
       "test score: 0.982222222222\n"
      ]
     }
    ],
    "source": [
+    "from skopt import BayesSearchCV\n",
     "from sklearn.datasets import load_digits\n",
     "from sklearn.svm import SVC\n",
     "from sklearn.model_selection import train_test_split\n",
@@ -78,7 +81,10 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "deletable": true,
+    "editable": true
+   },
    "source": [
     "## Advanced example \n",
     "\n",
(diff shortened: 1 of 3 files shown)


glouppe:
Thanks a lot! Let us know when it is ready for review.

fcharras:
Haven't tested it yet I might try it within next week. Also, let's see if the checks passes. Aside from this the changes are minors so I guess it can be looked at. (mostly copy/paste code in init, to the top of add_space() and fit())

betatim:
Thanks for spotting this and fixing it! To make bookkeeping easier could you add a line "Fixes #550 #551" in the top most comment for this PR? That way the issues get closed automatically when this gets merged. 📋

skopt/searchcv.py
```diff
@@ -268,43 +268,17 @@ class BayesSearchCV(BaseSearchCV):
 
     """
 
-    def __init__(self, estimator, search_spaces=None, optimizer_kwargs=None,
+    def __init__(self, estimator, search_spaces=None, optimizer_kwrgs=None,
```
betatim: `kwargs` is the standard in python so I think we should keep it

betatim:
General comment: having to use `hasattr` this much feels wrong to my python brain. The scikit-learn design principle says we should store all arguments to `__init__` as attributes, and not compute anything until `fit()` is called. All computed attributes are stored as `self.blah_`. Do we need to have `add_spaces()` and `step()`? This adds to the API which is trying to copy the scikit-learn which we are trying to replicate. It is also the source of all the `hasattr`s?

Not sure what the conclusion is, but wanted to start a discussion.

skopt/searchcv.py
fcharras: Thank you for reviewing.

Agreed, it should be optimizer_kwargs, and all the current optimizer_kwargs should be optimizers_kwargs_.

I'll make a new, clean pass for variable names and refactoring. I think it's going to pass all checks this time.

fcharras:
I very much agree with this, the current API that add_spaces and step offer could be dropped, it would make easier to stick to the standard sklearn API.

betatim:
Let's see what the others think.

iaroslav-ai:
First of all, thanks @fcharras for starting this :)
Yes, `add_spaces` and `step` should be removed - this will make things a lot easier. The idea was to have ask-and-tell interface for *SearchCV, but this can be achieved in a more simple / standard ways, like using callbacks #515.

fcharras:
I think the most straightforward change would be to consider that `add_spaces` and `step` are privates, and only called during the `fit`. This would enable defining all the attributes `self.foo_ = ...` only at the beginning of the `fit`. Then the definitions in `step` and `add_spaces` and all the `hasattr` checks could be removed. Tests and documentation on `add_space` and `step` should be updated / deleted accordingly.

iaroslav-ai:
@fcharras I agree - that would be the easiest solution. If you feel like / have time for implementing this I personally would be very happy if you do so :) Alternatively I will make a PR later.

fcharras:
Thank you ! I'd like to keep pushing this PR, new commits soon.

fcharras:
Here's a first shot. I finally totally deleted add_spaces because most of its content became obsolete with the change. Also deleted and moved some parts of step into fit.

examples/sklearn-gridsearchcv-replacement.ipynb
```diff
@@ -2,10 +2,7 @@
  "cells": [
```
fcharras: Woops, the changes in "metadata' are a mistake.

fcharras:
I think it can be reviewed. All the tests and notebooks passes on my local machine. The notebooks are somewhat long to run on CircleCI though.

skopt/searchcv.py
```diff
@@ -587,77 +509,37 @@ def _make_optimizer(self, params_space):
...
 
         return optimizer
 
-    def step(self, X, y, space_id, groups=None, n_jobs=1):
+    def _step(self, X, y, space_id, groups=None, n_jobs=1):
         """Generate n_jobs parameters and evaluate them in parallel.
 
         Having a separate function for a single step for search allows to
         save easily checkpoints for the parameter search and restore from
         possible failures.
```
betatim: should we remove this comment (L521, 522, 523) as well?

skopt/searchcv.py
```diff
@@ -673,35 +555,47 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
-                "calling fit method."
-            )
+        self.search_spaces_ = {}
+        for space, name in zip(search_spaces, list(range(len(search_spaces)))):
+            self.search_spaces_[name] = space
+
+        self.optimizer_kwargs_ = {} if self.optimizer_kwargs is None else\
+            self.optimizer_kwargs
+
+        # Instanciate optimizers for all the search spaces.
```
betatim: Instanciate -> Instantiate

skopt/searchcv.py
```diff
@@ -673,35 +555,47 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
-            raise ValueError(
-                "Please provide search space using `add_spaces` first before"
-                "calling fit method."
-            )
+        self.search_spaces_ = {}
+        for space, name in zip(search_spaces, list(range(len(search_spaces)))):
+            self.search_spaces_[name] = space
+
+        self.optimizer_kwargs_ = {} if self.optimizer_kwargs is None else\
+            self.optimizer_kwargs
```
betatim: I'd make this into:
```python
if self.optimizer_kwargs is None:
  self.optimizer_kwargs_ = {}
else:
  self.optimizer_kwargs_ = self.optimizer_kwargs
```

instead of trying to fit it on one line, then having to use `\` to continue on a second line :)

skopt/searchcv.py
```diff
@@ -673,35 +555,47 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
+        if isinstance(search_spaces, dict):
+            search_spaces = [search_spaces]
 
-        if len(self.search_spaces_) == 0:
-            raise ValueError(
-                "Please provide search space using `add_spaces` first before"
-                "calling fit method."
-            )
+        self.search_spaces_ = {}
+        for space, name in zip(search_spaces, list(range(len(search_spaces)))):
```
betatim: Would `for space_id, space in enumerate(search_spaces):` not also work?

Why do we use a dictionary instead of a list if we use the position as key? We should add a comment that reminds us about the reason or change to a `list`?

skopt/searchcv.py
```diff
@@ -673,35 +555,47 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
+            self.search_spaces_[name] = space
+
+        self.optimizer_kwargs_ = {} if self.optimizer_kwargs is None else\
+            self.optimizer_kwargs
+
+        # Instanciate optimizers for all the search spaces.
+        self.optimizer_ = {}
```
betatim: -> `self.optimizers_` ?

skopt/searchcv.py
```diff
@@ -722,10 +616,16 @@ def fit(self, X, y=None, groups=None):
             # do the optimization for particular search space
             while n_iter > 0:
                 # when n_iter < n_jobs points left for evaluation
-                n_jobs_adjusted = min(n_iter, self.n_jobs)
+                n_jobs_adjusted = min(n_iter, n_jobs)
```
betatim: we need to add something that protects us from doing too many calls towards the end: when `n_iter > 0 and n_iter < n_jobs_adjusted`. Say we have `n_iter = 3` and `n_jobs = 4` we would evaluate one model too many currently.

skopt/searchcv.py
```diff
@@ -722,10 +616,16 @@ def fit(self, X, y=None, groups=None):
...
 
-                self.step(
+                self._step(
                     X, y, space_id,
                     groups=groups, n_jobs=n_jobs_adjusted
                 )
                 n_iter -= n_jobs
+
+        # fit the best model if necessary
```
betatim: -> "Refit the best model on the the whole dataset" ?

betatim:
Should we add a test that checks that `BayesSearchCV` is compatible with the sklearn estimator semantics? So that this bug doesn't come back.

skopt/searchcv.py
fcharras: Ok, I think too that `_step` is mostly here for code clarity now. I don't know how things such as checkpoint would look, so removing.

skopt/searchcv.py
fcharras: The previous code was written with this dictionary, I don't know why. It seems right to use a list along with `enumerate`.

fcharras:
Thank you for the review. I'm currently addressing it.

About the test for sklearn estimator semantics, I was wondering about using http://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html , I need some more time to investigate because I don't know exactly what this helper checks.

Another simpler option could be to add tests using clone / set_params / get_params.

In fact I haven't done those tests yet.

fcharras:
I think this will address all the points you raised. Using lists to store search_space and optimizer make indeed the code a lot lighter. 

I provided a simple test for `clone()`, `set_params()` and `get_params()`, I'm not sure this is enough.

fcharras:
As for `check_estimator` at https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html it appears that GridSearchCV doesn't pass the test. BayesSearchCV fail with the same error: `AssertionError: "0 feature\(s\) \(shape=\(3, 0\)\) while a minimum of \d* is required." does not match "n_splits=3 cannot be greater than the number of members in each class."`. I don't think we can use this.

betatim:
A heads up on the CircleCI annoyance: #553 should fix that, once it is merged you can rebase your branch onto the latest master

fcharras:
I tried to use it on a real world problem, it seems to work just fine with sklearn, however there's an issue with multi-processing: even when n_jobs = 1, a step tries to use all the cores. Is it that the underlying optimizers spawn as much processes as possible ?

iaroslav-ai:
> even when n_jobs = 1, at some point it tries to use all the cores for a few seconds. Is it that the underlying optimizers spawn as much processes as possible ?

Yes, internally optimizer uses GaussianProcessRegressor by default. Inference with this estimator (used heavily during selection of next point to evaluate) runs on all cores if numpy by default runs on all cores (your's does that I assume). 

One can find [here](https://stackoverflow.com/questions/17053671/python-how-do-you-stop-numpy-from-multithreading) how to disable such behavior. Maybe would be good to turn this into an issue.

skopt/searchcv.py
```diff
@@ -673,59 +550,75 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
-            n_features is the number of features.
+        X : array-like or sparse matrix, shape = [n_samples, n_features]
+            The training input samples. Internally, it will be converted to
+            ``dtype=np.float32`` and if a sparse matrix is provided
+            to a sparse ``csc_matrix``.
 
         y : array-like, shape = [n_samples] or [n_samples, n_output]
-            Target relative to X for classification or regression;
+            Target relative to X for classification or regression (class
+            labels, as integers or strings).
```
iaroslav-ai: Maybe write here "class labels should be integers or strings", otherwise someone might that any targets should be represented as integers or strings

skopt/searchcv.py
```diff
@@ -673,59 +550,75 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
         # optimizer minimizes objective, hence provide negative score
         optimizer.tell(params, [-score for score in local_results])
 
-        # fit the best model if necessary
-        if self.refit:
```
iaroslav-ai: :+1:

skopt/tests/test_searchcv.py
```diff
@@ -42,9 +41,6 @@ def test_searchcv_runs(surrogate, n_jobs):
         X, y, train_size=0.75, random_state=0
     )
 
-    # None search space is only supported when only `step` function is used
-    assert_raises(ValueError, BayesSearchCV(SVC(), None).fit, (X, y))
```
iaroslav-ai: Maybe instead of this add test with zero length list?

skopt/searchcv.py
```diff
@@ -673,59 +550,75 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
         """Run fit on the estimator with randomly drawn parameters.
 
         Parameters
         ----------
-        X : array-like, shape = [n_samples, n_features]
-            Training vector, where n_samples in the number of samples and
-            n_features is the number of features.
+        X : array-like or sparse matrix, shape = [n_samples, n_features]
+            The training input samples. Internally, it will be converted to
+            ``dtype=np.float32`` and if a sparse matrix is provided
```
iaroslav-ai: Conversion to np.float32 is usually done with a particular estimator that is used with BayesSearchCV, but not BayesSearchCV itself. So one can have something like 
```python
import numpy as np
from sklearn.base import BaseEstimator, TransformerMixin
from sklearn.pipeline import make_pipeline
from sklearn.linear_model import Lasso

# strings as input
X = ['a'*np.random.randint(1,10) for i in range(100)]
# len of string should be estimated output
y = [len(x) for x in X]

# transformer on strings that converts them to features
class MyTransformer(BaseEstimator, TransformerMixin):
    def fit(self, X, y=None):
        return self

    def transform(self, X, y=None):
        X = [[len(x)] for x in X]
        X = np.array(X)
        return X

estimator = make_pipeline(
    MyTransformer(),
    Lasso()
)

model = BayesSearchCV(
    estimator=estimator,
    search_spaces={
        'lasso__alpha': (0.1, 10.0)
    },
    n_iter=8
)

model.fit(X, y)
print(model.score(X, y))
```
where inputs are strings. So consider deleting the comment about conversion of data.

fcharras:
Thanks for the review.

> runs on all cores if numpy by default runs on all cores (your's does that I assume).

I'm currently debugging it. It's indeed the `optimizer.tell` step that spawn that much threads. I'm pretty sure I have a standard python install / standard default ubuntu libraries for blas, mpi and such, I thought those wouldn't be parallelized by defaults ?

iaroslav-ai:
My reviews are nitpicks for most part :)
It is great that you removed `step` and other related stuff (I mean now code is ~100 lines shorter, just that is awesome). Originally the idea was to have ask-and-tell sorta interface for BayesSearchCV, but after all the feedback I do not find the idea so good (maybe I am just too much into ask - and - tell interface :D )

fcharras:
I've adressed your comments. 

The purpose of the PR is to fix the compatibility with sklearn and I think it is fulfilled now (I've tested it more extensively during the week with complex sklearn objects like nested Pipelines, calibrators,... no issue encountered). There might be other details that can be improved that remain from the last version (e.g documentation) that I missed. 

I'd like to continue polishing the class in new PRs. Some might demand more discussion or iterations (I'd like to reconsider the inheritance to BaseSearchCV, or at least how the _fit is defined, propose a change to the API and documentation so that `n_iter` becomes `n_iter / n_jobs`, set `pre_dispatch` to `n_jobs` and remove it from the `__init__` args, and consider using a single Parallel instance,...).

> After all the feedback I do not find the idea of ask and tell for *SearchCV so good, maybe I was just too much into it :D

As a heavy sklearn user, I was expecting that BayesSearchCV would _really_ be a drop-in replacement for GridSearchCV in all my pipelines. I think others will have the same expectations, and if the performance is good, this tool could be very popular once the PR is merged and the compatibility is flawless.

But the sklearn API and the compatibility requirements are somewhat inflexible and exclusive, so that comes at the cost of other possibly creative APIs like the ask-and-tell interface. In my opinion this is good too but it should be defined in a separate class.

So I think it's not about _not wanting_ ask-and-tell, but more about _really wanting_ sklearn compatibility, which is in my opinion one of the sanest framework.

glouppe:
Thanks a lot for your efforts! I am marking this for 0.5, as this is super important.

betatim:
Woah! I think I agree with most of what @fcharras wrote in https://github.com/scikit-optimize/scikit-optimize/pull/551#issuecomment-343646538

skopt/searchcv.py
```diff
@@ -44,7 +44,7 @@ class BayesSearchCV(BaseSearchCV):
         or ``scoring`` must be passed.
 
     search_spaces : dict, list of dict or list of tuple containing
-        (dict, int), or None.
+        (dict, int).
         One of 4 following cases:
```
betatim: -> "One of these cases:" (don't have four anymore and this way we won't have to update the text again when a new one is added ;))

skopt/searchcv.py
```diff
@@ -319,6 +291,11 @@ def _check_search_space(self, search_space):
         if isinstance(search_space, dict):
             search_space = [search_space]
 
+        if len(self.search_spaces) == 0:
```
betatim: In the rest of the method we use the `search_space` passed as argument. We should be consistent

skopt/searchcv.py
```diff
@@ -268,43 +266,17 @@ class BayesSearchCV(BaseSearchCV):
...
-        # instances for different search spaces. `str(space)` is used as key
-        # as `space` is a dict, which is unhashable. however, str(space)
-        # is fixed and unique if space is not changed.
-        self.optimizer_ = {}
-        self.cv_results_ = defaultdict(list)
-
-        self.best_index_ = None
-        self.multimetric_ = False
+        self.optimizer_kwargs = optimizer_kwargs
+        self._check_search_space(self.search_spaces)
```
betatim: What do you think of making this `self.search_spaces = self._check_search_space(search_spaces)` and removing L275?

In real scikit-learn style we should not check the arguments here but instead wait till `fit` is called :-/ Maybe that is actually the best thing to do. http://scikit-learn.org/stable/developers/contributing.html#instantiation WDYT?

skopt/searchcv.py
betatim: When does this happen? A user passes an empty list/dict?

skopt/searchcv.py
```diff
@@ -319,6 +291,11 @@ def _check_search_space(self, search_space):
         if isinstance(search_space, dict):
             search_space = [search_space]
 
+        if len(self.search_spaces) == 0:
+            raise ValueError(
+                "Please provide at least one search space"
```
betatim: -> `"The search_spaces parameter has to contain at least one search space not %s" % search_space` I find it helps the user (aka me debugging at 3am) a lot to have the parameter name and the current value of it in the message.

skopt/searchcv.py
```diff
@@ -673,59 +550,73 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
 
-        for space_id in sorted(self.search_spaces_.keys()):
-            elem = self.search_spaces_[space_id]
-
+        for search_space, optimizer in zip(search_spaces, optimizers):
             # if not provided with search subspace, n_iter is taken as
             # self.n_iter
-            if isinstance(elem, tuple):
-                space, n_iter = elem
+
```
betatim: no blank line here please

skopt/searchcv.py
```diff
@@ -673,59 +550,73 @@ def step(self, X, y, space_id, groups=None, n_jobs=1):
...
-        if len(self.search_spaces_) == 0:
-            raise ValueError(
-                "Please provide search space using `add_spaces` first before"
-                "calling fit method."
-            )
+        if self.optimizer_kwargs is None:
+            self.optimizer_kwargs_ = {}
+        else:
+            self.optimizer_kwargs_ = self.optimizer_kwargs
+
```
betatim: we need to call [`check_random_state`](http://scikit-learn.org/stable/developers/contributing.html#random-numbers) here and get random ints from it (`rng.randint(0, np.iinfo(np.int32).max)`) for each of our `Optimizer` instances. Otherwise the results won't be reproducible.

betatim: We should add a test doing a short optimisation with two different instances of `BayesSearchCV` to check things are indeed reproducible.

skopt/searchcv.py
fcharras: I guess it's OK to run code in the `__init__` that does not alter attributes, a similar check is performed in `GridSearchCV` https://github.com/scikit-learn/scikit-learn/blob/f3320a6f1ab2d060fd49c94a6e3386291a684ec7/sklearn/model_selection/_search.py#L1077

skopt/searchcv.py
fcharras: I'm changing it a little so that uses search_space, good catch. It will also be triggered both with an input empty dict or empty list. However it wouldn't trigger with a list of empty dicts. It looks like check_dimension (L333) would raise an error for any empty dict anyway.

This method  `check_search_space` already looks very similar to GridSearchCV's check_param_grid https://github.com/scikit-learn/scikit-learn/blob/f3320a6f1ab2d060fd49c94a6e3386291a684ec7/sklearn/model_selection/_search.py#L343 .

skopt/searchcv.py
fcharras: thank you for providing the code 😸

fcharras:
Thank you for the review, those commits should address it

fcharras:
I've been thinking about the test on reproducibility, using a SVM as a base estimator might not make the safest test because of its stability regardless of the parameters. I've changed it to a random forest, but now the test passes... only half the time.

fcharras:
Added a better test for the reproducibility. The test is definitely not passing, results are not reproducible... isn't https://github.com/scikit-optimize/scikit-optimize/pull/551/files#diff-5be74bdba5232db1bd3137d5827653f6R583 correct though ?

betatim:
To me `test_searchcv_reproducibility` looks like it should work and test what we want it to test. So if it reliability passes that is good ;)

skopt/tests/test_searchcv.py
```diff
@@ -124,3 +124,102 @@ def test_searchcv_runs_multiple_subspaces():
...
+    )
+
+    opt.fit(X_train, y_train)
+    best_estimator = opt.best_estimator_
+
+    opt2 = clone(opt).fit(X_train, y_train)
+    best_estimator2 = opt2.best_estimator_
+
+    for attr in ('C', 'gamma', 'degree', 'kernel'):
+        assert_equal(getattr(best_estimator, attr),
```
betatim: with pytest you can do `assert getattr(best_estimator, attr) == getattr(best_estimator2, attr)`. I would prefer explicitly writing out the four lines instead of having a for loop. In the interest of keeping our tests as simple as possible in terms of complexity and it makes it a bit simpler to see in the output which parameter is different.

betatim:
I think this looks pretty good. Maybe change the for loop in the test to satisfy my nitpicking.

skopt/tests/test_searchcv.py
```diff
@@ -124,3 +124,102 @@ def test_searchcv_runs_multiple_subspaces():
...
+            'degree': Integer(1, 8),
+            'kernel': Categorical(['linear', 'poly', 'rbf']),
+        },
+        n_iter=11, random_state=random_state
+    )
+
+    opt.fit(X_train, y_train)
+    best_estimator = opt.best_estimator_
+
+    opt2 = clone(opt).fit(X_train, y_train)
```
iaroslav-ai: I wonder if there might be a situation where the test will pass because the estimator being cloned was already fitted? That is, when the estimator is cloned before calling `fit` method, the test would fail. From looking at definition of clone method, I am not sure. If you think this might happen, consider adding another copy before fitting, and comparing it with the rest.

skopt/searchcv.py
```diff
@@ -268,43 +268,17 @@ class BayesSearchCV(BaseSearchCV):
 
     """
 
-    def __init__(self, estimator, search_spaces=None, optimizer_kwargs=None,
+    def __init__(self, estimator, search_spaces, optimizer_kwargs=None,
```
iaroslav-ai: Something that is probably out of scope for this PR: I realize now is that `optimizer_kwargs` are the same for all subspaces. However, one might want these to be different for different subspaces. An example would be that I want optimizer to try a few initial points, that I suspect already to be good for some subspace. This could be done by specifying kwargs for optimizer; However, in case of multiple subspaces this is not possible as kwargs for those are the same. Similarly, one might want to use decision trees based optimizers for subspaces with large number of iterations and models with fast fitting. 
So kwargs could be specified per subspace, like iterations number. For now one can simply have a separate BayesSearchCv per subspace. 
Edit: maybe lets leave this out, in order to avoid everlasting PR's :)

glouppe:
I had several quick looks over the past two weeeks. This now seems like a significant improvement over master. All the little things can be addressed later in smaller PR in my opinion.

+0.5 for merge on my side (since I have not reviewed the PR in full details)

iaroslav-ai:
Yes, I am fine also with merging this as it is.

fcharras:
Thank you! I think if the test on reproducibility passes here, then it passes consistently and there were some issue with my setup on my side that made me think it does not pass.

I can check again by tonight and maybe you could try it too to be sure. About the other little improvements (i have other in mind too) I could write some documentation that list them that could be addressed in further PR.

skopt/tests/test_searchcv.py
fcharras: I don't think this is an issue, I think clone always return a new un-fitted copy of the estimator, regardless of wether the input estimator is fitted or not. I'll double check it.

fcharras: Indeed, you can check https://github.com/scikit-learn/scikit-learn/blob/19eb62ff/sklearn/base.py#L30 and https://github.com/scikit-learn/scikit-learn/blob/19eb62ff/sklearn/base.py#L198 : `clone` return a fresh instance of the same class, and get the parameters from the signature of the `__init__` of the cloned instance.

skopt/tests/test_searchcv.py
fcharras: I changed to assert, but using `assert_equal` seems nicer because it also print the values that should have been equal (ex: `43.5 = 1003.32`)

skopt/searchcv.py
fcharras: I think this could be done in further PR too, I'd like to propose some documentation with further possible improvements.

fcharras:
I checked again, the last test is definitely not passing on my side, but definitely passing on the integrated tests here... On my side I ran it inside an ipython terminal in spyder. It's failing 100% of the time.

skopt/searchcv.py
betatim: @iaroslav-ai can you make an issue for this so we don't lose the idea?

betatim:
Tests pass on travis and on my laptop -> merging!

Thanks a lot for these improvements, as well as your patience for my never ending comments :)

fcharras:
Thank you!!

betatim:
@fcharras could you send me an email (<mygithubusername>@google.com)? We'd like to get in touch with you and our internet stalking skills aren't good enough to find your email address :)

PR #553, opened 2017-11-04
Fix install instructions and add conda-forge

betatim:
We shouldn't have removed the instruction to install `numpy` first. That has to wait till after we release the next version.

Updated the link to the next milestone, and did a small reformatting of the pytest instructions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,18 @@ The latest released version of scikit-optimize is v0.4, which you can install
 with:
 ::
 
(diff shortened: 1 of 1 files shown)


betatim:
I think this is uncontroversial, if someone could take a quick look then we could merge this so that CircleCI stops timing out.

betatim:
The 🚓 want to talk to you about breaking speed limits :))

Issue #559, opened 2017-11-19
Noise doesn't seem to be used in Gaussian Process Regressor?


SoManyUsernamesTaken:
I could be wrong, but I was browsing through the source code and as far as I can tell, noise is never used in Gaussian Process Regressors unless the kernel is set by the user. Specifically, in the code for `GaussianProcessRegressor`:
```
        if self.kernel is None:
            self.kernel = ConstantKernel(1.0, constant_value_bounds="fixed") \
                          * RBF(1.0, length_scale_bounds="fixed")
        elif self.noise == "gaussian":
            self.kernel = self.kernel + WhiteKernel()
        elif self.noise:
            self.kernel = self.kernel + WhiteKernel(
                noise_level=self.noise, noise_level_bounds="fixed"
            )
```

This means that if the kernel is `None` (which is the default behaviour) then the noise argument is ignored.

This occurs here:
https://github.com/scikit-optimize/scikit-optimize/blob/be5f87aab8d61d8bee123340715716bfd9506a3c/skopt/learning/gaussian_process/gpr.py#L186

To fix this, I believe the code should be modified to:

```
        if self.kernel is None:
            self.kernel = ConstantKernel(1.0, constant_value_bounds="fixed") \
                          * RBF(1.0, length_scale_bounds="fixed")
        if self.noise == "gaussian":
            self.kernel = self.kernel + WhiteKernel()
        elif self.noise:
            self.kernel = self.kernel + WhiteKernel(
                noise_level=self.noise, noise_level_bounds="fixed"
            )
```

SoManyUsernamesTaken:
To further investigate, I tried the following code. All 

```
import numpy as np
from skopt.learning import GaussianProcessRegressor

X = np.random.normal(size=[100,2])
Y = np.random.normal(size=[100])

g_none = GaussianProcessRegressor(noise=None) # default
g_none.fit(X,Y)
print(g_none.kernel)
print(g_none.predict(X[0:1], return_cov = True))

g_gaussian = GaussianProcessRegressor(noise='gaussian')
g_gaussian.fit(X,Y)
print(g_gaussian.kernel)
print(g_gaussian.predict(X[0:1], return_cov = True))

g_const = GaussianProcessRegressor(noise=1)
g_const.fit(X,Y)
print(g_const.kernel)
print(g_const.predict(X[0:1], return_cov = True))
```

All use the same kernel, and all return the same predictions with 0 (!) predicted covariance

[...]

PR #562, opened 2017-11-22
[MRG+1] Modified gpr so that kernel uses noise

SoManyUsernamesTaken:
See https://github.com/scikit-optimize/scikit-optimize/issues/559

### DIFF
--- a/skopt/learning/gaussian_process/gpr.py
+++ b/skopt/learning/gaussian_process/gpr.py
@@ -186,7 +186,7 @@ class GaussianProcessRegressor(sk_GaussianProcessRegressor):
         if self.kernel is None:
             self.kernel = ConstantKernel(1.0, constant_value_bounds="fixed") \
                           * RBF(1.0, length_scale_bounds="fixed")
-        elif self.noise == "gaussian":
+        if self.noise == "gaussian":
             self.kernel = self.kernel + WhiteKernel()
         elif self.noise:
             self.kernel = self.kernel + WhiteKernel(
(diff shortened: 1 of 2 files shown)


skopt/learning/gaussian_process/gpr.py
```diff
@@ -186,7 +186,7 @@ def fit(self, X, y):
         if self.kernel is None:
             self.kernel = ConstantKernel(1.0, constant_value_bounds="fixed") \
                           * RBF(1.0, length_scale_bounds="fixed")
-        elif self.noise == "gaussian":
+        if self.noise == "gaussian":
             self.kernel = self.kernel + WhiteKernel()
```
betatim: is there a way to check if `self.kernel` already contains a `WhiteKernel`? Or does someone know what happens when you add two `WhiteKernel`s together?

betatim:
Thanks for this PR. Could you also add a small test so that this bug doesn't come back?

skopt/learning/gaussian_process/gpr.py
SoManyUsernamesTaken: Not easily anyway. However not much will happen - it will just make optimisation a bit slower as there is one extra useless hyperparameter to perform gradient descent over, but you will still get the same minimum

SoManyUsernamesTaken:
Added test based on example code from the issue

betatim:
Looks good to me! Will hit merge in a few hours, just giving others a chance to look/comment.

SoManyUsernamesTaken:
By the way, should it be default behaviour to assume Gaussian noise? GPR's should be able to discover the amount of noise automatically in most situations

betatim:
I don't remember why that choice was made. It would make sense to allow noise by default given we focus on noisy black box functions. Maybe @MechCoder or @glouppe remember? We should do that in a new PR though.

betatim:
Merged! Thanks for spotting this and fixing it.

PR #564, opened 2017-11-27
Change so as to use gaussian noise by default

SoManyUsernamesTaken:
Also moved test from the test_gpr I created into the test_gpr that was pre-existing under learning/Gaussian_process/tests/test_gpr.py and changed it so that it also tests for noise by default.


SoManyUsernamesTaken:
It seems that on the contrary there are existing tests in place to ensure that the default behaviour is to make sure that noise is not included by default. I leave this up to you.

betatim:
Any thoughts @MechCoder and @glouppe on changing this default?

We'd have to change the tests that check that by default no noise is included. They are doing their job of detecting that we are changing the defaults ;)

iaroslav-ai:
Now we already have [noise="gaussian" by default](https://github.com/scikit-optimize/scikit-optimize/blob/691959d95b514dc511e0281c0d684ed5fba7f630/skopt/optimizer/gp.py#L17). Feel free to reopen this if there is anything missing.

PR #565, opened 2017-11-28
[MRG] Add link to mybinder.org

betatim:
The `postBuild` script get's executed at the end and install `skopt`.


glouppe:
Looks good! merging

Issue #560, opened 2017-11-19
Compatibility of the ask and tell API with acquisition function EIps


chschroeder:
It seems i cannot use acquisition functions that incorporate the runtime (EIps, PIps) when using the ask and tell API.

My objective function returns a tuple of (result, time), this part should be correct imho. The error is thrown during the lying step of the optimizer.

I get the following error:
```
[shortened]
File "/usr/local/lib/python3.6/site-packages/skopt/optimizer/optimizer.py", line 339, in ask
    opt.tell(x, y_lie)  # lie to the optimizer
  File "/usr/local/lib/python3.6/site-packages/skopt/optimizer/optimizer.py", line 417, in tell
    raise TypeError("expected y to be (func_val, t)")
TypeError: expected y to be (func_val, t)
```

Are the ask and tell api and EIps not supposed to work together, or am i missing something?
(Version: v0.4)

betatim:
Can't think of a reason why they should not work together, so this is most likely a bug. Would you be interested in proposing a fix? (We can work together to find the right thing to do if you want to).

chschroeder:
Sure. Howevery, after constructing a minimal reproducible example i think the error must be on my side.

```
"""
Based on examples/ask-and-tell.ipynb
"""
import numpy as np
from skopt import Optimizer

def objective(x):
    seconds = 1
    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2)), seconds

opt = Optimizer([(-2.0, 2.0)], 'ET', acq_func='EIps', acq_optimizer='sampling')

for i in range(9):
    next_x = opt.ask()
    f_val = objective(next_x)
    opt.tell(next_x, f_val)
```

Thanks for your fast response anyway! I will recheck the rest of my code and update this issue if the problem persists.

chschroeder:
I re-examined, everything is working fine :) The mistake was in another part of my code.

chschroeder:
Closed this too soon. If i pass `n_points=5` to the ask method i can reproduce the error.

```
"""
Minimal reproducible example for #560
(https://github.com/scikit-optimize/scikit-optimize/issues/560)
"""
import numpy as np

from skopt import Optimizer

def objective(x):
    seconds = 1
    return np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2)), seconds

opt = Optimizer([(-2.0, 2.0)], 'ET', acq_func='EIps', acq_optimizer='sampling')

[...]

PR #572, opened 2017-12-13
[MRG] Prevent the base_estimator from being wrapped multiple times

chschroeder:
The previous version used to create copies of the optimizer, which looked like this:

```
MultiOutputRegressor(estimator=MultiOutputRegressor(estimator=GaussianProcessRegressor(alpha=1e-10, copy_X_train=True,
             kernel=1**2 * Matern(length_scale=1, nu=2.5),
             n_restarts_optimizer=2, noise='gaussian', normalize_y=True,
             optimizer='fmin_l_bfgs_b', random_state=2061968400),
           n_jobs=1),
           n_jobs=1)
```
The estimator inside `MultiOutputRegressor` is again a `MultiOutputRegressor`, which is automatically added for acquisition functions that incorporate time (EIps, PIps).

Basically, each call to copy() would wrap another `MultiOutputRegressor` around the base_estimator.
This affects at least #560.

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -213,7 +213,8 @@ class Optimizer(object):
             raise ValueError(
                 "%s has to be a regressor." % base_estimator)
 
-        if "ps" in self.acq_func:
+        is_multi_regressor = isinstance(base_estimator, MultiOutputRegressor)
+        if "ps" in self.acq_func and not is_multi_regressor:
             self.base_estimator_ = MultiOutputRegressor(base_estimator)
         else:
             self.base_estimator_ = base_estimator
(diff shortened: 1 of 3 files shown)


betatim:
Thank you for this, especially for breaking it out into a separate PR. Really appreciate having small chunks to look at.

I think overall this should be uncontroversial as it fixes a clear bug and comes with a test 🎉 !

Thinking out loud a bit: this happens to us because we perform work in `__init__` instead of leaving it to later. In scikit-learn not doing anything in `__init__` is one of the few golden rules. It is a bit harder in `Optimizer` because in principle you can call `tell` or `ask` first, so we'd have to do the initialisation in both? And unlike `fit` on a sklearn estimator you are meant to call `ask`/`tell` multiple times without resetting the state. So right now we will keep having bugs like this for things that need initialising as we need to remember to "get it right" each time. Not sure I have a good answer right now. Any ideas? @glouppe?

If no one has an idea I propose to merge this later today.

chschroeder:
Thank you for the fast response and your feedback!

And i agree with your thoughts about this. I was having a similar thought, but then i decided to keep the changes minimal. If anyone comes up with a better solution, we have at least a test now for the refactoring.

betatim:
Merging for now. We can always improve later.

PR #574, opened 2017-12-13
[MRG] Ask and tell for runtime optimization

chschroeder:
**Update:**

*Description:*
This PR aims to solve the issue discussed in #560. Basically, it should fix the ask and tell API for the cases, where an acquisiton function that incorporates time is used.

*Status:*
- [x] (Unit) Tests
- [x] Code changes
- [x] At least some manual test runs
- ~~Adapt the comments~~ (--> separate task)

*Changes:*
- Most of the work has been done in #572 and #593 
- This issue changes the lying step so that it also generates a time value when lying.
- The time value is generated in a similar way as the lie for objective value. However, when no previous evaluations exist, then the maximum possible runtime time `log(float.max)` is used.

***

Current state:
* I provided (or rather extended) a test which demonstrates the failure

Problems:
* During the lying step the time values are not considered at all (correct me if i am wrong)
* Within the tell step they are somewhat considered, but the result creation fails.

~~Additional Suggestion: The result object built by `create_result` (skopt.utils) should probably contain information about the runtimes as well.~~ (Update: This is already present. Maybe we should explicitly state this in the docstrings.)

(Sorry for the ugly commit history, i mixed up the branches.)

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -339,17 +339,16 @@ class Optimizer(object):
 
             if strategy == "cl_min":
                 y_lie = np.min(opt.yi) if opt.yi else 0.0  # CL-min lie
-                t_lie = np.min(ti) if ti else sys.float_info.max
+                t_lie = np.min(ti) if ti is not None else log(sys.float_info.max)
             elif strategy == "cl_mean":
                 y_lie = np.mean(opt.yi) if opt.yi else 0.0  # CL-mean lie
-                t_lie = np.mean(ti) if ti else sys.float_info.max
+                t_lie = np.mean(ti) if ti is not None else log(sys.float_info.max)
             else:
                 y_lie = np.max(opt.yi) if opt.yi else 0.0  # CL-max lie
-                t_lie = np.max(ti) if ti else sys.float_info.max
+                t_lie = np.max(ti) if ti is not None else log(sys.float_info.max)
 
             # Lie to the optimizer.
             if "ps" in self.acq_func:
-
                 # Use `_tell()` instead of `tell()` to prevent repeated
                 # log transformations of the computation times.
                 opt._tell(x, (y_lie, t_lie))
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -335,7 +335,8 @@ class Optimizer(object):
             x = opt.ask()
             X.append(x)
 
-            ti = [t for (_, t) in opt.yi] if "ps" in self.acq_func else None
+            ti_available = "ps" in self.acq_func and len(opt.yi) > 0
+            ti = [t for (_, t) in opt.yi] if ti_available else None
 
             if strategy == "cl_min":
                 y_lie = np.min(opt.yi) if opt.yi else 0.0  # CL-min lie
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -1,3 +1,4 @@
+import sys
 import warnings
 from math import log
 from numbers import Number
@@ -333,13 +334,27 @@ class Optimizer(object):
         for i in range(n_points):
             x = opt.ask()
             X.append(x)
+
+            ti = [t for (_, t) in opt.yi] if "ps" in self.acq_func else None
+
             if strategy == "cl_min":
                 y_lie = np.min(opt.yi) if opt.yi else 0.0  # CL-min lie
+                t_lie = np.min(ti) if ti else sys.float_info.max
             elif strategy == "cl_mean":
                 y_lie = np.mean(opt.yi) if opt.yi else 0.0  # CL-mean lie
+                t_lie = np.mean(ti) if ti else sys.float_info.max
             else:
                 y_lie = np.max(opt.yi) if opt.yi else 0.0  # CL-max lie
-            opt.tell(x, y_lie)  # lie to the optimizer
+                t_lie = np.max(ti) if ti else sys.float_info.max
+
+            # Lie to the optimizer.
+            if "ps" in self.acq_func:
+
+                # Use `_tell()` instead of `tell()` to prevent repeated
+                # log transformations of the computation times.
+                opt._tell(x, (y_lie, t_lie))
+            else:
+                opt._tell(x, y_lie)
 
         self.cache_ = {(n_points, strategy): X}  # cache_ the result
 
--- a/skopt/tests/test_parallel_cl.py
+++ b/skopt/tests/test_parallel_cl.py
@@ -50,6 +50,7 @@ def test_constant_liar_runs(strategy, surrogate, acq_func):
     optimizer = Optimizer(
         base_estimator=surrogate(),
         dimensions=[Real(-5.0, 10.0), Real(0.0, 15.0)],
+        acq_func=acq_func,
         acq_optimizer='sampling',
         random_state=0
     )
--- a/skopt/tests/test_parallel_cl.py
+++ b/skopt/tests/test_parallel_cl.py
@@ -16,6 +16,10 @@ import pytest
 
 # list of all strategies for parallelization
 supported_strategies = ["cl_min", "cl_mean", "cl_max"]
+
+# test one acq function that incorporates the runtime, and one that does not
+supported_acq_functions = ["EI", "EIps"]
+
 # Extract available surrogates, so that new ones are used automatically
 available_surrogates = [
     getattr(sol, name) for name in sol.__all__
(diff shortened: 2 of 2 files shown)


betatim:
(commit history: if you feel like upping your git-magic level, `git rebase -i <select commits here>` is quite powerful for doctoring your history)

betatim:
@chschroeder is this ready for review or do you plan on working on the `create_result` bit in this PR?

chschroeder:
This is not finished yet. The checks should not be green either :sweat:. I will try to keep the changes as small as possible, however, at first glance it looked like i need to get most of it working in order to pass the test (except for the create_result extension maybe).

chschroeder:
Seems i messed up the rebase. Fixing this asap. (This was automatically closed somehow by my push).

betatim:
Looks like you are back on track :) CircleCI mostly just generates the webpage so the fact that it is green means our examples and docs work, not that your test isn't leading to a failure (as you'd expect). For the tests travis is the one that counts.

chschroeder:
I have some time planned for this during the next week. Will keep you updated :)

chschroeder:
The Problem seems to be that there are different understandings of how to handle the time values in different parts of the code in `skopt/optimizer/optimizer.py`. Mostly, the objective values and their corresponding time values are packed as tuples and unpacked again whenever needed. This was probably done to keep the interface as it is. However i feel both from the code and from a usability perspective we should handle the time values separately because as it is, it does not match the docs either (yi is a list of scalars).

For now, i will try to do the non API-breaking fix but i would like to propose to think about the interface here.

chschroeder:
The tests are passing now on my machine. I want to do at least one manual test as well.

Code-wise i need to clean up and comment. Maybe even some refactoring; The lying strategies could be handled more elegantly. Not sure if now is the right time for this.

Off topic:
I pulled the changes from #593. Is this the right way to do or should i wait until it is merged?

betatim:
I'd generally try and avoid merging other branches into a PR. It creates a confusing history when you look back at it. One thing that works well is to branch off of the branch for the PR you depend on. Then once that "base" PR is merged you rebase this branch onto master. We merge this PR.

We can rewrite the history of this branch at the end with the magic of `git rebase -i`. So for the moment don't worry. 

We should merge #593 before this then.

chschroeder:
Thanks, next time i will know how to handle multiple PRs better. At which time do we perform the `git rebase -i`? I tried this locally out of curiosity and was rather unsuccesful :sweat_smile:.

Update on the issue:
The first manual test was fine (= no errors or unexpectedly long runtimes encountered).

Next i will try the same setting with and without time optimization (e.g. 'EI'/'EIps') and compare the suggested points (they should differ). If you have any suggestions how to test for logic flaws here let me know.

betatim:
Maybe `git rebase -i` is already too tricky. This is my attempt at a short guide to rebasing. [This](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) is a longer guide, but it also makes assumptions specific to EdX. It is the guide I keep under my pillow to refer to when I am doing this.

We want to take this branch and move where it branches off of `master`, so that it includes the changes made in master in the mean time. This is called "rebase"ing in git jargon.

Step 0: make a copy of your directory so if things go wrong there is a backup.

Prep work:
1. `git checkout master`
1. `git fetch upstream` and `git merge upstream/master`

This should just work. It assumes you have a remote called "upstream" which points to https://github.com/scikit-optimize/scikit-optimize

Now checkout your branch. To rebase it `git rebase master`. This will either just work or will stop when there are merge conflicts (bits of code that are changed in your branch and master). When this happens sort out the bits of code between a big bunch of "<<<<<<<<<<". Then you need to tell git that you resolved the conflict with `git add <file-that-had-a-conflict>`, once all are resolved `git rebase --continue` to keep going. Eventually you will get to the end. And have successfully rebased. If you now run `git push` you will get a warning, and the only way to fix it is to use `git push -f` (push with force). This is because you rewrote history, which could get us all into hot water. Hence the need to use `-f`.

Let me know how it goes.

ps. if someone knows a good guide to rebasing that follows what we do, links welcome!
pps. Others find rebasing to complicated and prefer merging master into your PR branch. I dislike it but more for aesthetic reasons than anything based in hard facts.

chschroeder:
Thanks a lot for your excellent instructions @betatim! I got the desired result on the first try.

chschroeder:
OK, i managed to setup an optimization task (from a project a mine) with a parameter space that is somewhat restricted, but is still realistic. In a word, it optimized the parameters for a classification task performed by lightgbm.

Two configurations were tested using the ask and tell API for 100 iterations. For the parameters `n_points` was set to `2`, `n_initial_points` was set to `10`.
The `acq_func` was set to `EI` for the first configuration and to `EIps` for the second.

I observed the error (1 - accuracy) and the computation time (in seconds). Below are the stats for the resulting times (using `pandas.DataFrame.describe()`)

**EI**
```
count    100.000000
mean     184.681320
std      148.959615
min        1.353564
25%       51.510879
50%      154.634320
75%      303.293788
max      556.699259
Name: seconds, dtype: float64
```

**EIps**
```
count    100.000000
mean     120.093131
std      151.183997
min        1.440042
25%       17.541990
50%       64.032605
75%      181.043889
max      783.810727
Name: seconds, dtype: float64
```

* No real experimental settings unfortunately, i was working at the same time (without straining my machine too much ;) )

Still, i think for a rough test, we can still draw some conclusions:
* No errors anymore when using a "ps" acq_func :tada:
* EIps selects points which lead to shorter runtimes on average.
* Different points are evaluated and the best parameter configurations yielding roughly the same score. In other words, we see (as expected) a different behaviour AND the results do not get worse for this example when including the runtime in the optimization. (Cannot be seen here, but i kept the csv files if anyone is interested.)

skopt/optimizer/optimizer.py
```diff
@@ -333,13 +334,27 @@ def ask(self, n_points=None, strategy="cl_min"):
...
                 y_lie = np.mean(opt.yi) if opt.yi else 0.0  # CL-mean lie
+                t_lie = np.mean(ti) if ti else sys.float_info.max
             else:
                 y_lie = np.max(opt.yi) if opt.yi else 0.0  # CL-max lie
-            opt.tell(x, y_lie)  # lie to the optimizer
+                t_lie = np.max(ti) if ti else sys.float_info.max
+
+            # Lie to the optimizer.
+            if "ps" in self.acq_func:
+
```
betatim: Spurious empty line

skopt/optimizer/optimizer.py
```diff
@@ -333,13 +334,27 @@ def ask(self, n_points=None, strategy="cl_min"):
         for i in range(n_points):
             x = opt.ask()
             X.append(x)
+
+            ti = [t for (_, t) in opt.yi] if "ps" in self.acq_func else None
+
             if strategy == "cl_min":
                 y_lie = np.min(opt.yi) if opt.yi else 0.0  # CL-min lie
+                t_lie = np.min(ti) if ti else sys.float_info.max
```
betatim: Nitpick: in python `bool([]) == False` as well as `bool(None) == False` so the recommendation is to test `if ti is not None` if that is what you mean, instead of `if ti`. In this case we might end up with a hard to spot bug if `ti = []` for some crazy reason. (same for the below)

betatim:
Ready to merge?

skopt/optimizer/optimizer.py
chschroeder: Not a nitpick, that's a valid issue. Thanks for the correction, always happy to learn more. I've developed most of my code in Java, so I still have to learn a lot of the python specifics.

chschroeder: Another issue: `sys.float_info.max` should be `log(sys.float_info.max)`. If we did not use `_tell()` but `tell()`, the value would be log transformed. Will probably not make a difference in practice, but i feel this closer to the rest of the idea.

iaroslav-ai:
LGTM!

skopt/optimizer/optimizer.py
iaroslav-ai: Probably also less numerical issues with surrogate which estimates time

chschroeder:
Thanks for the review! I added something to the review as well: `log(sys.float_info.max)`. Please check that and tell me if you agree. Other than that, this is ready to merge :smile:

iaroslav-ai:
I think using `log(sys.float_info.max)` is a good idea - I would also expect that any constant that does not break the log operation would be fine. 

I think this will be merged tomorrow, together with the release of 0.5 :tada: and zenodo entry to finally have something to cite more than github :D.

iaroslav-ai:
@chschroeder thanks for the PR! With some recent experiments, I realized that having things running fast is a quite important feature :)

betatim:
Travis does not like the most recent changes :-/

chschroeder:
My bad, even for simple changes the tests have to run. I should know better :frowning:.Give me 20mins, the tests have already started. 

The issue is exactly the `ti = []` case you mentioned in your review. I would have gone back to checking using `if ti`, however, it makes no sense that `t_lie` is given a value for `acq_func`s without "ps" (even when it is unused there). I have a different solution now; will be pushed in some minutes.

betatim:
Merging! Thanks for fixing this bug.

glouppe:
Thanks @chschroeder for your contribution :)

PR #588, opened 2017-12-24
refactored out `fit` of `Optimizer` and added `mask` to the new `fit` method

jonathanng:


betatim:
Could you describe a bit why you think fixing values will help with dealing with subspaces? We should leave some hints for people from the future who are trying to understand things. If you know of a paper or some such to read I'd be interested.

skopt/optimizer/optimizer.py
```diff
@@ -528,3 +460,95 @@ def run(self, func, n_iter=1):
 
         return create_result(self.Xi, self.yi, self.space, self.rng,
                              models=self.models)
+
+    def fit(self, mask=None):
```
betatim: should we make this private? `_fit()` it isn't meant to be called by anyone right?

skopt/optimizer/optimizer.py
```diff
@@ -528,3 +460,95 @@ def run(self, func, n_iter=1):
...
+        """Fit a model to observed evaluations of the objective.
+
+        Parameters
+        ----------
+        * `mask` [None or list]
```
betatim: For me `mask` and arrays suggests masking out elements in an array, not really fixing them to a certain value. I will try and make a suggestion for a name I think would be a better fit. If anyone else has ideas post them :)

betatim:
Also: thanks for (first time) contributing! :)

skopt/optimizer/optimizer.py
jonathanng: That's a good point. Yes, mask is perhaps not the best name.

skopt/optimizer/optimizer.py
jonathanng: Wouldn't it make more sense to have `ask` call `fit`? Then we can make `mask` a parameter of `ask`.

betatim: Not sure I understand your comment :-/ `ask` should indeed call `fit` but I think it should be a private method, not something users of `Optimizer` have access to.

jonathanng: Take a look at latest commit, and let me know what you think.

iaroslav-ai:
Hey @jonathanng ,
thanks for your PR! Could you elaborate when would you need to fix dimensions, esp at runtime? 
If you wish to fix a dimension in subspace (eg to specify an estimator you use for some subspace in BayesSearchCV) you can use Categorical dimension. I actually have a pr that simplifies that a bit more here #515

jonathanng:
Here is one example:

Let's say my experiments are dependent on many dimensions including time of the day.

Let's also say it's 10 am, and I want to continue searching the space. I don't want the optimizer to tell me to do a 3 pm experiment because that's a waste of time. So fixing the time dimension at 10 am would allow the optimizer to give me points that are good search points.

I can think of other situations. Let me know if you want to hear more.

iaroslav-ai:
I see - sounds interesting

chschroeder:
I like the idea. Currently i use xgboost / lightgbm a lot on some big datasets. And parameter space for these implementations is large (read: larger than my i am willing to wait).

Based on this masking functionality one could implement a stagewise optimization.

betatim:
I'm not sure I understand the 10am use-case :-/ Below an example that I think is equivalent, let me know if it is (if yes I do understand the use case 😄):

You have a search space with three dimensions: A, B, and C. For some reason at the start you want to only explore A and B with a fixed C=10. At some later stage you might want to "unfix" C and search over the whole space.

Assuming I understood correctly: is there an advantage to search over A and B with C=10 and then unfixing C after a while compared to first running an `Optimizer` over a space with just A and B, then when you want to bring C into the game you create a new `Optimizer` instance with a search space over A, B and C, feeding in the values from the previous optimizer (over A and B only) as initial values (`x0` and `y0` arguments).

Trying to get the idea straight in my head, once that is done going to think about what is nicer for the user/less typing etc :)

skopt/optimizer/optimizer.py
```diff
@@ -393,11 +403,6 @@ def tell(self, x, y, fit=True):
 
         * `y` [scalar or list]:
             Value of objective at `x`.
-
-        * `fit` [bool, default=True]
```
betatim: we need to keep the `fit` argument with a deprecation warning for two releases. I would also like to keep it because I do use it as a way of adding observations to the optimizer when I know I have a batch of them and don't want to re-run the fitting because I will add several before calling `ask` again.

jonathanng: All the fit logic is moved to ask, so the fit argument for tell won't do anything.

betatim: Ah sorry. I thought I posted a comment (somewhere :-/) about the reasoning why the work of `fit` is done in `tell`.

TL;DR: we should keep it in `tell`. My reasoning is the following: the cycle of ask-and-tell is always: ask, evaluate expensive function, tell, check if we are done, if not ask, eval, tell, check if done, ...

For example when I `tell` a point I just evaluated I expect the surrogate I get from the optimizer to include information from that point so that I can decide if I need to ask for another point or if we are done or what else to do. Similarly if we were to plot the surrogate model after one round (ask, eval, tell) we'd expect it to be fitted to that one point. It feels a bit weird to me having to do: ask, eval, tell, ask, now plot and decide what we should do next.

What do you think?

Issue #570, opened 2017-12-12
How to name parameter dimensions?


Hvass-Labs:
Hello, I applaud your effort with this library and I'm sorry to bother you with such a simple question.

I would like to use skopt with Keras and TensorFlow for optimization of hyper-parameters of a Convolutional Neural Network.

I have installed skopt v.0.4 using pip.

Ideally I would like to name the hyper-parameters only once when I define the search-space, and when skopt calls my "fitness" function, it gives named arguments so I don't have to unpack a list of parameters, which is error-prone.

It appears that `BayesSearchCV` can take a dict of named parameter-dimensions. But I don't think I can use `BayesSearchCV` because I need to run `fit_generator()` from Keras instead of `fit()`, because I want to use larger datasets from the harddisk that don't fit in memory.

So I think I have to use `gp_minimize()` but this does not take a dict of named parameters, instead it takes a list of dimensions, which can be defined in various ways.

I tried doing the following to name the parameters:

    from skopt.space import Real, Categorical, Integer

    dimensions = [Real(name='learning_rate', low=1e-6, high=1e-3, prior='log-uniform'),
	          Integer(name='num_nodes', low=10, high=256),
	          Categorical(name='activation', categories=['relu', 'sigmoid'])]

And then I would call:

    from skopt import gp_minimize

    res = gp_minimize(func=f, dimensions=dimensions)

The docs say that `Integer`, `Real` and `Categorical` can take a `name` argument:

https://scikit-optimize.github.io/space/space.m.html#skopt.space.space.Integer

But I get the following error:

    TypeError: __init__() got an unexpected keyword argument 'name'

Going into the source-code for `space.py` we see that `name` is indeed missing:

    class Real(Dimension):
        def __init__(self, low, high, prior="uniform", transform=None):

Is this because I have installed an older version of skopt using pip?

Could you advise how to construct named dimensions in the search-space. Ideally I would only have to name each dimension once when I define it, and then in my fitness function  I would like to get named arguments somehow, so I don't need to unpack a list.

Thanks!

betatim:
You need to install scikit-optimize from GitHub directly for that feature. It did not yet exist when 0.4 was released.

However I do not think that it will call your objective function with named arguments. I like the idea, so if someone wanted to contribute support for that it would be welcome. One thing that comes to mind straight away is that it would have to continue to work for functions that have no named arguments for backwards compatibility. Not quite sure how much python magic we would have to perform to pull that off.

[...]

Issue #583, opened 2017-12-21
gp_minimize and company can not take points as dictionaries


pavelkomarov:
I had an issue a few days ago (#581) where I wondered whether I could use `BaseEstimator`'s `set_params`-like syntax when defining spaces and points with `BayesSearchCV`. That is `paramName` or `subObject__subObjectParamName` as keys to a dictionary-defined space. It turns out you can, so hooray.

However, I realized I can not actually use `BayesSearchCV` because my data is too large to fit in memory, so there is no way to call `.fit(data)`. As a workaround, I have defined my own objective function which handles batch-reads and does its own cross validation and returns a score. But now I am finding out that the `X_minimize` functions only take lists as their `dimensions` parameter, which is wildly inconvenient because it means I can't just call `estimator.set_params(**point)` inside my objective function, and all the spaces I defined based on dictionaries can't work.

I suggest updating all `X_minimize` functions to support points and spaces defined as dictionaries rather than requiring lists. The list-solution seems to me an outdated one.

pavelkomarov:
I did manage to find a hack to work around this issue, in case anyone else cares (and so I can't forget before I have a chance to implement it):
1. Use the Run object I defined to store, instead of the dictionary-defined space, the alphabetically-sorted keys and the result of skopt's `utils.dimensions_aslist(space)` separately.
2. Pass the listified values from the dictionary as the space to `X_minimize` with the `Run.objectiveFn` as the function to minimize.
3. In `Run`'s `objectiveFn` use the saved, alphabetized keys and the point passed by `base_minimize` on line 225 or 248 to reconstruct the dictionary, which is possible because dimensions are sorted by alphabetized key, and then use this dictionary-defined space to call `estimator.set_params(**dict)`.

Phew.

betatim:
Does https://github.com/scikit-optimize/scikit-optimize/blob/6deff296237422c8b60a5dadb1336dc5e1684c93/skopt/utils.py#L381 not work for you?

We decided to follow the interface of `scipy.optimize` which does not require objective functions to have named arguments. Maybe one way to make it more convenient is to provide a wrapper for objective functions that do support named arguments, something like:
```python
def objective(x=0, y=3., z=0.3):
  pass

def with_named_args(func, space):
  def f(*args):
    return func(point_asdict(*args, space))
  return f

gp_minimize(with_named_args(objective, space), space,...)
```

pavelkomarov:
[...]

PR #590, opened 2017-12-28
Added support for named arguments to objective function.

Hvass-Labs:
This is a solution for issue #570 about named arguments for the objective function.

@betatim proposed using the Python `inspect.signature()` function - first time I heard about it :-)

But I think this a simpler solution that also has backwards compatibility with the current calling style of `func(x)` where `x` is a list. The user just needs to give all dimensions proper names (otherwise an exception is raised), and then set `use_arg_names=True` when calling e.g. `gp_minimize()`. The `use_arg_names` is then passed all the way down to `base_minimize()` which wraps the objective-function so it will be called with the named arguments instead of a list `x` as usual.

I found the function `base_minimize()` very confusing. I added a few TODO-comments which is a hope that you will spend an hour cleaning and commenting this function before you merge the code. It is unlikely that anyone else can / will do it for you.

The code below tests the new feature. Try changing `use_arg_names` so it is called with both the old and new methods. You can also try changing the order of the arguments to `named_fitness()` and it should still work. Also try changing a name e.g. from `activation` to `activation2` and it should crash.

    import numpy as np
    from math import exp
    
    from skopt import gp_minimize
    
    dim_learning_rate = Real(name='learning_rate', low=1e-6, high=1e-2, prior='log-uniform')
    dim_num_dense_layers = Integer(name='num_dense_layers', low=1, high=5)
    dim_num_dense_nodes = Integer(name='num_dense_nodes', low=5, high=512)
    dim_activation = Categorical(name='activation', categories=['relu', 'sigmoid'])
    
    dimensions = [dim_learning_rate,
                  dim_num_dense_layers,
                  dim_num_dense_nodes,
                  dim_activation]
    
    default_parameters = [1e-4, 1, 64, 'relu']
    
    # TODO: Try changing the order of the args, which should work.
    # TODO: Also try renaming an arg, which should crash.
    def named_fitness(learning_rate, num_dense_layers, num_dense_nodes, activation):
        """New type of fitness function with named arguments."""
    
        fitness = ((exp(learning_rate) - 1.0) * 1000) ** 2 + \
                  (num_dense_layers) ** 2 + \
                  (num_dense_nodes/100) ** 2
    
        fitness *= 1.0 + 0.1 * np.random.rand()
    
        if activation == 'sigmoid':
            fitness += 10
    
        return fitness
    
    
    def model_fitness(x):
        """Typical fitness function with list-argument."""
        learning_rate, num_dense_layers, num_dense_nodes, activation = x
    
        fitness = named_fitness(learning_rate=learning_rate,
                                num_dense_layers=num_dense_layers,
                                num_dense_nodes=num_dense_nodes,
                                activation=activation)
    
        return fitness
    
    
    print(model_fitness(x=default_parameters))
    
    # TODO: Try changing this!
    use_arg_names = False
    
    func = named_fitness if use_arg_names else model_fitness
    
    search_result = gp_minimize(func=func,
                                dimensions=dimensions,
                                use_arg_names=use_arg_names,
                                n_calls=40,
                                x0=default_parameters)
    
    print(search_result.x)
    print(search_result.fun)
    
    for fitness, x in sorted(zip(search_result.func_vals, search_result.x_iters)):
        print(fitness, x)

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -1,10 +1,11 @@
+# TODO: This doc-string is way too short and quite meaningless.
 """
 Abstraction for optimizers.
 
 It is sufficient that one re-implements the base estimator.
 """
 
-import copy
+from copy import copy
 import inspect
 import numbers
 from collections import Iterable
@@ -17,7 +18,67 @@ from .optimizer import Optimizer
 from ..utils import eval_callbacks
 
 
+class _WrapFuncNamedArgs:
+    """
+    Wrapper for the objective function. This calls the objective
+    function `func()` with named arguments instead of a list of parameters.
+
+    For example, instead of calling `func([123, 3.0, 'hello'])`,
+    it calls `func(foo=123, bar=3.0, baz='hello')` for a search-space
+    with dimensions named `['foo', 'bar', 'baz']`
+    """
+
+    def __init__(self, func, dimensions):
+        """Create a callable object instance that wraps an objective function.
+
+        Parameters
+        ----------
+        * `func` [callable]:
+            Function to minimize. Takes a list of parameters and
+            returns the objective or fitness value.
+
+        * `dimensions` [list of Dimension instances]:
+            List of search-space dimensions. These must all be named.
+            Use a list of instances of `Real`, `Categorical` or `Integer`
+            from `skopt.space`.
+        """
+
+        # Set the objective function.
+        self.func = func
+
+        # Get the names of all search-space dimensions.
+        self.dim_names = [dim.name for dim in dimensions]
+
+        # Ensure all dimension names exist.
+        if None in self.dim_names:
+            raise ValueError("All dimensions must have names when `use_arg_names=True`.")
+
+    def __call__(self, x):
+        """This makes the object callable so it seamlessly wraps `func(x)`.
+
+        Parameters
+        ----------
+        * `x` [list]:
+            Parameters for a location in the search-space.
+            
+        Returns
+        -------
+        * `fitness` [float]:
+            Fitness of the objective function at the given location `x`.
+        """
+
+        # Create a dictionary with the named parameters.
+        x_dict = {name: value for name, value in zip(self.dim_names, x)}
+
+        # Call the objective function with these named arguments.
+        fitness = self.func(**x_dict)
+
+        return fitness
+
+
+# TODO: There is no doc-string here! What does this function do?!
 def base_minimize(func, dimensions, base_estimator,
+                  use_arg_names=False,
                   n_calls=100, n_random_starts=10,
                   acq_func="EI", acq_optimizer="lbfgs",
                   x0=None, y0=None, random_state=None, verbose=False,
@@ -44,6 +105,13 @@ def base_minimize(func, dimensions, base_estimator,
 
          NOTE: The upper and lower bounds are inclusive for `Integer`
          dimensions.
+
+    * `use_arg_names` [bool, default=False]:
+        Whether to use use names for the search-space dimensions
+        when calling `func()`, so instead of calling `func(x)` 
+        with a single list `x`, it calls e.g.
+        `func(foo=123, bar=3.0, baz='hello')` for a search-space
+         with dimensions named `['foo', 'bar', 'baz']`
 
     * `base_estimator` [sklearn regressor]:
         Should inherit from `sklearn.base.RegressorMixin`.
@@ -159,10 +227,20 @@ def base_minimize(func, dimensions, base_estimator,
         - `rng` [RandomState instance]: State of the random state
            at the end of minimization.
 
-        For more details related to the OptimizeResult object, refer
+        For more details related to the OptimizeResult object, refer to:
         http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html
     """
-    specs = {"args": copy.copy(inspect.currentframe().f_locals),
+
+    # TODO: Please clean up and comment the rest of this code.
+    # TODO: It is VERY hard to understand and there appears to be tautologies,
+    # TODO: which indicates that several people have worked on this and they
+    # TODO: also don't understand it and just keep adding code to fix bugs etc.
+
+    # Wrap the objective function if it uses dimension-names as arguments.
+    if use_arg_names:
+        func = _WrapFuncNamedArgs(func=func, dimensions=dimensions)
+
+    specs = {"args": copy(inspect.currentframe().f_locals),
              "function": inspect.currentframe().f_code.co_name}
 
     acq_optimizer_kwargs = {
@@ -170,6 +248,9 @@ def base_minimize(func, dimensions, base_estimator,
         "n_jobs": n_jobs}
     acq_func_kwargs = {"xi": xi, "kappa": kappa}
 
+    # TODO: These manipulations to x0 and y0 are VERY confusing!
+    # TODO: And further below there is a check: "if x0 is not None"
+    # TODO: but that is apparently always True after this.
     # Initialize with provided points (x0 and y0) and/or random points
     if x0 is None:
         x0 = []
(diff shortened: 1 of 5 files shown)


betatim:
Thanks for this. Let's see if we can get some feedback from others on introducing a new argument. Personally I quite like the idea of having a adaptor that takes care of the boiler plate conversion. Mostly because there are already a lot of arguments which leads to long doc strings, and long function signatures.

> I found the function `base_minimize()` very confusing. I added a few TODO-comments which is a hope that you will spend an hour cleaning and commenting this function before you merge the code. It is unlikely that anyone else can / will do it for you.

As a general rule a PR is edited by the contributor, including adding tests, making sure existing tests pass, documentation, code, etc. Frequently others will make suggestions or ask for changes which the person who created the PR then implements. Very rarely people do collaborate on a PR but this means a third person has to review it so it can be merged. Keeping the roles separated works very well for us.

Your comments on `base_minimize` are welcome, so thank you for taking the time to write them down. Source code is not the place to store them however, issues is the place to do that or if relevant to this PR create a comment on that line to discuss how best to address something.

This means I won't +1 this PR until you remove the TODO comments or convert them.

skopt/optimizer/base.py
```diff
@@ -1,10 +1,11 @@
+# TODO: This doc-string is way too short and quite meaningless.
```
betatim: We are working on this code base together and expect to work through issues collaboratively. If you've taken the time to understand this section, you probably understand there are tradeoffs. We can work through them. As it stands, these TODO comments feel like directed angst rather than a commitment to contribute to reducing confusion.

skopt/optimizer/base.py
```diff
@@ -17,7 +18,67 @@
 from ..utils import eval_callbacks
 
 
+class _WrapFuncNamedArgs:
```
betatim: Instead of creating (and having to maintain) our own function wrapper we could use [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps) like so:
```python
In [3]: def adaptor(f):
   ...:     @wraps(f)
   ...:     def wrapper(*args):
   ...:         print('modify args', args)
   ...:         return f(*args)
   ...:     return wrapper
   ...: 

In [4]: def my_objective(x, y, z):
   ...:     """doc here"""
   ...:     return x + y + z
   ...: 

In [5]: adapted_obj = adaptor(my_objective)

In [6]: adapted_obj?
Signature: adapted_obj(x, y, z)
Docstring: doc here
File:      ~/Downloads/<ipython-input-4-84edd263a788>
Type:      function
```
it takes care of forwarding doc strings, and signatures, etc.

skopt/optimizer/base.py
```diff
@@ -220,33 +303,42 @@ def base_minimize(func, dimensions, base_estimator,
     # setting the scope for these variables
     result = None
 
-    # User suggested points at which to evaluate the objective first
-    if x0 and y0 is None:
-        y0 = list(map(func, x0))
-        n_calls -= len(y0)
+    # TODO: I tried to clean this up a little, but it is VERY confusing.
+    # Pass user suggested initialisation points to the optimizer.
+    if x0 is not None:
```
betatim: `x0` should already be initialised as an empty list by the time we get here, so we should keep the old test (`if x0`).

I like the comment but splitting up the if clause makes it harder to understand

skopt/optimizer/base.py
```diff
@@ -220,33 +303,42 @@ def base_minimize(func, dimensions, base_estimator,
...
         result = optimizer.tell(x0, y0)
         result.specs = specs
 
         if eval_callbacks(callbacks, result):
             return result
 
-    # Bayesian optimization loop
+    # Bayesian optimization loop.
     for n in range(n_calls):
+        # Get next point in the search-space.
```
betatim: The next three comments restate what the line of code says so I would remove them

betatim:
I would provide the wrapper in `skopt.utils` and let the user pass in the wrapped objective instead of adding a new argument to influence the behaviour. This has two advantages: 

1. it does not increase the number of arguments. There is a feeling (and corresponding discussion) that there are already too many, and how could we reduce them
1. it provides the same functionality to power users who use `Optimizer` directly.

betatim:
One more thought on having named arguments.

Without any changes to scikit-optimize having named arguments (or a nice way to unpack them) already works (as long as the dimensions are defined in the same order):
```
# note the double parentheses 
def named_fitness((learning_rate, num_dense_layers, num_dense_nodes, activation)):
  pass
```
The use case in #583 and when I've wanted the arguments to be named was when I had to pass them on to something else (like a scikit-learn estimator) where you want a [`dict` to pass to `set_params()`](http://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html)

This makes me think we should remind people of argument unpacking in python and instead add support for objectives that receive a "single argument" like so: `objective(**params)` where we pass in a dictionary.

skopt/optimizer/base.py
Hvass-Labs: It is an indication that your code is frustrating to work on for others and the TODOs point to specific areas that might be improved. I asked myself last night if my old code was also undocumented and unpolished. I looked at ArrayOps which I did almost 12 years ago and look at how beautifully it was designed, implemented and documented. Even though it has been so long and I haven't worked on it since, and it is very advanced C++ programming, I was still able to grasp it again by simply browsing my docs and commented code. Take a look at the manual http://www.hvass-labs.org/projects/arrayops/files/ArrayOps2.pdf Then take a look at e.g. `plots.py` in your project which was particularly bad, or even the files we are discussing here. ArrayOps had almost no users (I think it was used in teaching advanced C++ but not really used in production). Your library has lots of users already. Why not make the code beautiful so others can contribute more easily? Right now there's a huge barrier to making contributions because your code is often impossible to understand.

Hvass-Labs:
@betatim Thanks again for the quick response! We don't agree on a lot of issues especially regarding commenting of code and who should do what on a project, but you have a strong sense of making clean API design and I appreciate that! Why don't you just lead the project? It is already my impression that you are the main developer on this project? You should just take charge.

I like that you keep reducing the ideas for API design until it is as simple as possible. That is how good design is done. It is important that you scrutinize the contributions from other team members as well. I don't think you would ever have approved `plots.py` if you had scrutinized it like you are doing with my PRs.

I am also learning something from your ideas and feedback, including Python features I wasn't aware of, so thanks for that!

Your final idea on this issue was this:

    # note the double parentheses 
    def named_fitness((learning_rate, num_dense_layers, num_dense_nodes, activation)):
        pass

I hadn't thought about unpacking `x` already in the function-declaration. But it is semantically similar to what I previously did myself:

    def named_fitness(x):
	learning_rate, num_dense_layers, num_dense_nodes, activation = x

However, this has a few problems:

1) It is a hack of the Python API and as you note, you have to write double-parentheses in the function declaration which is confusing.

2) One advantage of using Python's named arguments is that you have added security against making mistakes by re-ordering the arguments. This is probably an even more important reason than saving a line of code.

I thought about making a function-wrapper e.g. in `skopt.utils`. I think the main reason I decided against it was because named args is a feature I would personally use all the time, so I thought it would be better to include it directly in the API.

I now see that my wrapper-function is actually incorrect because `dimensions` may not be a list of `Dimension`-objects. Maybe it would indeed be better to make the wrapper-function external so we also avoid having to deal with all these special cases of the `dimensions` argument. I'll take a look and re-submit later. I'm working on `plots.py` now, which is taking forever because it is so hard to understand the code.

skopt/optimizer/base.py
Hvass-Labs: The code is easy to understand for you because you look at it every day. For newcomers the comments help and they don't do any harm to experts. But I'm not going to argue with you. I am just informing you that it would be easier for others to help you, if your code was better documented.

skopt/optimizer/base.py
Hvass-Labs: As I recall there was an `if x0:` statement here. That is not Python, that is C-style programming! And it can actually fail in Python if the object has overrided some dunder-method (I can't remember which one but I think there's a PyCon talk about this ...) In Python you should make it explicit what you are testing against e.g. by writing `if x0 is not None:`

I think I'll re-submit the wrapper function as discussed below and I'll remove all these changes to the other files and just add the wrapper-function with as few changes as possible. It would be nice if you would go over some of these files and polish them.

skopt/optimizer/base.py
Hvass-Labs: Thanks for the tip, I'll take a look.

glouppe:
Hi,

Thanks for your contribution. Can you remove your inline comments so that we can focus the review on changes that are concerned with this proposal? (i.e., adding support for named arguments)

We acknowledged that there might things that are difficult for newcomers to understand (as is with any complex bit of research code) and are open to discuss them, but in order to not waste everyone's time, it is probably more efficient to discuss them in standalone and targeted PRs.

Hvass-Labs:
A much better version of this is made in PR #592. So I am closing this PR.

PR #592, opened 2018-01-01
[MRG+1] Added support for named arguments with objective function (NEW version)

Hvass-Labs:
This is a much improved version of PR #590 which was also discussed in various issues including #570 and #583.

It might be a good idea if I update the doc-strings of all optimizers to briefly describe that `func` can be wrapped using this decorator so as to use named arguments. Do you agree?

Thanks to @betatim for suggesting the use of a decorator which made this version very elegant.

I have not worked very much with decorators and this is a fairly advanced usage scenario, so I think it is better to direct the interested user to do an internet search if they want an explanation on how advanced decorators work in Python. See my comments in the code.

However, I do hope you appreciate the otherwise detailed level of commenting I have added, which is what I have been discussing in my other issues and PRs. It really makes it so much easier for others to understand the intention of the code in case there a bugs, or in case they want to modify the code in the future.

It is extremely important to try and polish and document the code when it is first added to the library. If I had not polished and documented this code properly and left it for others to do, then they would have had to guess at the intended semantics, and they would have spent maybe 10x as long as it has now taken me to polish and document this code - and they would have been pissed off that I hadn't done it properly in the first place.

I admit openly that this version is superior to what any of us suggested individually, and I suppose this shows how great the design and implementation can be under collaboration - but it requires for each person to properly polish and document his additions to the code. That is what I understand to be real collaboration, and not cutting corners which creates a much larger burden for future developers.

You can test this addition with this code:

    from skopt.space import Real
    from skopt.optimizer import forest_minimize
    from skopt.utils import wrap_named_args
    
    dim1 = Real(name='foo', low=0.0, high=1.0)
    dim2 = Real(name='bar', low=0.0, high=1.0)
    dim3 = Real(name='baz', low=0.0, high=1.0)
    
    dimensions = [dim1, dim2, dim3]
    
    @wrap_named_args(dimensions=dimensions)
    def my_objective_function(foo, bar, baz):
        # print("foo", foo)
        # print("bar", bar)
        # print("baz", baz)
        return foo ** 2 + bar ** 4 + baz ** 8
    
    # Test that the wrapped objective function is now callable
    # using a list x instead of named arguments.
    print("Test 1:", my_objective_function(x=[0.1, 1.0, 0.5]))
    print("Test 2:", my_objective_function([0.1, 1.0, 0.5]))
    
    result = forest_minimize(my_objective_function, dimensions=dimensions,
                             n_calls=20, base_estimator="ET", random_state=4)
    
    print("Best objective value:", result.fun)
    print("Best parameters:", result.x)

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -240,8 +240,8 @@ class Real(Dimension):
                 self.transform_ == other.transform_)
 
     def __repr__(self):
-        return "Real(low={}, high={}, prior='{}', transform='{}', name='{}')".format(
-            self.low, self.high, self.prior, self.transform_, self.name)
+        return "Real(low={}, high={}, prior='{}', transform='{}')".format(
+            self.low, self.high, self.prior, self.transform_)
 
     def inverse_transform(self, Xt):
         """Inverse transform samples from the warped space back into the
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -240,8 +240,8 @@ class Real(Dimension):
                 self.transform_ == other.transform_)
 
     def __repr__(self):
-        return "Real(low={}, high={}, prior='{}', transform='{}')".format(
-            self.low, self.high, self.prior, self.transform_)
+        return "Real(low={}, high={}, prior='{}', transform='{}', name='{}')".format(
+            self.low, self.high, self.prior, self.transform_, self.name)
 
     def inverse_transform(self, Xt):
         """Inverse transform samples from the warped space back into the
--- a/skopt/tests/test_utils.py
+++ b/skopt/tests/test_utils.py
@@ -5,6 +5,8 @@ from sklearn.utils.testing import assert_array_equal
 from sklearn.utils.testing import assert_equal
 from sklearn.utils.testing import assert_true
 
+import numpy as np
+
 from skopt import gp_minimize
 from skopt import load
 from skopt import dump
@@ -20,6 +22,8 @@ from skopt.utils import dimensions_aslist
 from skopt.utils import has_gradients
 from skopt.utils import cook_estimator
 from skopt.utils import normalize_dimensions
+from skopt.utils import use_named_args
+from skopt.space import Real
 
 
 def check_optimization_results_equality(res_1, res_2):
@@ -165,3 +169,59 @@ def test_normalize_dimensions(dimensions, normalizations):
     space = normalize_dimensions(dimensions)
     for dimension, normalization in zip(space, normalizations):
         assert dimension.transform_ == normalization
+
+
+@pytest.mark.fast_test
+def test_use_named_args():
+    """
+    Test the function wrapper @use_named_args which is used
+    for wrapping an objective function with named args so it
+    can be called by the optimizers which only pass a single
+    list as the arg.
+    
+    This test does not actually use the optimizers but merely
+    simulates how they would call the function.
+    """
+
+    # Define the search-space dimensions. They must all have names!
+    dim1 = Real(name='foo', low=0.0, high=1.0)
+    dim2 = Real(name='bar', low=0.0, high=1.0)
+    dim3 = Real(name='baz', low=0.0, high=1.0)
+
+    # Gather the search-space dimensions in a list.
+    dimensions = [dim1, dim2, dim3]
+
+    # Parameters that will be passed to the objective function.
+    default_parameters = [0.5, 0.6, 0.8]
+
+    # Define the objective function with named arguments
+    # and use this function-decorator to specify the search-space dimensions.
+    @use_named_args(dimensions=dimensions)
+    def func(foo, bar, baz):
+        # Assert that all the named args are indeed correct.
+        assert foo == default_parameters[0]
+        assert bar == default_parameters[1]
+        assert baz == default_parameters[2]
+
+        # Return some objective value.
+        return foo ** 2 + bar ** 4 + baz ** 8
+
+    # Ensure the objective function can be called with a single
+    # argument named x.
+    res = func(x=default_parameters)
+    assert (isinstance(res, float))
+
+    # Ensure the objective function can be called with a single
+    # argument that is unnamed.
+    res = func(default_parameters)
+    assert (isinstance(res, float))
+
+    # Ensure the objective function can be called with a single
+    # argument that is a numpy array named x.
+    res = func(x=np.array(default_parameters))
+    assert (isinstance(res, float))
+
+    # Ensure the objective function can be called with a single
+    # argument that is an unnamed numpy array.
+    res = func(np.array(default_parameters))
+    assert (isinstance(res, float))
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -1,4 +1,5 @@
 from copy import deepcopy
+from functools import wraps
 
 import numpy as np
 from scipy.optimize import OptimizeResult
@@ -16,10 +17,7 @@ from .learning.gaussian_process.kernels import ConstantKernel
 from .learning.gaussian_process.kernels import HammingKernel
 from .learning.gaussian_process.kernels import Matern
 
-from .space import Categorical
-from .space import Integer
-from .space import Real
-from .space import Space
+from .space import Space, Categorical, Integer, Real, Dimension
 
 __all__ = (
     "load",
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
                                    "(%s)" % type(dimension))
 
     return Space(transformed_dimensions)
+
+
+def wrap_named_args(dimensions):
+    """
+    Wrapper for an objective function that uses named arguments to make it
+    compatible with optimizers that use a single list of parameters.
+
+    Your objective function can be defined as being callable using named
+    arguments: `func(foo=123, bar=3.0, baz='hello')` for a search-space
+    with dimensions named `['foo', 'bar', 'baz']`. But the optimizer
+    will only pass a single list `x` of unnamed arguments when calling
+    the objective function: `func(x=[123, 3.0, 'hello'])`. This wrapper
+    converts your objective function with named arguments into one that
+    accepts a list as argument, while doing the conversion automatically.
+
+    The advantage of this is that you don't have to unpack the list of
+    arguments `x` yourself, which makes the code easier to read and
+    also reduces the risk of bugs if you change the number of dimensions
+    or their order in the search-space.
+
+    Example Usage
+    -------------
+    # Define the search-space dimensions. They must all have names!
+    dim1 = Real(name='foo', low=0.0, high=1.0)
+    dim2 = Real(name='bar', low=0.0, high=1.0)
+    dim3 = Real(name='baz', low=0.0, high=1.0)
+
+    # Gather the search-space dimensions in a list.
+    dimensions = [dim1, dim2, dim3]
+
+    # Define the objective function with named arguments
+    # and use this function-decorator to specify the search-space dimensions.
+    @wrap_named_args(dimensions=dimensions)
+    def my_objective_function(foo, bar, baz):
+        return foo ** 2 + bar ** 4 + baz ** 8
+
+    # Now the function is callable from the outside as
+    # `my_objective_function(x)` where `x` is a list of unnamed arguments,
+    # which then wraps your objective function that is callable as
+    # `my_objective_function(foo, bar, baz)`.
+    # The conversion from a list `x` to named parameters `foo`, `bar`, `baz`
+    # is done automatically.
+
+    # Run the optimizer on the wrapped objective function which is called as
+    # `my_objective_function(x)` as expected by `forest_minimize()`.
+    result = forest_minimize(func=my_objective_function, dimensions=dimensions,
+                             n_calls=20, base_estimator="ET", random_state=4)
+
+    # Print the best-found results.
+    print("Best fitness:", result.fun)
+    print("Best parameters:", result.x)
+
+    Parameters
+    ----------
+    * `dimensions` [list(Dimension)]:
+        List of `Dimension`-objects for the search-space dimensions.
+
+    Returns
+    -------
+    * `wrapped_func` [callable] 
+        Wrapped objective function.
+    """
+
+    def decorator(func):
+        """
+        This uses more advanced Python features to wrap `func` using a
+        function-decorator. To get explanations of how this works in Python,
+        please do an internet search for e.g.:
+        Python function decorator with additional arguments.
+        """
+
+        # Ensure all dimensions are correctly typed.
+        if not all(isinstance(dim, Dimension) for dim in dimensions):
+            raise ValueError("All dimensions must be instances of the Dimension-class.")
+
+        # Ensure all dimensions have names.
+        if any(dim.name is None for dim in dimensions):
+            raise ValueError("All dimensions must have names.")
+
+        @wraps(func)
+        def wrapper(x):
+            # Ensure the number of dimensions match
+            # the number of parameters in the list x.
+            if len(x) != len(dimensions):
+                msg = "Mismatch in number of search-space dimensions. " \
+                      "len(dimensions)=={} and len(x)=={}"
+                msg = msg.format(len(dimensions), len(x))
+                raise ValueError(msg)
+
+            # Create a dict where the keys are the names of the dimensions
+            # and the values are taken from the list of parameters x.
+            arg_dict = {dim.name: value for dim, value in zip(dimensions, x)}
+
+            # Call the wrapped objective function with the named arguments.
+            objective_value = func(**arg_dict)
+
+            return objective_value
+
+        return wrapper
+
+    return decorator
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -565,21 +565,61 @@ def use_named_args(dimensions):
     def decorator(func):
         """
         This uses more advanced Python features to wrap `func` using a
-        function-decorator. To get explanations of how this works in Python,
-        please do an internet search for e.g.:
-        Python function decorator with additional arguments.
+        function-decorator, which are not explained so well in the
+        official Python documentation.
+
+        A good video tutorial explaining how this works is found here:
+        https://www.youtube.com/watch?v=KlBPCzcQNU8
+
+        Parameters
+        ----------
+        * `func` [callable]:
+            Function to minimize. Should take *named arguments*
+            and return the objective value.
         """
 
         # Ensure all dimensions are correctly typed.
         if not all(isinstance(dim, Dimension) for dim in dimensions):
-            raise ValueError("All dimensions must be instances of the Dimension-class.")
+            # List of the dimensions that are incorrectly typed.
+            err_dims = list(filter(lambda dim: not isinstance(dim, Dimension),
+                                   dimensions))
+
+            # Error message.
+            msg = "All dimensions must be instances of the Dimension-class, but found: {}"
+            msg = msg.format(err_dims)
+            raise ValueError(msg)
 
         # Ensure all dimensions have names.
         if any(dim.name is None for dim in dimensions):
-            raise ValueError("All dimensions must have names.")
+            # List of the dimensions that have no names.
+            err_dims = list(filter(lambda dim: dim.name is None, dimensions))
+
+            # Error message.
+            msg = "All dimensions must have names, but found: {}"
+            msg = msg.format(err_dims)
+            raise ValueError(msg)
 
         @wraps(func)
         def wrapper(x):
+            """
+            This is the code that will be executed every time the
+            wrapped / decorated `func` is being called.
+            It takes `x` as a single list of parameters and
+            converts them to named arguments and calls `func` with them.
+            
+            Parameters
+            ----------
+            * `x` [list]:
+                A single list of parameters e.g. `[123, 3.0, 'linear']`
+                which will be converted to named arguments and passed
+                to `func`.
+        
+            Returns
+            -------
+            * `objective_value` 
+                The objective value returned by `func`.
+            """
+
             # Ensure the number of dimensions match
             # the number of parameters in the list x.
             if len(x) != len(dimensions):
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -501,10 +501,10 @@ def normalize_dimensions(dimensions):
     return Space(transformed_dimensions)
 
 
-def wrap_named_args(dimensions):
+def use_named_args(dimensions):
     """
-    Wrapper for an objective function that uses named arguments to make it
-    compatible with optimizers that use a single list of parameters.
+    Wrapper / decorator for an objective function that uses named arguments
+    to make it compatible with optimizers that use a single list of parameters.
 
     Your objective function can be defined as being callable using named
     arguments: `func(foo=123, bar=3.0, baz='hello')` for a search-space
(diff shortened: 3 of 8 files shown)


Hvass-Labs:
I wonder if a better name for `@wrap_named_args` would perhaps be `@use_named_args`? Any other suggestions?

glouppe:
Thanks! This looks good at first sight. Can you add tests though before we start the review process?

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
...
+                      "len(dimensions)=={} and len(x)=={}"
+                msg = msg.format(len(dimensions), len(x))
+                raise ValueError(msg)
+
+            # Create a dict where the keys are the names of the dimensions
+            # and the values are taken from the list of parameters x.
+            arg_dict = {dim.name: value for dim, value in zip(dimensions, x)}
+
+            # Call the wrapped objective function with the named arguments.
+            objective_value = func(**arg_dict)
```
glouppe: I am not convinced the comments above (line 591, 592 and 595) bring anything helpful except repeating the semantics of the code.

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
                                    "(%s)" % type(dimension))
 
     return Space(transformed_dimensions)
+
+
+def wrap_named_args(dimensions):
+    """
+    Wrapper for an objective function that uses named arguments to make it
```
glouppe: Wrapper -> Decorator? That's the standard name for that sort of things in Python.

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
...
+        Wrapped objective function.
+    """
+
+    def decorator(func):
+        """
+        This uses more advanced Python features to wrap `func` using a
+        function-decorator. To get explanations of how this works in Python,
+        please do an internet search for e.g.:
+        Python function decorator with additional arguments.
+        """
```
glouppe: I would just remove that, or make that comment shorter. This is standard practice in Python.

Hvass-Labs:
Before I start, let me once again say that I think the *idea* of your project is absolutely awesome! But your code is often really, really bad! I wish you could take this as a big compliment that you are on to something very important with this library, but you need to improve certain aspects, and that starts with accepting that your code "style" is actually really awful!

As I wrote in https://github.com/scikit-optimize/scikit-optimize/pull/579#issuecomment-354550973 I expected you would demand that I remove all the comments from my code. That is not going to happen! If I add code to your project, then it will be beautifully commented and explained like this, otherwise I will retract the PRs. I will not have my name associated with crappy code! I get job offers from all over the world - even from places you probably wouldn't believe! A part of the reason is that I am so good at explaining what I do and write code so well. Whether you choose to remove all the comments afterwards is your business, but you would be making a huge mistake. Let me explain why.

There is one main thing that both you and @betatim misunderstand: This code is *not* written for *you*! It is written for the person who in 5 years has a bug or wants to modify the code. None of us will be around to help, or we will certainly have forgotten what exactly the code is supposed to do if it hasn't been documented really well. My code can be understood very easily by even first-year students, in part because the code itself is very clean and polished, but especially because it is so well documented *in English*. You don't need to have a deep understanding of `skopt`, `matplotlib`, or any other API, because the English comments describe very well what is supposed to happen. Yes, people could look up what all those API calls are doing, but I am saving them a ton of time by describing it *in English*, so they can very quickly understand what the *intention* is with the code and how everything fits together.

Your code on the other hand is extremely hard to understand - both because it is often very poorly programmed (it really is!), but the problem is then amplified greatly by the lack of comments, so the *intention* of the code isn't even clear. This makes it one gigantic puzzle to fix bugs or extend your code. Your code is confusing even to yourself, as I can see from the discussions here when bugs come up, semantics are unclear, etc.

I already showed @betatim some work I did 12 years ago on advanced C++ called ArrayOps. It was beautifully designed, coded and documented. Even though I had not really programmed C++ in 12 years - I was still able to grasp my old work from just browsing the manual and the well-documented code for 10 minutes or so the other day. Same goes for my SwarmOps project, which is also quite old now - and by the way is a much simpler way of doing hyper-parameter optimization than skopt, which is why I am so excited about skopt!

Now I am of course becoming well-known for my TensorFlow tutorials. They do not get the most views, but I will bet you that those tutorials have taught more people how to program TensorFlow than any other tutorials. And the reason people really like them, is in part because I am so disciplined and careful in polishing and explaining everything. You don't have to lookup API calls, or guess at semantics or anything like that, because it is all beautifully explained and commented *in English*.

With these examples it should be obvious that at least when I was younger I was a far better software engineer than you are now. And I was then, and am still, far better than you at explaining complicated things and documenting my code so it can be understood by mostly everyone. So I really don't understand why you don't want to learn from my way of doing things, or at least welcome it in the code that *I* have added to your project, instead of stubbornly insisting that your way is better.

http://www.hvass-labs.org/projects/arrayops/files/ArrayOps2.pdf
https://github.com/Hvass-Labs/swarmops/blob/master/swarmops/DE.py
https://github.com/Hvass-Labs/TensorFlow-Tutorials

glouppe:
Instead of keeping repeating how bad we are at writing code and how we dare to publish undocumented lines of Python, can we stick to the review process?

- I asked for tests, which is a necessary thing if you wish to contribute to this project.
- I expressed **minor** concerns regarding the unnecessary verbosity of some of the comments you added. There is no need to take this personally nor to attack us every time we express our opinion.

skopt/utils.py
Hvass-Labs: I think you are deliberately nitpicking here and although I'm not a Python expert I doubt that what you're saying is strictly true. We are importing and using something called *wraps* from `functools` and they repeatedly talk about *wrapping* functions in the documentation, and they even call it `wrapper` in their code examples. Yes, these are also called function decorators in Python when used with the @ symbol. But this code may be used by people who are not Python experts so they may not know what a decorator is, but they most likely know what a function-wrapper is.

I'll tell you what. I'll call it a Wrapper / Decorator so everyone is happy :-) Besides I think `use_named_args` is a better name as well.

https://docs.python.org/2/library/functools.html

skopt/utils.py
Hvass-Labs: You are wrong. These comments explain *in English* what is actually fairly unusual Python programming techniques that certainly won't be understood very well by people who are not Python experts. Once again, this code is *not* written for *you*!

skopt/utils.py
Hvass-Labs: This is an advanced usage of Python decorators. It is poorly understood even by experienced Python programmers as I could see on StackOverflow - I myself have never used this technique before and I had to look it up. It took me quite a while to find out how to do it, and it is not obvious where to find information on this. There is no good doc-page for this in the Python manuals. So to save people a lot of time, I give a good suggestion for a google search-string if they want to understand what is going on in this code.

Please stop being so goddamn stingy with comments in the code!

glouppe: > Please stop being so goddamn stingy with comments in the code!

I must admit, I laughed reading that.

Please read my comment again: I asked to either remove or shorten your comment. Telling people to google if they dont understand Python does not really help. 

So let's be constructive. Instead I suggest to point the readers to the page you read yourself to implement this pattern. This can be shortly expressed in 1 or 2 lines at most, and is more helpful to everyone.

Hvass-Labs:
I am trying to convince you that in 5 years nobody will understand your code if you don't put more effort into it. Not even you will understand it. And yes I find it offensive when you want me to write as crappy code as you do, when I have clearly put effort into making it beautiful and easy to understand for everyone.

Now I have added tests. And I am certain that you are unhappy with the comments in the code. But I glanced at some of your other tests and they are completely incomprehensible to anybody except yourself. And in 5 years you won't even know what they're doing.

I really wish I knew what to say to make you realize this. But I don't think I can, so let's just finish this without killing each other :-)

skopt/utils.py
Hvass-Labs: It was a bunch of pages and none of them were really good. I picked up a little info here and a little there. If you google for the 'wrong' keywords you find even worse pages. If there was a single good webpage explaining this I would of course have linked to it. I added these keywords because they really helped me. Again, this is not a common Python feature that is being used and even experienced developers on StackOverflow were confused about it.

Again you are deliberately nitpicking what you want changed in this PR.

You can bulldoze all these comments a month from now and I will never notice, but I suggest you instead have some empathy for who is going to be reading the code in the future.

skopt/space/space.py
```diff
@@ -240,8 +240,8 @@ def __eq__(self, other):
                 self.transform_ == other.transform_)
 
     def __repr__(self):
-        return "Real(low={}, high={}, prior='{}', transform='{}')".format(
-            self.low, self.high, self.prior, self.transform_)
+        return "Real(low={}, high={}, prior='{}', transform='{}', name='{}')".format(
+            self.low, self.high, self.prior, self.transform_, self.name)
```
glouppe: Can you also do that for dimensions of other types, for consistency? (Integer, etc).

glouppe: Disclaimer: this is not a nitpick.

glouppe: Also, since names are `None` by default, I am not sure we actually want to print `name=None` in the default case. What do others think?

Hvass-Labs: Now there's a good review request! :-)

It is already done in PR #579 which also fixes other things with `Dimension`-names. I just needed a single one for testing in this PR.

glouppe: Can you backport from #579 please? (That's why we prefer self-contained atomic PRs that target a specific goal, to avoid these situations...)

Hvass-Labs: I'm not an expert at git so I thought this would merge seamlessly as the change is identical for the two PRs. But this code was actually not required in this PR, I only needed it for some print-testing. So I just removed it from this PR.

glouppe:
Alright, thanks. These changes are good to go for me. I'll let a second opinion pitch in for merge.

Hvass-Labs:
Very good, thank you!

Hvass-Labs:
@betatim is familiar with this and I give him partial credit for the solution as he came up with the idea of using a decorator (although not the actual implementation). So perhaps he could test if he is happy with it as well?

Hvass-Labs:
I think it would be helpful to add a short comment to the doc-strings of all optimizers so people know they can use this.

This is the doc-string from `forest_minimize` but I think they are all identical:

    * `func` [callable]:
        Function to minimize. Should take a array of parameters and
        return the function values.

After:

    * `func` [callable]:
        Function to minimize. Should take an array named `x` of parameters
	and return the objective value as a scalar of type `float`.

        If you have a search-space where all dimensions have names,
        then you can use `skopt.utils.use_named_args` as a decorator
        on your objective function, in order to call it directly
        with the named arguments.

I don't think you support multi-objective optimization, right? So `func` only returns scalar values, and not arrays, right?

Should I add this doc-string to all optimizers?

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
...
+
+        # Ensure all dimensions are correctly typed.
+        if not all(isinstance(dim, Dimension) for dim in dimensions):
```
iaroslav-ai: Might be good to print the value of `dim` which is not an instance of `Dimension` - could help solve bugs faster for future users.

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
...
+            raise ValueError("All dimensions must be instances of the Dimension-class.")
+
+        # Ensure all dimensions have names.
+        if any(dim.name is None for dim in dimensions):
+            raise ValueError("All dimensions must have names.")
```
iaroslav-ai: Maybe same as above - print which dimension does not have a name, should help identify a wrong dimension faster

skopt/utils.py
```diff
@@ -501,3 +499,104 @@ def normalize_dimensions(dimensions):
...
+
+        @wraps(func)
+        def wrapper(x):
+            # Ensure the number of dimensions match
```
iaroslav-ai: Consider adding a docstring explaining the type of x and what it is. It seems that Pycharm for instance can infer expected type of variable from docstring, providing extra variable checks and warnings, which sometimes save me bugs. Possibly could also simplify comments as per discussion below

skopt/tests/test_utils.py
```diff
@@ -165,3 +169,59 @@ def test_normalize_dimensions(dimensions, normalizations):
...
+
+    # Ensure the objective function can be called with a single
+    # argument that is a numpy array named x.
+    res = func(x=np.array(default_parameters))
+    assert (isinstance(res, float))
+
+    # Ensure the objective function can be called with a single
+    # argument that is an unnamed numpy array.
+    res = func(np.array(default_parameters))
+    assert (isinstance(res, float))
```
iaroslav-ai: Maybe add checks whether with unnamed dimensions / non - `Dimension` objects an exception rises. I imagine that in future we would want to extend your code to support also implicit dimension definitions with names, which could look like:
```python
(0.1, 1.0, 'log-uniform', 'foo')  # equivalent to Real(0.1, 1.0, prior='log-uniform', name='foo')
# currently we can only use
# (0.1, 1.0, 'log-uniform') 
# and then the dimension is unnamed
```
which might temper with these exceptions, and could lead to subtle bugs.  While such definition of dimension seems to be somewhat ugly at first glance, it is appealing in that it leads to a more compact code.

iaroslav-ai:
I guess @betatim is busy, so I will take a look.

iaroslav-ai:
I think this is a really cool feature, thanks @Hvass-Labs ! As for multi-objective optimization, I will take a look whether #369 interferes

iaroslav-ai:
Ah see starting from here

https://github.com/scikit-optimize/scikit-optimize/blob/80345a8fbcf2630b6ada9fc5a865a26193a05a50/skopt/optimizer/forest.py#L80

Some functions can return two values. 

Maybe just state `...and return the objective value.`

skopt/tests/test_utils.py
iaroslav-ai: (in case you need further info on various dimension definitions, see [here](https://github.com/scikit-optimize/scikit-optimize/blob/master/examples/hyperparameter-optimization.ipynb) or [here](https://github.com/scikit-optimize/scikit-optimize/blob/master/examples/sklearn-gridsearchcv-replacement.ipynb))

skopt/utils.py
iaroslav-ai: I agree with @glouppe. It would be better to state which pages exactly the user should visit where the helpful information is; Otherwise, every user who wants to understand the code will do the search from scratch, possibly ending on wrong pages. 

After some search, these two links seem to explain what is going on:

Explanation to `@wraps`:
 [https://docs.python.org/2/library/functools.html#functools.wraps](https://docs.python.org/2/library/functools.html#functools.wraps)

Explanation to decorator with parameters:
 [https://stackoverflow.com/questions/5929107/decorators-with-parameters](https://stackoverflow.com/questions/5929107/decorators-with-parameters)

iaroslav-ai:
Overall the PR looks good to me! Some of the comments can be fixed in different PR.

Hvass-Labs:
@iaroslav-ai I'm glad you like the feature and thanks for the constructive comments! I'll start working on them now. If you have the time then please review PR #579. I really would like these two PRs to be completed soon so I can finish my tutorial on skopt  and TensorFlow. And I think you guys would like to get rid of me as well :-)

skopt/utils.py
Hvass-Labs: Done.

skopt/utils.py
Hvass-Labs: Done.

skopt/tests/test_utils.py
Hvass-Labs: Hmm ... I have tested myself whether the exceptions are thrown and the new detailed error-messages are correct. I'm not sure a unit-test adds much "safety" to this? And I don't quite understand what subtle bugs might arise in the future?

Furthermore, regarding your short-hand notation I think it is actually quite problematic, both in terms of reading your code that uses skopt, such as this:

    space  = [(1, 5),                           # max_depth
              (10**-5, 10**0, "log-uniform"),   # learning_rate
              (1, n_features),                  # max_features
              (2, 100),                         # min_samples_split
              (1, 100)]                         # min_samples_leaf

In my opinion, it would be much more clear if you wrote:

    space  = [Int(1, 5, name='max_depth'),
              Int(10**-5, 10**0, prior="log-uniform", name='learning_rate'),
              Int(1, n_features, name='max_features'),
              Int(2, 100, name='min_samples_split'), 
              Int(1, 100, name='min_samples_leaf')] 

I can also see in your code that the short-hand notation causes problems other places and it requires numerous weird tests, etc.

You will need even more strange tests if you allow naming in the short-hand notation, because now you don't know whether the 3rd arg is the name of the dimension, or if it defines e.g. the prior-mapping. You can fix that with more weird tests, such as whether the 3rd string arg is one of the allowed prior-mappings, otherwise it is assumed to be the name. But what if the user has made a typo in the prior-mapping? ... Suddenly the API and code is very confusing and bloated. It's not going to be pretty! :-)

In my opinion, it is better to aim at making the code clear even if it means it becomes significantly more verbose. Having compactness of the code as a major goal is only really justified in performance-critical applications - in which case the code can be so hard to read, that extensive commenting is then needed.

skopt/utils.py
Hvass-Labs: I concede and have rewritten these doc-strings. However, I didn't like the links you found as I found them confusing :-) I managed to find a 7 minute youtube video that explains it quite well, so I added that link. I hope everyone is happy with this.

skopt/utils.py
Hvass-Labs: Done.

Hvass-Labs:
I have pushed the fixes @iaroslav-ai requested so I think everything is ready to be merged now.

Thanks to everyone for contributing to this!

skopt/tests/test_utils.py
iaroslav-ai: Yeah I also expect the short-hand definition of dimension with names to be somewhat complex :)
Short-hand seems to go nicely for quick prototyping with spaces in `BayesSearchCV`, but there actually one does not need named spaces, so yeah not sure whether we would need it in the end. Lets see what future users say.

iaroslav-ai:
LGTM!

iaroslav-ai:
Thank you @Hvass-Labs for this PR!

I think I speak for everyone when I say that we all appreciate your contribution and efforts in this PR. I am also looking forward to your tutorial on tensorflow + skopt, so let us know when that is made! I will take a look at another PR (soon).

:clinking_glasses: :tada:

skopt/tests/test_utils.py
betatim: I agree we should try as much as possible to gently persuade people to not use the shorthand notation (anymore), for example by only allowing you to do simple stuff through it. Along the lines of `explicit is better than implicit`.

Hvass-Labs: I don't think gentle persuasion is my thing so I'll leave that to you! :-)

Seriously though, I think it would be a wise decision to aim for crystal clear semantics of the API rather than having a million clever little features. I can see both in your code and your github threads that you are also confused about your code and its semantics, and I'm glad that you have now realized the confusion is 10x for newcomers :-)

Hvass-Labs:
@iaroslav-ai Thanks for the nice words and I'm glad that I was able to contribute a little to your fantastic project!

pavelkomarov:
This still does not address my issue (#583), because the `dimensions` parameter must still be a list. A list with parameters with names is fine, but even better (because it agrees more naturally with `BayesSearchCV`) is to let `dimensions` be a dictionary mapping parameter names to `Real` or other dimensions.

Hvass-Labs:
@pavelkomarov I tried following your arguments in #583 and here, but it would be easier to understand if you provided a small code snippet of what you want to do.

But I do agree that the semantics of defining dimensions and search-spaces could perhaps be a bit more clear. But the problem is that we need this for different things. Personally I would have been happy if dimensions *always* had to be named so the objective function would be called directly with named arguments, instead of using this decorator. I also do not like the scipy API and think skopt could have been far more elegant if it was not built around that. But I don't think there's much chance of making such a huge change to skopt, that's why I just added this decorator-hack, which works well enough for me.

Here's an example of what you would do now to define the search-space dimensions:

    dim1 = Real(name='foo', low=0.0, high=1.0)
    dim2 = Real(name=None, low=2.0, high=3.0)
    dim3 = Real(name='baz', low=4.0, high=5.0)

    dimensions = [dim1, dim2, dim3]

And you can then define your objective function as:

    @use_named_args(dimensions=dimensions)
    def func(foo, bar, baz):
        pass

Once again, I agree that it would have been more elegant to omit the decorator and have the framework automatically call `func` with named args. I had made another PR #590 which would instead call e.g. `gp_minimize(func, ..., use_arg_names=True)` so there was no need for the decorator, but @betatim was not too happy with adding another arg to all the optimizers.

Now the framework supports both the old scipy-style of calling `func` and using named args by just adding the decorator. I think it is a reasonable solution.

Why do you want to use a dict to define the dimensions? Would it look something like this?

    dimensions =
    {
        'foo': (0.0, 1.0),
        'bar': (2.0, 3.0),
        'baz': (4.0, 5.0)
    }

I prefer the more proper object-oriented style above. See also the discussion in https://github.com/scikit-optimize/scikit-optimize/pull/592#discussion_r160021709 why I prefer that way.

betatim:
In the world of scikit-learn and other machine-learning libs having a dictionary based definition is indeed natural and what people expect. Which is why `BayesSearchCV` let's you do that.

There are many applications for scikit-optmize beyond tuning hyper-parameters for ML models, with their preferred way of defining search spaces.

With the new decorator you can do the following, which I think is a good compromise between convenience for scikit-learn style parameter setting and the general purpose nature of `*_minimize()`:
```python
@use_named_args(dimensions)
def func(**kwargs):
  my_sklearn_clf.set_params(**kwargs)
  ..do stuff and measure performance..

```
If you name all the dimensions in your search space.

betatim:
> we should standardize the way spaces are defined.

I don't think that is possible. If you come from scikit-learn you prefer dictionary based definitions, if you come from `scipy` you prefer a difference way, if you come from `hyperopt` you prefer [yet another way](http://hyperopt.github.io/hyperopt/), and the list goes on. Given that you can arrive in scikit-optimize from these different places I think the best we can do is offer users a familiar interface for the part of skopt that is "for them". This makes it easy and familiar to get started.

If you then want to dig deeper, really tweak things, or have other power user/special needs scikit-optimize provides the `Optimizer` class which has all the "brains" in it plus the tools that are used to build all the different "interfaces" between scikit-learn, hyperopt, scipy, etc and `Optimizer`.

Think of scikit-optimize as Ikea furniture. You can put it together according to the spec from the shop, but you can also heavily modify it or put it together in new ways.

The fact that @pavelkomarov managed to quickly put together a copy of `gp_minimize` that worked for out of core learning is a plus point. It doesn't have to be part of the core library (where there are a lot of compromises that need making).

pavelkomarov:
"quickly" was a couple of days, so I'd still lobby to tweak base_minimize so it can support the dictionary form natively rather than requiring people make a list of dimensions with names = keys from their dictionary if they're coming from an sklearn frame of mind. But it's more useful for this to be concrete in a pull request.

PR #593, opened 2018-01-01
[MRG+1] Fix copy for ask and tell with computation time

chschroeder:
I just noticed the optimizer's `copy()` method tampers with the objective values `yi`. Only affects acquisition functions with computation time.

This is required for #574.

Short explanation:
* The values in `yi` are in this case tuples of `objective_value` and `time`
* Each copy would again apply the logarithm on the time values

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -402,6 +402,7 @@ class Optimizer(object):
         check_x_in_space(x, self.space)
         self._check_y_is_valid(x, y)
 
+        # take the logarithm of the computation times
         if "ps" in self.acq_func:
             if is_2Dlistlike(x):
                 y = [[val, log(t)] for (val, t) in y]
@@ -523,24 +524,22 @@ class Optimizer(object):
 
         if "ps" in self.acq_func:
             if is_2Dlistlike(x):
-                if np.ndim(y) == 2 and np.shape(y)[1] == 2:
-                    # TODO: refactor
-                    pass
-                else:
+                if not (np.ndim(y) == 2 and np.shape(y)[1] == 2):
                     raise TypeError("expcted y to be a list of (func_val, t)")
             elif is_listlike(x):
-                if np.ndim(y) == 1 and len(y) == 2:
-                    # TODO: refactor
-                    pass
-                else:
+                if not (np.ndim(y) == 1 and len(y) == 2):
                     raise TypeError("expected y to be (func_val, t)")
+
         # if y isn't a scalar it means we have been handed a batch of points
         elif is_listlike(y) and is_2Dlistlike(x):
-            # TODO: why no check here?
-            pass
+            for y_value in y:
+                if not isinstance(y_value, Number):
+                    raise ValueError("expected y to be a list a scalars")
+
         elif is_listlike(x):
             if not isinstance(y, Number):
                 raise ValueError("`func` should return a scalar")
+
         else:
             raise ValueError("Type of arguments `x` (%s) and `y` (%s) "
                              "not compatible." % (type(x), type(y)))
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -413,8 +413,11 @@ class Optimizer(object):
         return self._tell(x, y, fit=fit)
 
     def _tell(self, x, y, fit=True):
-        """Same as `tell()` but without checks and without any modification
-        of x or y."""
+        """Perform the actual work of incorporating one or more new points.
+        See `tell()` for the full description.
+
+        This method exists to give access to the internals of adding points
+        by side stepping all input validation and transformation."""
 
         if "ps" in self.acq_func:
             if is_2Dlistlike(x):
@@ -438,7 +441,7 @@ class Optimizer(object):
             raise ValueError("Type of arguments `x` (%s) and `y` (%s) "
                              "not compatible." % (type(x), type(y)))
 
-        # optimizer learned somethnig new - discard cache
+        # optimizer learned something new - discard cache
         self.cache_ = {}
 
         # after being "told" n_initial_points we switch from sampling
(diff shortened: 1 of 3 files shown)


chschroeder:
Travis seems to have connection issues. Other than that, this is ready for review.

betatim:
(restarted travis)

skopt/optimizer/optimizer.py
```diff
@@ -520,6 +520,31 @@ def tell(self, x, y, fit=True):
...
+                    raise TypeError("expcted y to be a list of (func_val, t)")
+            elif is_listlike(x):
+                if not (np.ndim(y) == 1 and len(y) == 2):
+                    raise TypeError("expected y to be (func_val, t)")
+
+        # if y isn't a scalar it means we have been handed a batch of points
+        elif is_listlike(y) and is_2Dlistlike(x):
+            for y_value in y:
+                if not isinstance(y_value, Number):
+                    raise ValueError("expected y to be a list a scalars")
```
betatim: -> "expected y to be a list _of_ scalars not %s." % y

Not sure if we should display `y`, my reaction to exceptions like this is usually: "but y is a damn list of scalars!!!11!" so having it shown to me helps. But it could be a huge list ??

skopt/optimizer/optimizer.py
```diff
@@ -520,6 +520,31 @@ def tell(self, x, y, fit=True):
         return create_result(self.Xi, self.yi, self.space, self.rng,
                              models=self.models)
 
+    def _check_y_is_valid(self, x, y):
+        """Checks if y is valid (with respect to x in some cases)."""
```
betatim: How could we reformulate it so that the doc string tells us something the function name doesn't already tell us?

Don't have a good suggestion right now for wording. Basically outline what kind of things we are checking ... without repeating the whole code. Maybe something like "Check if the shape and types of `x` and `y` are consistent."?

skopt/optimizer/optimizer.py
```diff
@@ -400,40 +400,40 @@ def tell(self, x, y, fit=True):
...
-                    raise TypeError("expcted y to be a list of (func_val, t)")
+                y = [[val, log(t)] for (val, t) in y]
+            elif is_listlike(x):
+                y = list(y)
+                y[1] = log(y[1])
+
+        return self._tell(x, y, fit=fit)
+
+    def _tell(self, x, y, fit=True):
+        """Same as `tell()` but without checks and without any modification
```
betatim: Maybe something like:
"""Perform the actual work of incorporating a new point

This method exists to give access to the internals of adding a new point by side stepping the input validation.
"""

It doesn't actually do the same as `tell`, it does all the hard work that `tell` used to do :)

betatim:
(can you put a `[MRG]` at the start of the title if ready for review? sklearn style PR titles :))

Let's see what travis says. I like the basic idea and the bug fix. It is always a bit tricky to see from the diff if code just got moved or also modified in the move. Could you highlight any points where you changed something as well as moved it (if any)? Otherwise  I think our tests should have this covered well enough, I like the way tests were edited and a new test was added. So if the robot overlords think we didn't break anything this should be good to go :)

betatim:
In the spirit of you not being able to read my mind: I really love your style of contributing!

skopt/optimizer/optimizer.py
chschroeder: How about "expected y to be a list of scalars but encountered 'z' at index 10"?

I thought about displaying the list partly (e.g. "[..., 1, 'z', 1,...]") but then the index of the invalid y_value is still unknown.

This is the check i added, there is a version for a single y value as well. If we change this, we should adapt the other error message as well.

skopt/optimizer/optimizer.py
chschroeder: I fully agree. This was added because i was asking myself at some point, why i need `x` to validate `y`.

"Check if the shape and types of x and y are consistent." sound good to me.

chschroeder:
Title is fixed :) Sorry, did not know the sklearn style titles.

The changes can be summarized as follows:
* Separate `old_tell()` into `tell()` and `_tell()` so that the transformations of the time values can be bypassed
  * `tell()` behaves the same as it previously did
  * `copy()` calls `_tell()` now (this is the actual bugfix)
  * the copy method's test was adjusted for an optimizer which incorporates computation time
* In order to make the previous possible i moved the log transformation of the time values into `tell()`
  * The existing code did three things at the same time:
    * Validate x / y
    * Take log(y) instead of y if time is incorporated
    * Update the internal x / y values (append)
  * Excluding the log transformation the remaining code still did two things
    * The validation was extracted to a separate function.
    * Thereby a noticed that one case had no validation. Fixed this and added a test.
    * Downside is of course that the if/else that identifies the cases is duplicated. I think this is fine at this point and makes the function more readable.

chschroeder:
Thanks for your kind words and for your help, i really appreciate it :)! This is a great project i am happy to contribute.

skopt/optimizer/optimizer.py
chschroeder: Or rather "shape*s* and types"?

chschroeder:
I just corrected the typo for "expected y to be a list of scalars".
There are four errors of this kind and if we treat them equally we should change all of them.
Except for that, i tried to incorporate all of your feedback.

(By the way, shouldn't ask and tell have their return values in the docstrings?)

chschroeder:
Regarding travis: "The job exceeded the maximum time limit for jobs, and has been terminated.".
No further knowledge about travis, but the tests seem to pass.

betatim:
> Regarding travis: "The job exceeded the maximum time limit for jobs, and has been terminated.".

Sorry for the slowness on this. There is a maximum time a travis job can run for, and we apparently exceeded that. We did a cleanup/refactor of our tests a while back to speed them up. It seems like we need to do that again :-/

For the moment I restarted travis in the hopes that they timed out because of "slow" CPUs.

betatim:
The alternative is that one of the changes in this PR causes the tests to run for a lot longer than usual. Though I can't see what would have done that.

betatim:
LGTM! @glouppe?

glouppe:
Thanks for your contribution!

chschroeder:
> Sorry for the slowness on this. There is a maximum time a travis job can run for, and we apparently exceeded that. We did a cleanup/refactor of our tests a while back to speed them up. It seems like we need to do that again :-/

No hurry :)! I just pointed out travis because i can't restart this on my own (without additional commits at least?). Thanks for restarting the tests.

PR #602, opened 2018-01-14
return_train_score default to False

fcharras:
Computing the train score on each split during hyper-parameter search can be a significant overhead, and is not used by the algorithm for model selection, hence it's only useful for occasional exploration purposes. Most of the time it's not really useful.

Sklearn is already planning to set the default to `False`, see the documentation for `return_train_score`: http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html

This PR implement the change for BayesSearchCV.

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -140,8 +140,8 @@ class BayesSearchCV(BaseSearchCV):
         FitFailedWarning is raised. This parameter does not affect the refit
         step, which will always raise the error.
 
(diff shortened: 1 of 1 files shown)


iaroslav-ai:
Lets wait for tests to pass (I mean should all pass but just to be sure :D ) and merge

iaroslav-ai:
All green, horray! Merging

PR #616, opened 2018-01-25
[MRG+1] Fix handling of Categorical only spaces

### LINKED ISSUE #615: TypeError: %d format: a number is required
Here is my example:

```python
    from skopt import Optimizer
    from skopt.utils import dimensions_aslist
    from skopt.space import Integer, Categorical, Real

    NN = {
        'activation': Categorical(['identity', 'logistic', 'tanh', 'relu']),
        'solver': Categorical(['adam', 'sgd', 'lbfgs']),
        'learning_rate': Categorical(['constant', 'invscaling', 'adaptive']),
        'hidden_layer_sizes': Categorical([(100,100)])
    }

    listified_space = dimensions_aslist(NN)
    acq_optimizer_kwargs = {'n_points': 20, 'n_restarts_optimizer': 5, 'n_jobs': 3}
    acq_func_kwargs = {'xi': 0.01, 'kappa': 1.96}

    optimizer = Optimizer(listified_space, base_estimator='gp', n_initial_points=10,
        acq_func='EI', acq_optimizer='auto', random_state=None,
        acq_optimizer_kwargs=acq_optimizer_kwargs, acq_func_kwargs=acq_fun_kwargs)

    rand_xs = []
    for n in range(10):
        rand_xs.append(optimizer.ask())

    rand_ys = [1,2,3,4,5,6,7,8,9,10]

    print rand_xs
    print rand_ys

    optimizer.tell(rand_xs, rand_ys)
```

Running with `acq_optimizer='lbfgs'` I was seeing `ValueError: The regressor <class 'skopt.learning.gaussian_process.gpr.GaussianProcessRegressor'> should run with acq_optimizer='sampling'.` But by tracing the Optimizer's `_check_arguments()` code I was able to figure out that my base_estimator must simply not have gradients in this Categoricals-only case. Changing to `acq_optimizer='auto'` solves that problem.

But now I see a new `TypeError: %d format: a number is required` thrown deep inside skopt/learning/gaussian_process/kernels.py. *If I print the transformed space before it gets passed to the Gaussian process, I see that it isn't really transformed at all: It still has strings in it!*

Adding a numerical dimension like `alpha: Real(0.0001, 0.001, prior='log-uniform')` causes the construction and the `.tell()` to succeed because the transformed space is then purely numerical.

*So the way purely-categorical spaces are transformed should be updated.*

Or there is an other possibility: Does it even make sense to try to do Bayesian optimization on a purely categorical space like this? Say I try setting (A,B), setting (A,C), setting (X,Z), and setting (Y,Z). For the sake of argument say (A,B) does better than (A,C) and (X,Z) does better than (Y,Z). Can we then suppose (X,B) will do better than (Y,C)? Who is to say (Y,C) isn't a super-combination or that the gains we seem to see from varying the first parameter to X or the second to B are unrelated? It seems potentially dangerous to reason this way, so perhaps the intent is that no Bayesian optimization should be possible in purely Categorical spaces. If this is the case, an error should be thrown early to say this. Furthermore, if this is correct, then how are point-values in Categorical dimensions decided during optimization? Are the numerical parameters optimized while Categoricals are selected at random?

It seems the previous paragraph should be wrong: If I try many examples with some parameter set to some value and observe a pattern of poor performance, I can update my beliefs to say "This is a bad setting". It shouldn't matter whether a parameter is Categorical or not; Bayesian optimization should be equally powerful in all cases.

betatim:
Fixes #615 

The original error comes because the creation of the message for the exception fails. The reason the exception was being raised in the first place was because we computed the number of dimensions for the `HammingKernel` incorrectly.

### DIFF
--- a/skopt/utils.py
+++ b/skopt/utils.py
@@ -315,6 +315,7 @@ def cook_estimator(base_estimator, space=None, **kwargs):
     if base_estimator == "GP":
         if space is not None:
             space = Space(space)
+            space = Space(normalize_dimensions(space.dimensions))
             n_dims = space.transformed_n_dims
             is_cat = space.is_categorical
 
@@ -558,7 +559,7 @@ def use_named_args(dimensions):
 
     Returns
     -------
-    * `wrapped_func` [callable] 
+    * `wrapped_func` [callable]
         Wrapped objective function.
     """
 
@@ -606,17 +607,17 @@ def use_named_args(dimensions):
             wrapped / decorated `func` is being called.
             It takes `x` as a single list of parameters and
             converts them to named arguments and calls `func` with them.
-            
+
             Parameters
             ----------
             * `x` [list]:
                 A single list of parameters e.g. `[123, 3.0, 'linear']`
                 which will be converted to named arguments and passed
                 to `func`.
-        
+
             Returns
             -------
-            * `objective_value` 
+            * `objective_value`
                 The objective value returned by `func`.
             """
 
--- a/skopt/tests/test_space.py
+++ b/skopt/tests/test_space.py
@@ -475,14 +475,6 @@ def test_dimension_name_none(dimension):
     assert dimension.name is None
 
 
-def test_dimension_name():
-    notnames = [1, 1., True]
-    for n in notnames:
-        with pytest.raises(ValueError) as exc:
-            real = Real(1, 2, name=n)
-            assert("Dimension's name must be either string or None." == exc.value.args[0])
-
-
 @pytest.mark.fast_test
 def test_space_from_yaml():
     with NamedTemporaryFile() as tmp:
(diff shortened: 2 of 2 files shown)


skopt/utils.py
```diff
@@ -315,6 +315,7 @@ def cook_estimator(base_estimator, space=None, **kwargs):
     if base_estimator == "GP":
         if space is not None:
             space = Space(space)
+            space = Space(normalize_dimensions(space.dimensions))
```
betatim: Having to do this is pretty ugly. The reason for doing this is that `normalize_dimensions` contains the logic for noticing that all dimensions are categorical and explicitly sets the transform to "identity". We should refactor this :-/

betatim:
After thinking a bit I think we need to restructure things so that there is a `check_space()` function. Similar to what [`check_array()`](http://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html) does. What makes this annoying is that what this function should do depends on what optimiser you use.

pavelkomarov:
Can we get this merged and close #615 ?

iaroslav-ai:
I merge in few hours if no other comments will appear.

PR #627, opened 2018-02-01
[MRG] added n_jobs support to optimizer

jonathanng:


jonathanng:
I find that for large # of datapoints, GP is very slow. This pull request does not fix that. Anyone else seeing this issue?

betatim:
I think the fact that GPs slow down as the number of data points increases is one of the drawbacks of GPs. I tried to find the relevant bit of the scikit-learn docs that explains the scaling behaviour but I couldn't find it.

iaroslav-ai:
Yup a normal thing for GP:

[https://arxiv.org/pdf/1502.05700.pdf](https://arxiv.org/pdf/1502.05700.pdf) see Figure 1 there

@jonathanng maybe add this somewhere to the comments if you have time? Otherwise we could add it later

iaroslav-ai:
I still think having a control over n_jobs would be very beneficial, esp. if you run multiple optimizations at the same time in order to test for example the algorithm

skopt/optimizer/optimizer.py
```diff
@@ -136,7 +136,7 @@ class Optimizer(object):
     """
     def __init__(self, dimensions, base_estimator="gp",
                  n_random_starts=None, n_initial_points=10,
-                 acq_func="gp_hedge",
+                 n_jobs=1, acq_func="gp_hedge",
```
iaroslav-ai: Please add description of `n_jobs` to docstrings.

iaroslav-ai:
Beyond the `n_jobs` docs thingy LGTM! Please add [MRG] to the beginning of the name of the PR if you are ready to merge, or [WIP] if you wish to work a bit more on this.

skopt/optimizer/optimizer.py
glouppe: +1

PR #629, opened 2018-02-04
fixed partial_dependence to match prior

jonathanng:
Currently `partial_dependence` uses linear spacing to approximate the partial_dependence. For spaces that are log-uniform, this is problematic. In addition, `partial_dependence` does not work with categories.

In this PR, I refactored out the code, and allowed the space prior to determine what kind of spacing is used. In addition, I added support for categorical dimensions and a `return_std` flag that will help with plotting if a confidence interval is needed.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -1,12 +1,13 @@
 """Plotting functions."""
 
-import numpy as np
+from skopt.space import Categorical
+from skopt.space import Integer
 
+import numpy as np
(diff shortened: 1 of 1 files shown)


betatim:
I think I know what you are doing here, but for people from the future who are digging through old PRs could you write a few sentences in the top comment of this PR explaining what the problem is, how you are fixing it and why ;)

jonathanng:
@betatim, just updated it!

Issue #623, opened 2018-01-31
Evaluate one point with all threads in BayesSearchCV 


rene-rex:
Hello

I am using BayesSearchCV to find hyper parameters for an expensive ML problem. However, I don't get why `n_jobs` is at the same time the number of points to asked from the optimizer and the number of parallel threads in the joblib call. Is it possible to have something like `n_points=1, n_jobs=10, cv=10`? Thus, only one point is evaluated each step and the CV uses all threads.

iaroslav-ai:
Thanks @rene-rex for pointing this out! That would be a nice feature to have a finer control over the parallelism. This is kinda on the edge of being a bug even. 

@rene-rex if you would open a PR with `n_points` argument that would be appreciated :) Alternatively, I can take a look later.

rene-rex:
@iaroslav-ai Can do! I will open a PR this week. However, I may need some help regarding proper tests.

iaroslav-ai:
Thanks again @rene-rex  :) Forgot to close this issue.

PR #630, opened 2018-02-06
[MRG+1] separate n_points from n_jobs

rene-rex:
This was easier than I thought. Implement issue #623

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -64,9 +64,11 @@ class BayesSearchCV(BaseSearchCV):
         is a number of iterations that will be spent optimizing over
         this subspace.
 
-    n_iter : int, default=128
+    n_iter : int, default=50
         Number of parameter settings that are sampled. n_iter trades
-        off runtime vs quality of the solution.
+        off runtime vs quality of the solution. Consider increasing
+        ``n_points`` if you want to try more parameter settings in
+        parallel.
 
     optimizer_kwargs : dict, optional
         Dict of arguments passed to :class:`Optimizer`.  For example,
@@ -83,7 +85,13 @@ class BayesSearchCV(BaseSearchCV):
         Parameters to pass to the fit method.
 
     n_jobs : int, default=1
-        Number of jobs to run in parallel.
+        Number of jobs to run in parallel. At maximum there are
+        ``n_points`` times ``cv`` jobs available during each iteration.
+
+    n_points : int, default=1
+        Number of parameter settings to sample in parallel. If this does
+        not align with ``n_iter``, the last iteration will sample less
+        points. See also :func:`~Optimizer.ask`
 
     pre_dispatch : int, or string, optional
         Controls the number of jobs that get dispatched during parallel
@@ -638,9 +646,8 @@ class BayesSearchCV(BaseSearchCV):
 
             # do the optimization for particular search space
             while n_iter > 0:
-                # when n_iter < n_jobs points left for evaluation
+                # when n_iter < n_points points left for evaluation
                 n_points_adjusted = min(n_iter, n_points)
-                print(n_points_adjusted)
 
                 optim_result = self._step(
                     X, y, search_space, optimizer,
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -271,12 +271,13 @@ class BayesSearchCV(BaseSearchCV):
 
     def __init__(self, estimator, search_spaces, optimizer_kwargs=None,
                  n_iter=50, scoring=None, fit_params=None, n_jobs=1,
-                 iid=True, refit=True, cv=None, verbose=0,
+                 n_points=1, iid=True, refit=True, cv=None, verbose=0,
                  pre_dispatch='2*n_jobs', random_state=None,
                  error_score='raise', return_train_score=False):
 
         self.search_spaces = search_spaces
         self.n_iter = n_iter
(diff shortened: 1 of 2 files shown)


betatim:
Good work!

Could you add a test for the specific scenario you had in mind with `n_points=1, n_jobs=10, cv=10`? I think it is worth testing that this does actually run "in parallel". Maybe use `n_jobs=2`.

skopt/searchcv.py
```diff
@@ -271,12 +271,13 @@ class BayesSearchCV(BaseSearchCV):
 
     def __init__(self, estimator, search_spaces, optimizer_kwargs=None,
                  n_iter=50, scoring=None, fit_params=None, n_jobs=1,
-                 iid=True, refit=True, cv=None, verbose=0,
+                 n_points=1, iid=True, refit=True, cv=None, verbose=0,
```
betatim: Can you add `n_points` to the doc string and modify the `n_jobs` docs so that people know what they both do and how they interact?

skopt/searchcv.py
```diff
@@ -642,13 +639,14 @@ def fit(self, X, y=None, groups=None, callback=None):
             # do the optimization for particular search space
             while n_iter > 0:
                 # when n_iter < n_jobs points left for evaluation
-                n_jobs_adjusted = min(n_iter, n_jobs)
+                n_points_adjusted = min(n_iter, n_points)
+                print(n_points_adjusted)
```
betatim: Spurious `print` statement

skopt/searchcv.py
```diff
@@ -642,13 +639,14 @@ def fit(self, X, y=None, groups=None, callback=None):
             # do the optimization for particular search space
             while n_iter > 0:
                 # when n_iter < n_jobs points left for evaluation
```
betatim: The comment needs updating

rene-rex:
I addressed all your comments and also took the liberty to refactor the tests a bit.

rene-rex:
It seems Travis cant you use more than one thread and thus the timing test for actual parallelization fails. However, I don't known an other way to test if two threads are used. Any ideas?

rene-rex:
I screwed up a test... working on it.

skopt/tests/test_searchcv.py
```diff
@@ -78,6 +83,22 @@ def test_searchcv_runs(surrogate, n_jobs, n_points):
...
+    """
+
+    start_time = time.time()
+    test_searchcv_runs(surrogate='gp', n_jobs=1, n_points=1, cv=10)
+    time_one_job = start_time - time.time()
+    start_time = time.time()
+    test_searchcv_runs(surrogate='gp', n_jobs=2, n_points=1, cv=10)
+    time_two_jobs = start_time - time.time()
+
+    assert_greater(time_one_job, time_two_jobs, 'Two jobs are not faster than one job')
```
betatim: 😄

Let's see how often this gives a false alarm, then we might have to think of a better way of detecting two jobs vs one job.

skopt/tests/test_searchcv.py
```diff
@@ -78,6 +83,22 @@ def test_searchcv_runs(surrogate, n_jobs, n_points):
...
+    """
+    Test whether two parallel jobs acutally work faster than
+    one job.
+    """
+
+    start_time = time.time()
+    test_searchcv_runs(surrogate='gp', n_jobs=1, n_points=1, cv=10)
+    time_one_job = start_time - time.time()
+    start_time = time.time()
+    test_searchcv_runs(surrogate='gp', n_jobs=2, n_points=1, cv=10)
```
betatim: Does anyone have any experience with calling a `test_` function from another one? I've not seen it done before and wondering if it will confuse pytest/coverage reporting. If no one has any experience then let's leave it like this and find out :)

rene-rex: I do this in one of my other projects over and and over - works like a charm!

rene-rex:
`test_parallel_cv` failed 1 out of 3 times... I guess we have to come up with something better. However, I have no ideas right now.

skopt/tests/test_searchcv.py
rene-rex: I added a guard to skip the test if there is only one CPU. Apparently, it doesn't help much.

rene-rex:
@betatim Since the timing test is too brittle, I removed it but the parameter combination is still tested. Do you think the code is ready to be merged?

skopt/tests/test_searchcv.py
```diff
@@ -77,6 +108,16 @@ def test_searchcv_runs(surrogate, n_jobs):
     assert_greater(opt.score(X_test, y_test), 0.9)
 
 
+def test_parallel_cv():
```
iaroslav-ai: This takes a few seconds to run. Please mark with `@pytest.mark.slow_test`.

rene-rex: @iaroslav-ai Done!

rene-rex:
Are we ready to merge?

iaroslav-ai:
LGTM!

iaroslav-ai:
Lets say that we give it a day for others to comment if there is anything.

iaroslav-ai:
Merging

iaroslav-ai:
Thank you @rene-rex ! :tada:

rene-rex:
😃

PR #639, opened 2018-02-19
[MRG+1] Deal with matplotlib being broken on some OSX systems.

### LINKED ISSUE #637: import of matplotlib causes error on Mac OS X (depending on backend)
Since skopt import matplotlib for its plotting, it runs into some known issues with Mac OSX
```
    from matplotlib.backends import _macosx
RuntimeError: Python is not installed as a framework. The Mac OS X backend will not be able to function correctly if Python is not installed as a framework. See the Python documentation for more information on installing Python as a framework on Mac OS X. Please either reinstall Python as a framework, or try one of the other backends. If you are using (Ana)Conda please install python.app and replace the use of 'python' with 'pythonw'. See 'Working with Matplotlib on OSX' in the Matplotlib FAQ for more information.
```

this can be fixed by 

```
import matplotlib
matplotlib.use('PS')
```

before importing `matplotlib.pyplot`

but it would be nice if skopt could import such that it ignores those plotting-related issues for cases when one is not interested in plotting

betatim:
Fixes #637 

I am not sure what I think of this. On the one hand it feels like it is the user's problem if they can't setup `matplotlib` properly, on the other hand there are plenty of use-cases for skopt without plotting.

Generally we should try and find out how other libraries deal with this.

### DIFF
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -32,7 +32,7 @@ popd
 conda create -n testenv --yes python pip pytest nose
 source activate testenv
 
-pip install -e.
+pip install -e .[plots]
 export SKOPT_HOME=$(pwd)
 
 conda install --yes jupyter
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -32,7 +32,7 @@ popd
 conda create -n testenv --yes python pip pytest nose
 source activate testenv
 
-pip install -e .[plots]
+pip install -e '.[plots]'
 export SKOPT_HOME=$(pwd)
 
 conda install --yes jupyter
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -32,7 +32,7 @@ if [[ "$COVERAGE" == "true" ]]; then
     pip install pytest-cov coverage coveralls
 fi
 
-pip install -e.
+pip install -e .[plots]
 python --version
 python -c "import numpy; print('numpy %s' % numpy.__version__)"
 python -c "import scipy; print('scipy %s' % scipy.__version__)"
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -32,7 +32,7 @@ if [[ "$COVERAGE" == "true" ]]; then
     pip install pytest-cov coverage coveralls
 fi
 
-pip install -e .[plots]
+pip install -e '.[plots]'
 python --version
 python -c "import numpy; print('numpy %s' % numpy.__version__)"
 python -c "import scipy; print('scipy %s' % scipy.__version__)"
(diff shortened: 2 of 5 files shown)


iaroslav-ai:
Now that I think about this I needed to resolve matplotlib dependencies on my computing nodes where I do not really need it. So +1 from me.

betatim:
I think this would still lead the an exception if you tried to run skopt on a machine without matplotlib. If we want to support `matpltlib` as a optional dependency we should do something else.

There are a few tools that do "delayed" imports but they all appear like hacks and would require us to add a new dependency.

The only alternative I can think of is that `skopt.plots` isn't imported in `__init__.py` and you only get it by explicitly importing it.

In pandas the imports are done inside a plotting class that the data frame only instantiates when needed. Not sure that maps well to our free standing functions.

glouppe:
An easy fix would be to move the imports to matplotlib inside the `plot_*` functions.

iaroslav-ai:
I would be +1 on the idea of @glouppe .

betatim:
Didn't read @glouppe's comment till just now. Instead I removed the import of `plots.py` from `__init__.py` which doesn't seem to break anything (`import skopt.plots` `from skopt import plots` still work).

iaroslav-ai:
I like it. What do you think about adding a test where the plots module is imported? I think the machine image used for tests includes already matplotlib. Might be nice to have it to make sure importing still works, in case we decide to change how matplotlib is imported in future for any reason.

betatim:
Not sure I know a good way to do that/will get around to adding the test. We'd need to add a new entry in the build matrix (I think).

build_tools/circle/install.sh
```diff
@@ -32,7 +32,7 @@ popd
 conda create -n testenv --yes python pip pytest nose
 source activate testenv
 
-pip install -e.
+pip install -e .[plots]
```
carlosdanielcsantos: This doesn't work, needs single quotes:

`pip install -e '.[plots]'`

build_tools/travis/install.sh
```diff
@@ -32,7 +32,7 @@ if [[ "$COVERAGE" == "true" ]]; then
     pip install pytest-cov coverage coveralls
 fi
 
-pip install -e.
+pip install -e .[plots]
```
carlosdanielcsantos: Same here

build_tools/circle/install.sh
betatim: 👍

build_tools/travis/install.sh
betatim: 👍

carlosdanielcsantos:
👍 should we merge this?

iaroslav-ai:
LGTM :+1:

betatim:
Thanks for the patience and working together spirit @carlosdanielcsantos!

betatim:
For my education, how did you merge this PR @glouppe? By rebase or something? See the [discussion here](https://github.com/scikit-optimize/scikit-optimize/pull/649#issuecomment-374725496) for the source of my confusion :)

glouppe:
Rebase and merge, using the github button.

betatim:
That explains why there is no merge commit for me to find :)

Issue #640, opened 2018-02-26
Migrate the docs to sphinx


chschroeder:
This is an umbrella issue to track the status of migrating the docs to sphinx. Originally started in #431 by @betatim, we will try now to split this in multiple smaller PRs and break as little as possible in the meantime.

**Scope**
* [ ] *Tooling and prerequisites* (#641): Cherry pick as much as possible from betatim's existing setup.
  * [x] Building the documentation and creating a html version should behave correctly.
  * [x] Include the napoleon plugin so that we can stay with the Numpy docstring style.
  * [ ] Have the first examples show up correctly in the new sphinx docs (to verify that everything is set up correctly).
* [ ] *Migrate the existing docs*
  * *TODO: Divide the task in reasonable chunks and keep track of them here* 
* [ ] Misc
  * [ ] Add citing information (Zenodo DOI)
  * [ ] Update copyright year information (currently "(c) 2017")

glouppe:
Thanks for working on this :)

chschroeder:
No need to thank me, i didn't do much yet ;).

holgern:
Done

PR #641, opened 2018-02-26
[WIP] Sphinxing

chschroeder:
First part of #640. 

**Scope**
* [x] *Tooling and prerequisites*: Cherry pick as much as possible from betatim's existing setup.
  * [x] Building the documentation and creating a html version should behave correctly.
  * [x] Include the napoleon plugin so that we can stay with the Numpy docstring style.
  * [x] Have the first examples show up correctly in the new sphinx docs (to verify that everything is set up correctly). (*)

(*): Restrict this to the section `Top level minimization functions`, mainly `dummy_minimize`.

Update:
Almost finished. I need to look over all the changes again, then i set it to MRG. All aesthetical, content-related or structural changes are not yet done and will be processed in further PRs.


docs/conf.py
```diff
@@ -61,9 +61,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.4'
+version = '0.5'
```
chschroeder: Can we extract this variable to somewhere? Easy place to miss. At least it must be added to the release checklist.

chschroeder:
I thought about changing the circleci setup so that we can build the old and new docs in parallel. In the end i decided it's probably not worth the hassle, and since this does not really break anything i'd rather invest the time in the actual sphinx migration.

chschroeder:
Currently testing `nbsphinx`. It seems to have problems whenever the notebooks are not in a subdirectory of the docs directory. (Short: `scikit-optimize/examples` causes problems, `scikit-optimize/docs/examples` works fine.) Any ideas / opinions about this?

betatim:
I ran into the problem with `nbsphinx` as well and didn't find a solution that worked except for moving things to `docs/examples/`. I'd do that. Maybe we can make `scikit-optimize/examples/` a soft-link to `scikit-optimize/docs/examples/` to send visitors to the new location.

betatim:
I spent a few hours yesterday investigating [`lib2to3`](https://docs.python.org/2/library/2to3.html) to see if we can reuse its infrastructure to automatically rewrite doc strings from markdown to rST. I think the answer is yes, but it might be one of these projects where you spend more time building a tool than it would take to do it manually. I'll keep people posted.

👍  to not spending time trying to build the docs in parallel.

Can you see the "artefacts" that Circle CI produces? For example https://1096-54340642-gh.circle-artifacts.com/0/tmp/circle-artifacts.pVMZdSQ/index.html

chschroeder:
Thanks for sharing your thoughts and experiences! Now i feel better moving the examples.
How will a soft-link behave for Windows users? I thought of that at first too at first, but discarded it due to that.

I can't see the artifacts on circleci. The link you posted, however, works for me.

betatim:
> I can't see the artifacts on circleci. The link you posted, however, works for me.

Ok, that is a bit annoying. You will have to keep building the docs locally to see if it works :(

> soft-link on Windows

I have no idea what happens on Windows with soft-links. Do we know someone who has access to a windows machine?

betatim:
From a quick google it seems that symlinks aren't a thing at all on windows. My next best idea would be to place a README.md in the examples directory containing a short explanation where the examples have moved to and a link to them.

chschroeder:
More complicated than i expected. I just realized this needs to work for python 2 and 3. Will update my progress later on.

chschroeder:
The basic stuff should be working now. Obviously the resulting documentation is everything but complete. I refrained myself from making changes to the structure, contents or layout of the documentation.

(There are warning during the build of the docs because some parts are references but still have the old docstring. These warning will disappear when the rest is migrated.)

chschroeder:
Last thought, when we merge this too early, will the current state of the docs be deployed to https://scikit-optimize.github.io/ ?

Then we should probably keep this at a separate branch.

glouppe:
We could disable for the time being the automatic deployment of the doc, maybe?

betatim:
Not sure about disabling doc deployment.

We already have parts of the docs which are "broken" in terms of layout because we use a mixture of markdown and reST in the doc strings.

If we auto-deploy this we end up with partially broken docs which annoys users.

If we do not auto-deploy this we will forget about it and end up with out-of-date docs.

Neither solution seems particularly good :-/

betatim:
This is what the docs currently look like when using this PR: https://1104-54340642-gh.circle-artifacts.com/0/tmp/circle-artifacts.wqEdI82/index.html

glouppe:
That's not bad to be fair. We should populate the left menu a bit, e.g. by having direct links to the `minimize` functions and find a way to add back the notebooks.

betatim:
Let's move the example notebooks [as discussed](https://github.com/scikit-optimize/scikit-optimize/pull/641#issuecomment-370320067).

chschroeder:
We could also do the work on a separate branch (with auto deploy disabled). There haven't been that many merges during the last few days, so i think we could keep up with the changes in the master branch.

chschroeder:
The notebooks' files have already been moved. I only added one as an example to the docs, the rest can be easily added as well. Still, everything else would be missing in these docs.

chschroeder:
By the way, did you change the output of https://scikit-optimize.github.io? This style from pdocs is really nice. As far as i remember, i only saw the sphinx alabaster style at the docs page before.

chschroeder:
I reincluded the notebooks (i.e. linked to them in the docs). I had to change some heading levels because they were displayed as top-level items (# -> ##). 

@betatim For the "Strategy comparison" notebook, there was a different top-level heading and sphinx took that instead of the label. I changed it to match the notebook's filename for now. This is your notebook, do you agree here? Or should this be solved differently.

betatim:
I am -1 on any kind of parallel branch strategy. In my experience this almost always leads to a lot of pain when you do try to merge.

I think if we can make a issue that lists the things that need to be done "quickly" after we merge this to get to a good state and each of us tackles one of them we can make the switch fairly quickly without having to have a parallel universe. We could even start on the things that are missing in new PRs now (building on this branch and then rebasing once it is merged).

docs/index.rst
```diff
@@ -0,0 +1,49 @@
...
+.. toctree::
+    :maxdepth: 2
+    :caption: Contents:
+
+    intro
+    api/index
+
+.. toctree::
+    :maxdepth: 2
+    :caption: Notebooks:
```
betatim: -> Example notebooks? Or it is just me who thinks it isn't obvious that these are examples?

docs/index.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. toctree::
+    :maxdepth: 2
+    :caption: Contents:
+
+    intro
+    api/index
+
+.. toctree::
+    :maxdepth: 2
```
betatim: Do you think it is helpful to have the subheadings of the notebooks? I'd change it to maxdepth 1 and just show the top level heading.

docs/intro.rst
```diff
@@ -0,0 +1,21 @@
...
+      return (np.sin(5 * x[0]) * (1 - np.tanh(x[0] ** 2)) *
+              np.random.randn() * 0.1)
+
+  res = gp_minimize(f, [(-2.0, 2.0)])
+
+For more read our `introduction to bayesian optimization`_ and the other
+`examples`_.
+
+.. _introduction to bayesian optimization: https://scikit-optimize.github.io/notebooks/bayesian-optimization.html
+.. _examples: https://github.com/scikit-optimize/scikit-optimize/tree/master/examples
```
betatim: Need to update the target link to the new notebook

docs/api/minimize_functions.rst
```diff
@@ -0,0 +1,11 @@
...
+
+Top level minimization functions
+================================
+
+Easy to get started with these functions. They mirror the ``scipy.optimize``
+API.
+
+.. autofunction:: skopt.dummy_minimize
+
+.. autofunction:: skopt.gp_minimize
```
betatim: Add these two as well?
```
.. autofunction:: skopt.forest_minimize
.. autofunction:: skopt.gbrt_minimize
```

docs/index.rst
chschroeder: Fine with me. I took the heading from https://scikit-optimize.github.io, other than that there was no specific reason for the wording.

chschroeder:
Thanks for the review! If i understand correctly you would prefer if this PR was somewhat more complete. I have no strong opinion here, so i can live with this either way. I tried to keep it as small as possible for easier reviewing, but i understand the argument against branching as well.

To do for this task:
* Adopt the structure from the old docs (https://scikit-optimize.github.io)
* Check / Update links to the notebooks

Besides, it would be easier for me to know, what is specifically excluded here, i.e. what do we need to do in the next PR.

* Correct the markup of existing docs.
* Styling? (I think in terms of visuals, the current sphinx style is not on par with the old docs :P). Do you agree here?

docs/intro.rst
chschroeder: Good catch! I totally missed this one.

betatim:
I think the effort you put into this PR as it is is already huge! Last time I tried to get this switch made I wanted to convert "all the docs" in one go (and failed). So this time I think we should get "a large chunk" changed and then merge it. Then make more small PRs to fix up the things that are broken. The downside is that the docs will be partly broken. If there are other ideas please mention them :)

What do you mean with "adopt structure of old docs"? What is where and TOC entry wise?

chschroeder:
Alright, then let's finish as much as possible here. I was just afraid to produce too many changes at once.

I don't see any other options where we would not break anything (except a big merge that unrealistically fixes everything).

With "structure" i was referring to the TOC. Basically, every page that has been present before should be visible and reachable (regardless of some broken markup).

chschroeder:
In the meanwhile i got the hang of most of the sphinx stuff. I am struggling to insert some special links (e.g. Module Index) in the Navigation. Besides, the structure of most of the sub pages is too cluttered as it is right now. Put it back to WIP for now.

Somewhat offtopic:
I really like the style from pdoc (https://github.com/BurntSushi/pdoc) that you chose for the current docs. Unfortunately, it does not seem there is a sphinx template for this.

Would you be interested in having this as a sphinx theme? Not sure to what extent this is possible (not familiar with the limits of sphinx templates) though.

chschroeder:
This has stalled right now. Unfortunately i have neither had much time lately, nor was i happy with the resulting docs (produced in this branch).

I haven't gotten a reply, but i think we need a suitable sphinx theme in the long run.
This is probably focusing at the wrong place, yet i feel regardless what i change at the docs now, it just looks "unfinished" and not as good as the current docs do.

holgern:
Sphinx migration will continue in PR #811 , thanks for your work.

PR #645, opened 2018-03-12
Sets matplotlib as optional requirement through 'plotting'

### LINKED ISSUE #637: import of matplotlib causes error on Mac OS X (depending on backend)

carlosdanielcsantos:
This PR tries to deal with the issues discussed in #637 by doing the following:

- matplotlib becomes an optional requirement:
    - `pip install scikit-optimize` **will not** install matplotlib
    - `pip install 'scikit-optimize[plotting]'` **will** install matplotlib

- matplotlib functions are only imported at use time (e.g., when trying to plot convergence). If matplotlib is not present, the import error will only be raised when trying to plot things.

Closes #637

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ dist/
 
 # vim users
 .*.swp
+
+# pytest
+.pytest_cache
--- a/build_tools/circle/install.sh
+++ b/build_tools/circle/install.sh
@@ -32,7 +32,7 @@ popd
 conda create -n testenv --yes python pip pytest nose
 source activate testenv
 
-pip install -e.
+pip install -e '.[plotting]'
 export SKOPT_HOME=$(pwd)
 
 conda install --yes jupyter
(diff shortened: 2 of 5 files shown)


betatim:
(This is a duplicate of #639, let's use #637 to resolve that.)

glouppe:
No strong opinion between #639 and this PR. Whatever works is good for me.

PR #647, opened 2018-03-13
[MRG+2] Dimensions supporting boolean np.array

### LINKED ISSUE #646: Dimensions: boolean np.array not supported
Despite the fact that numpy.array are supported types for dimensions, the built-in Python bool and the numpy.bool_ have very different semantics from each other (check [this](https://github.com/numpy/numpy/issues/9646) numpy issue), meaning `isinstance(np.array([True])[0], bool)` returns `False` as the bool value is implicitly casted to `numpy.bool_`

For this reason, `any([isinstance(d, (str, bool)) for d in dimension])` always returns `False` if dimension is a np.array with boolean values, which makes skopt's `check_dimensions` function raise a ValueError with the message "Invalid dimension {}. Read the documentation for supported types."

carlosdanielcsantos:
Closes #646

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -78,7 +78,8 @@ def check_dimension(dimension, transform=None):
         return Categorical(dimension, transform=transform)
 
     if len(dimension) == 2:
-        if any([isinstance(d, (str, bool)) for d in dimension]):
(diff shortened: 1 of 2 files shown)


glouppe:
LGTM. +1

iaroslav-ai:
LGTM. Thanks for this!

PR #649, opened 2018-03-16
Releasing v0.5.2

carlosdanielcsantos:


betatim:
Not sure, I think we should add the changes so far under a new heading so that we have them available for 0.6.

Let's see if others have opinions (on Monday) as so far we didn't really do small releases like this. I think I like the idea though to try and step up the release velocity of scikit-optimize.

carlosdanielcsantos:
In our internal libraries, we release every version (including minors) with a changelog. And because the author that introduces a change is usually the most qualified person to write their PR changelog, we make updating the changelog a mandatory to-do for a PR to be merged. The downside of this is that the changelog file will often be conflicting with the base branch and we usually need to rebase before merging.

betatim:
Opinions @glouppe or @iaroslav-ai ?

glouppe:
Since this one is ready, we could make a release. However, in general, I would vote against making micro-releases every time we change two lines of code. We should wait to have some content in-between releases.

carlosdanielcsantos:
So, what about this https://github.com/scikit-optimize/scikit-optimize#making-a-release?

From this one can infer that contributors are welcome to make releases whenever they need (which is, in fact, the case here).

glouppe:
To be fair, I am not sure we were really expecting non-admin to propose releases, but I am glad to be proven wrong :) The README should be improved to make this point clear.

That being said, we could certainly do with more frequent releases. We just need to find the right balance.

betatim:
Let's do it!

@carlosdanielcsantos are you planning on adding changelog entries before I cut this release?

---

All people are welcome to propose changes (including a new release!). Most proposals will require a conversation to decide if it should be accepted, modified, rejected, etc Different people will have different needs and so the most frequent outcome will be a compromise :)

carlosdanielcsantos:
@betatim I can do it if you think it's worth it

betatim:
```
$ git log --merges --first-parent master --pretty=format:"%h %<(10,trunc)%an %<(15)%ar %s" v0.5.1.. 
b748b47 Iaroslav.. 7 days ago      Merge pull request #647 from carlosdanielcsantos/dimensions_bool_nparray
e67d0b0 Iaroslav.. 2 weeks ago     Merge pull request #630 from rene-rex/evaluate-one-point-with-all-threads
```

So I'd add those two and #639 (not sure why it isn't in the git log??)

carlosdanielcsantos:
It seems that one was rebased and merged with fast-forward. Ok, adding those three!

carlosdanielcsantos:
I guess we should also update the README with instructions on how to install skopt with matplotlib, right?

iaroslav-ai:
@carlosdanielcsantos thanks for this! If you could update README explaining now optional plots installation that would be splendid.

iaroslav-ai:
My opinion on this: I like the idea of releasing more frequently; For example with BayesSearchCV, we had a few issues from users just because it was not released. 

One idea could be to make release after substantial PR's, and ask author / authors of PR to make release. This way the most qualified person would write a changelog. 

I would say to release changelog with every new release. This way a history of changes over minor releases is stored in a more explicit manner.

carlosdanielcsantos:
@iaroslav-ai , done. For me it's good to be merged, but feel free to adjust the README to taste.

iaroslav-ai:
lgtm!

iaroslav-ai:
Thank you @carlosdanielcsantos ! :tada: :smiley: 

The draft release is: https://github.com/scikit-optimize/scikit-optimize/releases/tag/untagged-3d4a04d0e0ff63e4e4ed 

I will release this tomorrow if no one has any comments on this.

carlosdanielcsantos:
Thank you very much!

betatim:
Made the release public.

Issue #651, opened 2018-03-30
Allow space to be configured via yaml file


Scott-Graham-Bose:
Would it be of value to support reading in the space configuration via a yaml file?

what I'm imagining would have a yaml file something like this:
```    
- Real:
       low: 0.0001
       high: 1
       prior: log-uniform
       name: learning_rate
- Categorical:
       categories: [8, 16, 32]
       name: batch_size
```

and then being able to use it like this:
```
from skopt.utils import use_named_args

@use_named_args('path/to/config.yml')
```

or potentially with a separate method
``` 
from skopt.utils import load_space

space = load_space('path/to/config.yml')
```

happy to contribute if this is worthwhile and would fit in to common use cases beyond my own.

betatim:
My thoughts: why not? It would make it easier to share definitions of spaces, especially for common things to optimise like scikit-learn classifiers.

I'd maybe implement it as `a_space = Space.from_yaml(path)`, so a class method for `Space`.

iaroslav-ai:
I believe this is solved now - feel free to object :)
Thank you @Scott-Graham-Bose for  your contribution!

PR #655, opened 2018-04-05
[MRG+1] Adding ability to define Space parameters via yaml file

gramhagen:
Per discussion in #651 

added classmethod in Space.from_yaml() to allow for external file to define Space parameters
added unittest in test_space.py

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -1,7 +1,6 @@
 import numbers
 import numpy as np
 import yaml
-import sys
 
 from scipy.stats.distributions import randint
 from scipy.stats.distributions import rv_discrete
@@ -559,9 +558,22 @@ class Space(object):
         Parameters
         ----------
         * `yml_path` [str]:
-           Full path to yaml configuration file
+            Full path to yaml configuration file, example YaML below:
+            Space:
+              - Integer:
+                  low: -5
+                  high: 5
+              - Categorical:
+                  categories:
+                  - a
+                  - b
+              - Real:
+                  low: 1.0
+                  high: 5.0
+                  prior: log-uniform
         * `namespace` [str, default=None]:
-           Namespace within configuration file to use
+           Namespace within configuration file to use, will use first
+             namespace if not provided
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


skopt/space/space.py
```diff
@@ -550,6 +552,49 @@ def is_real(self):
...
+        -------
+        * `space` [Space]:
+           Instantiated Space object
+        """
+        with open(yml_path, 'rb') as f:
+            cfg = yaml.load(f)
+
+        this_module = sys.modules[__name__]
+        # Dynamically create list of subclass names for Dimension
+        dimension_classes = ['Real', 'Integer', 'Categorical']
```
betatim: I'd make a dictionary like `{'real': Real, 'categorical': Categorical, ...}` to take care of converting from the yaml strings to classes. Purely because `this_module` and `getattr` looks like there will be some subtle edge cases where it stops working that we will run into at some point in the future (but that is just a feeling).

skopt/space/space.py
```diff
@@ -550,6 +552,49 @@ def is_real(self):
...
+        * `namespace` [str, default=None]:
+           Namespace within configuration file to use
+
+        Returns
+        -------
+        * `space` [Space]:
+           Instantiated Space object
+        """
+        with open(yml_path, 'rb') as f:
+            cfg = yaml.load(f)
```
betatim: where possible i'd spell out variables, so `cfg` -> `config` and so on

skopt/space/space.py
```diff
@@ -550,6 +552,49 @@ def is_real(self):
...
+
+        this_module = sys.modules[__name__]
+        # Dynamically create list of subclass names for Dimension
+        dimension_classes = ['Real', 'Integer', 'Categorical']
+
+        # Extract space options for configuration file
+        options_name = namespace or list(cfg.keys())[0]
```
betatim: `list(cfg.keys())[0]` could be `next(config)`? Not sure which is nicer.

What do you think about making this into an explicit:
```python
if namespace is None:
  namespace = next(config) # or list(config.keys())[0]
```
explicitly comparing to `None` and overwriting `namespace`. Feels like a "standard pattern" for dealing with default arguments

skopt/space/space.py
```diff
@@ -550,6 +552,49 @@ def is_real(self):
...
+        options = cfg if isinstance(cfg, list) else cfg[options_name]
+
+        # Populate list with Dimension objects
+        dimensions = []
+        for option in options:
+            key = list(option.keys())[0]
```
betatim: this having to convert iterators into lists this feels/reads awkward but I don't have a better idea how to make it look "nicer" :-(

skopt/tests/test_space.py
```diff
@@ -479,3 +482,41 @@ def test_dimension_name():
...
+    tmp_file.close()
+
+    space = Space([(0.0, 1.0), (-5, 5),
+                   ("a", "b", "c"), (1.0, 5.0, "log-uniform"), ("e", "f")])
+
+    space2 = Space.from_yaml(tmp_file_path)
+    assert_equal(space, space2)
+
+    os.unlink(tmp_file_path)
+    assert not os.path.exists(tmp_file_path)
```
betatim: context managers are great, sad we can't use it here :( But nice to see you clean up and check for cleanup!

betatim:
Thanks for this. I left a few comments, mostly nitpicks.

Should we add a short description/example of the yaml format expected to the doc string? Right now you'd have to dive into the tests to confirm/deny if it looks how you'd expect it to look.

skopt/space/space.py
gramhagen: sure no problem

skopt/space/space.py
gramhagen: i believe list(cfg.keys())[0] is equivalent to next(iter(cfg.keys())), hard to say if there's an improvement, maybe it's slightly nicer not to deal with indexing /shrug

agreed on the namespace, i will clean this section up a bit

skopt/tests/test_space.py
gramhagen: agreed, alternatively i could mock the open command, but i was hesitant to introduce another package to support testing on python 2

skopt/space/space.py
gramhagen: haha, i was hoping i just didn't know the clean way =)

skopt/tests/test_space.py
gramhagen: i think you can actually use 'with open', i'm just note sure it will work on windows, however given this is just to test functionality maybe that is ok? if not I can switch it back

skopt/tests/test_space.py
```diff
@@ -479,3 +481,40 @@ def test_dimension_name():
...
+            - Real:
+                low: 1.0
+                high: 5.0
+                prior: log-uniform
+            - Categorical:
+                categories:
+                - e
+                - f
+        """)
+        tmp.flush()
```
betatim: Ah! I hadn't thought of this. 👍

betatim:
LGTM. Will leave it to sit a bit to see if anyone else wants to comment and then merge it.

betatim:
Thanks for your patience and congratulations on landing your first PR in scikit-optimize!

PR #665, opened 2018-04-25
[MRG] add new callback `CheckpointSaver`

cschell:
Adds a new callback that can be used to save a checkpoint after each iteration with `skopt.dump`.


cschell:
I'm not sure about how to test the file-stuff, since I have no experience with pytest. For now the test is pretty hacky, maybe someone with more know-how can help me out here?

PR #668, opened 2018-04-29
[MRG] Don't lose dimension name when using `normalize_dimensions`

cschell:
I noticed that the result object of `gp_minimize` contained dimension objects that didn't have their assigned names anymore ([example](https://gist.github.com/cschell/e70bb403d79e4e73250d90952e33d130)). This seemed to be caused by `utils.normalize_dimensions()`, which did not pass on the `name`.


betatim:
Thanks a lot for the contribution and apologies for the astronomically slow response time.

Also 🎉  for your first contribution to scikit-optimize!

gwerbin:
Is this PR included in the `0.6` tag in this repo?

Issue #666, opened 2018-04-27
refit in BayesianSearchCV does not respect fit_params


lunik1:
I am currently running an optimisation for a classifier which uses `sample_weights` when fitting. I pass these to `BayesianSearchCV `via the `fit_params` argument. These then seem to be accounted for during the optimisation (as I get different results when including/excluding the weights), but when the best classifier is retrained on all the folds the weights do not seem to be included.

I have checked this by taking the best parameters and using them to train a new instance of the base classifier both with and without the sample weights. Without the weights the performance is identical to the classifier returned given by `best_estimator_`, with the weights performance is on par with classifiers optimised using a GridSearch/manual tuning.

jkleint:
I ran into this as well.  Here is some (Python 2) code to reproduce.  You can see it does 1 iter of 2 CV folds with the `fit_params` `{'fp': 1}`, then does the final refit without any `fit_params`.

```
#!/usr/bin/env python
from __future__ import print_function
import sys

import numpy as np
from sklearn.base import BaseEstimator, RegressorMixin
from skopt import BayesSearchCV


class DebugRegressor(BaseEstimator, RegressorMixin):
    def __init__(self, knob):
        self.knob = knob

    def fit(self, X, y, **kwargs):
        print('fit params', kwargs, file=sys.stderr)
        return self

    def predict(self, X):
        return X

    def set_params(self, **params):
        print('set_params', params, file=sys.stderr)
        return self


def main():
    model = BayesSearchCV(DebugRegressor(1), {'knob': (1, 5)}, cv=2, n_iter=1, verbose=2, fit_params={'fp': 1})
    model.fit(np.arange(5), np.arange(5))


if __name__ == "__main__":
    main()
```

```
Fitting 2 folds for each of 1 candidates, totalling 2 fits
[CV] knob=3 ..........................................................
set_params {'knob': 3}
fit params {'fp': 1}
[CV] ........................................... knob=3, total=   0.0s
[Parallel(n_jobs=1)]: Done   1 out of   1 | elapsed:    0.0s remaining:    0.0s
[CV] knob=3 ..........................................................
set_params {'knob': 3}
fit params {'fp': 1}
[CV] ........................................... knob=3, total=   0.0s
[Parallel(n_jobs=1)]: Done   2 out of   2 | elapsed:    0.0s finished
set_params {'knob': 3}
fit params {}
```

jkleint:
It seems (?) this easy...  I hope?  ;)

[...]

PR #670, opened 2018-04-30
Include fit_params in BayesSearchCV refit

jkleint:
Issue #666


iaroslav-ai:
Hey @jkleint ,
thanks for the fix! Indeed something that was missing.

madmy016:
Hi,
I faced an issue with hyper-parameter tuning of CatBoostClassifier using BayesSearchCV that the categorical features indices passed as fit parameters are not used in refit and hence refit failing.

I am using scikit-optimize version 0.5.2 and in my local copy, modified the file searchcv.py to replace line 510  as shown in the attachment with the line 

`self.best_estimator_.fit(X, y, **(self.fit_params or {}))`

image

PR #671, opened 2018-05-01
[WIP] Provide option to center objective plots around minimum

ShivamSarodia:
Today, the plot_objective function always evaluates the partial dependence across a random sampling of the space. For my project, this is undesired behavior, because the values of the function far from the minimum vary wildly, causing the partial dependence plots to be useless at best and misleading at worst. I only care about the behavior in cross sections that include the minimum of the function.

I suspect this may be the case for other projects too, so I put together this PR. Let me know if this feature is something you'd like to add -- I'd be happy to modify the way this feature is triggered, add a test, and/or make any necessary clean ups.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -9,6 +9,8 @@ from matplotlib.ticker import MaxNLocator
 
 from scipy.optimize import OptimizeResult
 
+from .utils import expected_minimum
+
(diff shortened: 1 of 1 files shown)


holgern:
Is provided in PR #830

Issue #660, opened 2018-04-18
Visualization of results using categorical parameters


christianblust:
Hi, 
I'm trying to optimize a `RandomForestClassifier `including [criterion ](url)as a parameter (`gini `vs. `entropy`). Are the functions `plot_evaluations `and `plot_convergence `intended for optimization including categorical parameters? `plots.plot_objective(res_rf)` results in the following error.

```
Traceback (most recent call last):
  File "C:/Users/chris/PycharmProjects/FormationDetection/TestHyperOpt.py", line 176, in <module>
    plots.plot_objective(res_rf)
  File "C:\Users\chris\PycharmProjects\FormationDetection\venv\lib\site-packages\skopt\plots.py", line 343, in plot_objective
    rvs_transformed, n_points)
  File "C:\Users\chris\PycharmProjects\FormationDetection\venv\lib\site-packages\skopt\plots.py", line 245, in partial_dependence
    yi = np.linspace(bounds[0], bounds[1], n_points)
  File "C:\Users\chris\PycharmProjects\FormationDetection\venv\lib\site-packages\numpy\core\function_base.py", line 108, in linspace
    start = asanyarray(start) * 1.0
TypeError: ufunc 'multiply' did not contain a loop with signature matching types dtype('<U32') dtype('<U32') dtype('<U32')
```
However `plot_evaluations(res_rf )`creates a figure but it doesn't seem to be correct.

My code:
```
space= [Integer(1, 20, name='max_depth'),
        Integer(2, 100, name='min_samples_split'),
        Integer(5, 30, name='min_samples_leaf'),
        Integer(100,1000, name='n_estimators'),
        Integer(1, n_features, name='max_features'),
        Categorical(['gini', 'entropy'], name='criterion')
        ]

def objective(params):
    max_depth = params[0]
    min_samples_split = params[1]
    min_samples_leaf = params[2]
    n_estimators = params[3]
    max_features = params[4]
    criterion = params[5]

    rf = RandomForestClassifier(random_state=42, max_depth=max_depth, min_samples_split=min_samples_split,
                                min_samples_leaf=min_samples_leaf, n_estimators=n_estimators, max_features=max_features, criterion=criterion)
    return -np.mean(cross_val_score(rf, x_train, y_train, cv=skf, scoring="accuracy"))

counter = 0
def onstep(res):
    global counter
    list = res.x
[...]

Issue #396, opened 2017-06-09
Categorical dimensions in plotting functions


betatim:
Categorical dimensions are not very well supported in the plotting functions. See for example #395.

A good idea on how to show these dimensions is needed or providing a better error message that helps the user deduce what went wrong.

marcelo-dalmeida:
Hi everyone

The errors that could be given are too broad

Per example, mine is: 
`ValueError: setting an array element with a sequence.`

It has been 2+ years since the official issue report and 1+ year since the fix first proposal.

If fixing it properly still isn't mature enough, please consider including some error handling, at least saying that is not supported, as it is somewhat difficult to associate that's the actual reason.


Thank you all for the work done and keep with the good work

holgern:
Solved in version 0.7

PR #675, opened 2018-05-10
[MRG] Support for plotting categorical dimensions

jkleint:
This adds support for categorical dimensions in `plot_evaluations()` and
`plot_objective()`.  It maps categorical values to their indices in the
Dimension for making plots, and labels the ticks with categories. For computing
partial dependence, it accounts for categoricals being one-hot encoded when
transformed.

We manually handle mapping and labeling categoricals rather than letting
matplotlib do it because a) matplotlib only supports categorical plots in recent
versions (2.1+), b) it doesn't guarantee consistent ordering of categories
across plots, and c) it's nice to preserve the original order of categories
given by the user.

`plot_evaluations()`:

We make one histogram bin per categorical value.  We just have to be careful to
preserve categoricals when making arrays with `dtype=object`, and let matplotlib
automatically set the histogram range.

I added a `_map_categories()` helper function that takes an array of points and
converts categories to their integer indices, while also converting the minimum
point and returning an array indicating which dimensions are categorical.

`plot_objective()`:

The plotting function itself just needs to use the mapped integer
representations for categoricals; computing the `partial_dependence` needed some
generalization. Conceptually it's straightforward, the implementation just takes
a little care. While in `plot_evaluations()` we plot every category, here we
sample categories, because the function explicitly requests it, there is a
not-unreasonable interpretation, and it could be computationally prohibitive
otherwise.  Basically, we take the order of the categories in the Dimension, and
choose (up to) `n_points` approximately-evenly-spaced integer indices of those
categories.  Since there was some repeated code to choose these points, I
factored it out into a `_evenly_sample()` helper function.  The remaining
wrinkle was that the code assumed a 1:1 mapping from number of raw Space
dimensions to number of transformed dimensions, but Categoricals break that
assumption by being one-to-many (they are one-hot encoded).  So when we fix a
value for Dimension _i_, we need to figure out where in the transformed space
_i_ went.

`_format_scatter_plot_axes()` needed a few changes to handle categorical
dimensions.  Axis limits should not be set with the dimension bounds since
they're not meaningful.  Categorical axes get a custom tick formatter that maps
integer ticks back to category names.  For locating ticks, we still use
`MaxNLocator`, but set it to only locate ticks at integer locations for
categoricals.

I added simple smoke tests for plots, just testing that they don't crash.  To
use matplotlib non-interactively, the backend needs to be set to Agg early,
before anything else using matplotlib is imported.  This is done in plots.py by
detecting if running under pytest.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -1,13 +1,21 @@
 """Plotting functions."""
+import sys
+from itertools import count
+from functools import partial
+
+# For plot tests, matplotlib must be set to headless mode early
+if 'pytest' in sys.modules:
+    import matplotlib
+    matplotlib.use('Agg')
 
 import numpy as np
-
 import matplotlib.pyplot as plt
 from matplotlib.pyplot import cm
 from matplotlib.ticker import LogLocator
-from matplotlib.ticker import MaxNLocator
-
+from matplotlib.ticker import MaxNLocator, FuncFormatter
 from scipy.optimize import OptimizeResult
+
+from .space import Categorical
 
 
 def plot_convergence(*args, **kwargs):
@@ -195,6 +203,9 @@ def _format_scatter_plot_axes(ax, space, ylabel, dim_labels=None):
     if dim_labels is None:
         dim_labels = ["$X_{%i}$" % i if d.name is None else d.name
                       for i, d in enumerate(space.dimensions)]
+    # Axes for categorical dimensions are really integers; we have to
+    # label them with the category names
+    iscat = [isinstance(dim, Categorical) for dim in space.dimensions]
 
     # Deal with formatting of the axes
     for i in range(space.n_dims):  # rows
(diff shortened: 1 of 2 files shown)


jkleint:
This addresses issues #396 and #660.  Here is an example optimizing the hyperparameters of a decision tree; the categoricals are `criterion` and a `dummy` param.

image

image

Here is the example code.

```
"""Test skopt plotting with categorical dimensions."""
import numpy as np
from sklearn.datasets import load_breast_cancer
from sklearn.tree import DecisionTreeClassifier
from sklearn.model_selection import cross_val_score
from skopt.space import Integer, Categorical
from skopt import plots, gp_minimize


SPACE = [
    Integer(1, 20, name='max_depth'),
    Integer(2, 100, name='min_samples_split'),
    Integer(5, 30, name='min_samples_leaf'),
    Integer(1, 30, name='max_features'),
    Categorical(['gini', 'entropy'], name='criterion'),
    Categorical(list('abcdefg'), name='dummy'),
]


def objective(params):
    rf = DecisionTreeClassifier(**{dim.name: val for dim, val in zip(SPACE, params) if dim.name != 'dummy'})
    return -np.mean(cross_val_score(rf, *load_breast_cancer(True)))


res = gp_minimize(objective, SPACE, n_calls=50)
plots.plot_evaluations(res)
plots.plot_objective(res)
```

jkleint:
I realized this depends on some functionality only available in matplotlib 2.1+, and even then may not give consistent results.  I am working on an update, so don't look too closely yet.

jkleint:
OK, should be ready for people to try.  I'd especially appreciate testing on older versions of matplotlib.  Let me know what you think.

DavidS3141:
I am using this with updated matplotlib and it looks fine. Of course for long names of the categories, the names overlap. I've noticed you put them down below in an angle so that they fit well, but at the top they are horizontally still. If you put them in an angle there, it would fit better. You could also do this on the left. As you can see in the following (1. image) these names also wont fit just using this plain plotting command. I am certainly no expert in plotting, so I use `plt.tight_layout()` quite often in such cases, but in this case it seems to break more stuff, than it fixes (see 2. image).

Few additional suggestions:
- the xtick labels are slightly visible inside the plot matrix, not completly overlapped by the other plots
- you actually put the name of the variable on the axis which is very helpful! The Reals don't have their name displayed but are shown as X0, X1, etc.. Is this something you could add to your PR?

Thanks a lot for this PR! It is really necessary for skopt and should hopefully get merged soon.

objective
evaluations

frederikfaye:
Does this fix what https://github.com/scikit-optimize/scikit-optimize/pull/629 fixes? That PR claims it fixes:

> Currently partial_dependence uses linear spacing to approximate the partial_dependence. For spaces that are log-uniform, this is problematic. In addition, partial_dependence does not work with categories.

bionicles:
this issue prevents us from getting plots

bionicles:
@betatim do you think we can please merge this soon? otherwise users must copy paste this code and that's annoying

jkleint:
Thanks for merging this!  Scikit-opt is awesome!

PR #684, opened 2018-05-29
[MRG] Function for plotting the cumulative regret

stefanocereda:
This PR adds a function for plotting the cumulative regret.

The code is totally similar to plot_convergence, maintaining the same signature. The cumulative regret is defined as the sum of the obtained values minus the global optimum: at epoch `i` the Cumulative Regret will be: ``CR(i) = \sum_{n=0}^i (f(x_n) - optimum)`` (assuming we are minimizing).
Intuitively, it measures how much an optimizer is effective at picking good points.

The global optimum can be given via the `true_minimum` argument or, if missing, the minimum value in the results will be used.

As an example, the function can be used in the following way:
```
from skopt import dummy_minimize, gp_minimize
from skopt.benchmarks import branin
from skopt.plots import plot_regret

D=[[-5.0, 10.0], [0.0, 15.0]]
rd = dummy_minimize(branin, D)
rg = gp_minimize(branin, D)

plot_regret(('dummy', rd), ('gp', rg))
```
Producing the following image:
figure_1
From which its clear how random is learning nothing and gp_minimize is able to progressively learn the branin function and then finds the optimum.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -93,6 +93,89 @@ def plot_convergence(*args, **kwargs):
     return ax
 
 
+def plot_regret(*args, **kwargs):
+    """Plot one or several cumulative regret traces.
+
+    Parameters
+    ----------
+    * `args[i]` [`OptimizeResult`, list of `OptimizeResult`, or tuple]:
+        The result(s) for which to plot the cumulative regret trace.
+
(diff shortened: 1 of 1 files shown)


betatim:
Looks good to me and like a useful addition to the plotting tools. Thanks for this!

Should we add this to one of the examples to help people discover it?

PR #687, opened 2018-06-05
Initialization for Optimizer and BayesCV

### LINKED ISSUE #686: [Question] Set starting parameters using BayesSearchCV?
Is there a way to set the starting parameters using `BayesSearchCV` similar to the `x0` argument in `gp_minimize`?  I tried using `optimizer_kws` but it didn't work for this feature.

iaroslav-ai:
This PR adds feature of initial points to the Optimizer class. Fixes #686 . 

The initialization interface is similar to *_minimize functions in x0 argument. The differences are:
* in xy0 argument, which should contain a list of pairs of (point, objective). The reason I make this interface is to enable providing evaluated and not evaluated points at the same time, which afaik is not possible with *_minimize functions.
* the number of initial points for the optimizer is n_initial_points + len(x0) + len(xy0). This is done in order to ensure that the algorithm will work same as if no initialization was done, in order to avoid unexpected behavior, where e.g. some points from x0 are quitely skipped due to insufficient n_initial_points.

Let me know if you have any concerns.

This PR contains update to code + example + tests.

### DIFF
--- a/examples/optimizer-bcv-initial-points.ipynb
+++ b/examples/optimizer-bcv-initial-points.ipynb
@@ -1,0 +1,319 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to provide intial points to the Optimizer and BayesSearchCV class\n",
+    "\n",
+    "Iaroslav Shcherbatyi, June 2018."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "np.random.seed(123)\n",
+    "\n",
+    "%matplotlib inline\n",
+    "import matplotlib.pyplot as plt"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Problem statement\n",
+    "\n",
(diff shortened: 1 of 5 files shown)


betatim:
I am not sure `Optimizer` should have a list of starting points. If you have some (historical) points you can `tell()` them to your instance instead of passing them to the constructor. This is what the `*_minimize` functions do as well. Supporting a mixture of initial points that are evaluated, points that aren't yet, then maybe some random points if the init points are not enough adds a lot of complexity :-/

For `BayesSearchCV` I think we should support `x0` and `y0` to mirror `*_minimize`.

iaroslav-ai:
Yeah true. Plus for `BayesSearchCV` providing stuff via `x0` of Optimizer would set this x0 for all optimzers of all subspaces :/
I will add a bit later initialization points to `BayesSearchCV`.

iaroslav-ai:
Ok, lets add parameters `x0` and `y0` for the `BayesSearchCV`. I would say these need to be lists of lists of values, so that initialization with multiple subspaces can be done. This would look like this:
```python

model = BayesSearchCV(
    estimator = LinearSVC(),
    search_spaces=[ # multiple search spaces
        {'C': [0.1. 1.0]},
        {'C': [1.1. 10.0]},
    ],
    x0=[ # a list per every subspace; if only one subspace - can be a list of dict
        [{'C': 0.1}, {'C': 0.2}],
        [{'C': 5.1}],
    ],
    y0=[ # corresponds to x0
        [0.8, 0.85],
        [0.7]
    ]
)
```
WDYT about such interface?

jolespin:
Is this available in the current development version?

iaroslav-ai:
@jolespin sorry, not yet. Will get back to it soon hopefully

iaroslav-ai:
Not sure how soon I could get back to this. If someone else would like to take over let me know. Closing this for now.

PR #688, opened 2018-06-06
[MRG] Improve error message if input check fails in _gaussian_acquisition

mirca:
This PR adds a slightly more informational error message for the input checking in ``_gaussian_acquisition``.


betatim:
Obrigado (I think)!

mirca:
That's right xD

PR #690, opened 2018-06-06
[MRG] Fix bad indentation in test_gaussian_acquisition_check_inputs

mirca:
The unit test I wrote earlier had an indentation bug. This PR fixes it.
cc @betatim


betatim:
Interesting, I wonder why this didn't lead to the tests failing?

mirca:
The ``__exit__`` method is called right after the line ``vals = _gaussian_acquisition(np.arange(1, 5), model)`` raises the exception. That's why the wrong indented (a with a syntax error) line ``assert("it must be 2-dimensional" in exc.value.args[0])`` was not even executed.

PR #692, opened 2018-06-10
Use native python types inside BayesSearchCV

### LINKED ISSUE #691: Types of parameters set by BayesSearchCV are incompatible with some sklearn models
The following code raises an exception:
```python
from sklearn.datasets import load_iris
from sklearn.svm import SVC

model = BayesSearchCV(
    estimator=SVC(),
    search_spaces={'kernel': ['rbf', 'linear']}
)

X, y = load_iris(True)
model.fit(X, y)
```

The important line in exception is:

```bash
TypeError: Argument 'kernel' has incorrect type (expected str, got numpy.str_)
```

It would be better if internally `BayesSearchCV` would convert arguments that it is passing to the estimators to native python types, such as by using `np.asscalar`.

Related to

iaroslav-ai:
Fixes #691

### DIFF
--- a/skopt/tests/test_searchcv.py
+++ b/skopt/tests/test_searchcv.py
@@ -5,6 +5,8 @@ search with interface similar to those of GridSearchCV
 import pytest
 import time
 
+import numpy as np
+
 from sklearn.utils.testing import assert_greater
 from sklearn.datasets import load_iris, make_classification
 from sklearn.model_selection import train_test_split
@@ -12,6 +14,7 @@ from sklearn.pipeline import Pipeline
 from sklearn.svm import SVC, LinearSVC
 from sklearn.tree import DecisionTreeClassifier
 from sklearn.base import clone
+from sklearn.base import BaseEstimator, ClassifierMixin
 from sklearn.externals.joblib import cpu_count
 
 from skopt.space import Real, Categorical, Integer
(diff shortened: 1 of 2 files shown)


skopt/tests/test_searchcv.py
```diff
@@ -314,3 +317,21 @@ def test_searchcv_total_iterations():
...
+    # Test whether the parameters passed to the
+    # estimator of the BayesSearchCV are of standard python
+    # types - float, int, str
+
+    # Below is example code that used to not work.
+    X = np.random.randn(10, 1)
+    y = X[:, 0] > 0
+
+    model = BayesSearchCV(
+        estimator=SVC(),
```
betatim: Instead of `SVC` can use use our own classifier class that checks the types of the parameters?

skopt/tests/test_searchcv.py
```diff
@@ -5,13 +5,16 @@
...
+import numpy as np
+
 from sklearn.utils.testing import assert_greater
 from sklearn.datasets import load_iris, make_classification
 from sklearn.model_selection import train_test_split
 from sklearn.pipeline import Pipeline
 from sklearn.svm import SVC, LinearSVC
 from sklearn.tree import DecisionTreeClassifier
 from sklearn.base import clone
+from sklearn.base import BaseEstimator, ClassifierMixin
```
betatim: Where these imports missing before or why did they get added?

skopt/tests/test_searchcv.py
iaroslav-ai: :+1:

skopt/tests/test_searchcv.py
iaroslav-ai: That was actually for a custom classifier :) Did not get to it yesterday

betatim:
(aside: what happened to pep8speaks?)

OrkoHunter:
> (aside: what happened to pep8speaks?)

Hi @betatim ! By default, @pep8speaks makes a comment only if any style issue is found in the PR. Trying to be the least noisy bot :) If you wish, you can change the behavior by setting [`no_blank_comment` to `False`](https://github.com/OrkoHunter/pep8speaks#configuration).

I hope I am not intruding by making this comment here. I look around the comments involving pep8speaks to understand more about community to improve the project. :)

Issue #566, opened 2017-11-28
plot_* functions with BayesSearchCV Result?


galtay:
Given that I have the results of a call to the `fit` method of `BayesSearchCV`, is there something from that object I can pass to the plotting functions (plot_convergence, plot_objective, ...)?  

```
bscv = BayesSearchCV(...).fit(X, y)
plot_objective(???)
```

betatim:
Currently no. We don't attach/store the result returned whe ncalling `optimizer.tell(..)` in `BayesSearchCV`. I can't think of a reason why not to do that though, so we should.

Do you want to open a PR? I would create a `skopt_result_` variable in `BayesSearchCV` which gets set to the latest result. One thing to take care of is when there are multiple search spaces and optimizers. And maybe a better name for the variable.

galtay:
Great library!  Don't have time at the moment.  Might pick it up over the holiday break.  Will let you know.

christianblust:
Any news on this feature? What would be the alternative to get the plots for bayesian hyper parameter tuning? Just following the approach described in https://github.com/scikit-optimize/scikit-optimize/blob/master/examples/hyperparameter-optimization.ipynb?

betatim:
No one has had time to work on this. Would be a good first contribution though.

PR #695, opened 2018-06-12
[MRG] Return optimizer results in BayesSearchCV

bernardocaldas:
This partially solves #566 
I've created a dictionary called optimizer_results_ that stores the optimizer results as value and the optimizer as key. This makes it work if the user specifies multiple optimizers.

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -663,7 +663,7 @@ class BayesSearchCV(BaseSearchCV):
 
                 if eval_callbacks(callbacks, optim_result):
                     break
-            
+
             self.optimizer_results_[optimizer] = optim_result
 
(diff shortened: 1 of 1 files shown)


bernardocaldas:
@betatim @iaroslav-ai  any feedback?

iaroslav-ai:
Hi @bernardocaldas , thank you for this PR!
A few things:
* Please create `optimizer_results_` using `@property` decorator, similar to `best_params_` for example, to make sure that if the model is not fitted, a meaningful exception is thrown.
* Please add the description of new parameter into the doc-string of the class.
* It would be great if you could add a test that ensure consistency of values returned by the property, as well as that exception is thrown.

iaroslav-ai:
Ooops wrong button clicked in a rush

PR #696, opened 2018-06-12
[MRG+1] Add versioneer support

tritemio:
Add support for automatic and granular versioning with [versioneer](https://github.com/warner/python-versioneer).

To try it, checkout the branch, build the package (this computes the version) and install it:

```
python setup.py build
pip install -e .
```
After that you get:

```
In [1]: import skopt

In [2]: skopt.__version__
Out[2]: 'v0.5.2+36.g5bba7b9'
```
The version is: _last version tag_ `+` _number of commits after last version_ `.g` _git commit hash_.
If you checkout a tagged version the version string will change automatically.


tritemio:
FYI, I manually edited only `setup.py` and `setup.cfg`, the rest is boilerplate added by versioneer (using `versioneer install`).

betatim:
Looks good to me.

Where do you change the version when making a new release?

betatim:
Nice! I don't think there is a particular reason for us using `vX.Y.Z` as tag vs `X.Y.Z`. With the version not having to be set anymore we remove another annoying step from the release procedure! 🎉 

Have you used release-via-tagging-directly-to-PyPI with versioneer? We tag the release on GitHub and then travis builds it and directly pushes it to PyPI. If not that is fine, just means we need to find out if that all works via some other mechanism.

tritemio:
No, I haven't used the automatic upload. I usually do it by hand.

tritemio:
However, there is no reason why auto-deployment to PyPI should not work.

betatim:
Let's see if any of the other devs have opinions, otherwise I'll merge it. Thanks for making our life easier :)

betatim:
Thank you!

Issue #723, opened 2018-10-02
BayesSearchCV does not work with Scikit-learn v0.20.0


GuillemGSubies:
I did not find any similar issue. The only i found is [this](https://www.kaggle.com/c/home-credit-default-risk/discussion/64004). I use Scikit-learn 0.20.0 and the latest version of scikit-optimize. 
This is releated to #718 

Here there is the code and the error i get (it works perfectly with sklearn v0.19):

```python
from sklearn.ensemble import RandomForestClassifier
from sklearn.feature_selection import VarianceThreshold, SelectKBest
from sklearn.model_selection import train_test_split
from skopt import BayesSearchCV

selector = VarianceThreshold()
XX = selector.fit_transform(X)
selector = SelectKBest(k=425)
XX = selector.fit_transform(XX, y)
X_train, X_test, y_train, y_test = train_test_split(XX, y, test_size=0.4, random_state=0)
rf = BayesSearchCV(
    RandomForestClassifier(n_jobs=-1),
    {
        'n_estimators': (100, 10000),  
        'criterion': ["gini", "entropy"],
        'max_features': (0.1, 1, "uniform"),
        'min_samples_split': (2, 16),
        'min_samples_leaf': (1, 10)
    },
    n_iter=32
)
rf.fit(X_train, y_train)
```

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-1734270d0466> in <module>()
      8         'min_samples_leaf': (1, 10)
      9     },
---> 10     n_iter=32
     11 )
     12 rf.fit(X_train, y_train)

TypeError: Can't instantiate abstract class BayesSearchCV with abstract methods _run_search
```

iaroslav-ai:
HI @GuillemGSubies , yes, I can confirm that there is an issue with most recent sklearn. A workaround for now would be to 
```python
from skopt import BayesSearchCV

# include below until https://github.com/scikit-optimize/scikit-optimize/issues/718 is resolved
class BayesSearchCV(BayesSearchCV):
    def _run_search(self, x): raise BaseException('Use newer skopt')
```
This implements a new `_run_search` method in abstract sklearn class from which BayesSearchCV implements. This should raise an exception if in future BayesSearchCV actually uses _run_search.

iaroslav-ai:
[...]

PR #724, opened 2018-10-03
Fix build with newest scikit-learn

iaroslav-ai:
A new abstract method `_run_search` has been introduced in superclass of `BayesSearchCV`, which needs to be implemented. If it is not, then an exception is thrown when `BayesSearchCV` is instantiated. 

This should fix things so that tests do not fail.


iaroslav-ai:
#723

betatim:
Seems like a good temporary fix.

Issue #726, opened 2018-10-28
Numpy broadcasting errors in gaussian_ei, gaussian_pi


hinnefe2:
When I run `skopt.acquisition.gaussian_ei` I get a `ValueError` related to a shape mismatch
```
~/miniconda3/envs/brewlog/lib/python3.7/site-packages/skopt/acquisition.py in gaussian_ei(X, model, y_opt, xi, return_grad)
    284     exploit = improve * cdf
    285     explore = std[mask] * pdf
--> 286     values[mask] = exploit + explore
    287 
    288     if return_grad:

ValueError: shape mismatch: value array of shape (47,47) could not be broadcast to indexing result of shape (47,1)
```

This seems to be the same issue as #632. I think this is due to numpy broadcasting in the division [here](https://github.com/scikit-optimize/scikit-optimize/blob/80d8a7b08c870bb032f06d5e4c34794c10376fd7/skopt/acquisition.py#L282):
```
mu, std = model.predict(X, return_std=True)
...
mask = std > 0
improve = y_opt - xi - mu[mask]
scaled = improve / std[mask]
```
The issue is that calling `predict` with `return_std=True` returns `mu` as a 2D vector but `std` as a 1D vector according to the [docs](http://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor.predict).

Dividing an `(N, 1)` vector by an `(N,)` vector gives an `(N, N)` vector, hence the shape mismatch.

MWE:
```
import numpy as np

from sklearn.gaussian_process import GaussianProcessRegressor
from sklearn.gaussian_process.kernels import RBF
from skopt.acquisition import gaussian_ei

X = np.linspace(0, 10, 100).reshape(-1, 1)
y = 3 * X

gpr = GaussianProcessRegressor(kernel=RBF()).fit(X, y)

mu, std = gpr.predict(X, return_std=True)

print(mu.shape)
# (100, 1)

print(std.shape)
# (100,)

print((mu / std).shape)
# (100, 100)

gaussian_ei(X, gpr)
# ValueError: shape mismatch: value array of shape (47,47) could not be broadcast to indexing result of shape (47,1)
```

iaroslav-ai:
Fixed by #727 
I remember having same problems as you do at some point, thank you for fixing this.

PR #727, opened 2018-10-28
Fix numpy broadcasting issues in gaussian_ei, gaussian_pi

hinnefe2:
This PR fixes the numpy broadcasting issues in `acquisition.gaussian_ei` and `acquisition.gaussian_pi` described in #726.

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -201,6 +201,9 @@ def gaussian_pi(X, model, y_opt=0.0, xi=0.01, return_grad=False):
         else:
             mu, std = model.predict(X, return_std=True)
 
+    # make sure mu and std have the same shape so we can divide them below
+    std = std.reshape(mu.shape)
+
     values = np.zeros_like(mu)
     mask = std > 0
     improve = y_opt - xi - mu[mask]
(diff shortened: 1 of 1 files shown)


iaroslav-ai:
Thank you @hinnefe2 for this PR!

Just a context if we look back at this PR at some point: GaussianProcessRegressor accepts as input argument y, which can be a matrix, in which case it is assumed that the model should have multiple outputs (1 or more), determined by the number of columns. At the moment it seems however that we assumed that the model is fit with y which is a vector, but not a matrix; It seems so from the shape of the output of gaussian_ei for example given in docstrings.

@hinnefe2 your example in #726 is similar to
```python3
import numpy as np
from sklearn.gaussian_process import GaussianProcessRegressor
from skopt.acquisition import gaussian_ei

X = np.random.randn(100, 3)
y = np.random.randn(100, 1)

model = GaussianProcessRegressor()
model = model.fit(X, y)

gaussian_ei(X, model)
```
As you mention, this breaks gaussian_ei. If y = np.random.randn(100, 1), then everything is fine. I wonder, do you have any particular use case in which to use multiple outputs, that is, with y.shape = (N, 1) or with more than one output? My suspicion would be that when there are multiple outputs, a dedicated acquisition function might be required.

hinnefe2:
Thanks for the clarification. I wasn't actually trying to fit a multi-output model, I had just trained my model with a (N, 1) vector instead of an (N,) vector (like you mentioned).

I replaced the reshaping with a check on the dimensionality of `mu` and `std` along with an informative error message. Let me know if you think this is worth merging, if not I can close the PR now that I realize what my issue was.

Thanks!

iaroslav-ai:
Great, thank you @hinnefe2! LGTM!

Issue #735, opened 2018-12-04
Giving prior='log-uniform' to Integer dimension


frederikfaye:
**TL;DR:** I would like to sample logarithmically over an Integer Dimension - is this possible? If not, is there an easy hack?

I would like to logarithmically uniformly sample my variable `batch_size`, which is an integer,  from a distribution over, say, (128,1024). A hack would be to sample another integer 'log2_batch_size' over (7,10), but what if I wanted sampling from in between the values of [128,512,256,1024] as well?

homarb:
Would this work?
```python
Categorical(np.logspace(1,10, num=10, base=2, dtype=np.int)
```

mfinno:
I think this (hack) is a bit better:

`Real(128, 1024, name='batch_size', prior='log-uniform')`

You would need to cast the output as int, but this method will cover the entire range in a log-uniform way. Also `Categorical` does not work with sckit-opt plots.

frederikfaye:
@homarb I don't think that is desirable; beside the reasons @mfinno gives, I don't think gp_minimize works as well with categoricals, as the distance function is far less informative (if someone could confirm this, I would appreciate it). 

@mfinno, I like your idea, but I'm leaving the issue open to see if anyone want to implement this functionality. Btw, https://github.com/scikit-optimize/scikit-optimize/pull/675#issue-187282057 should solve the plotting of categorical dimensions.

liebscher:
I just ran into this issue as well. For example, should we want to search over n_estimators for a random forest, the parameter only takes integer values (thus excluding a Real space). However, it seems overly hack-y to code this as Categoricals. Additionally, using `np.logspace` would violate the notion that this function is continuous. 

It might seem more roundabout and perhaps pedantic, but I suggest that a decent solution to this would be: given base 2 high and low values, log2 transform them, then choose random integers in [log2(min), log2(max)], then transform them back to linear space. I believe this would keep everything as integers and eliminate the need to do any casting.

frederikfaye:
@liebscher Yep, that is also what I ended up doing, but that can only yield sampled values that are powers of 2, even though there are integers in between.

liebscher:
[...]

Issue #712, opened 2018-08-15
Integer dimension with log-uniform as prior


DavidS3141:
I could not find a issue addressing this, so here it goes.

I use this wonderful tool for hyper parameter tuning of neural networks. One common parameter is for e.g. the batch size, good values go usually from 32 to the order of 1000 depending on the project and the size of the dataset, as well as some other minor things. This is a parameter which should be probed logarithmically, so a prior log-uniform key like the one for Reals would make a lot of sense.

Right now I am emulating this by using actually a Real batch_size and inside my neural network casting it to an integer, as well as manually caching results for the same batch size (and the other parameters, which happens rarely I admit).

Is there a design reason I am not seeing, why this functionality was left out for the Integer class?

holgern:
Solved in 0.7

PR #739, opened 2018-12-17
[MRG] Add log-uniform prior to Integer space

liebscher:
After a few requests and confusions about this, there didn't seem to be any reason to not implement it. Referring to issues #735 and #712 . This is my first PR to this library, so while I'm happy to contribute I also wish to respect the current code and will make changes/take suggestions happily!

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -297,7 +297,7 @@ class Real(Dimension):
 
 
 class Integer(Dimension):
-    def __init__(self, low, high, transform=None, name=None):
+    def __init__(self, low, high, prior="uniform", transform=None, name=None):
         """Search space dimension that can take on integer values.
 
         Parameters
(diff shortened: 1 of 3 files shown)


frederikfaye:
I think this should include an option to transform using `np.log2()` as well as `np.log10()` (which would entail creating another class `Log2()`). 

It is just a convenience to the user not to need to calculate the equivalent `low` and `high` values if they actually want to sample something that usually makes more sense in `log2` space (with batch size).
 
Just to demonstrate that it makes a difference:

```python
import numpy as np

min_value = 13
max_value = 1042

r_10 = np.random.uniform(np.log10(min_value), np.log10(max_value), 10**4)
r_2 = np.random.uniform(np.log2(min_value), np.log2(max_value), 10**4)

fig,ax = plt.subplots()
ax.hist(r_10, label='log10 transformed')
ax.hist(r_2, label='log2 transformed')
plt.legend()
plt.show()
```

skopt/space/space.py
```diff
@@ -188,6 +200,10 @@ def __init__(self, low, high, prior="uniform", transform=None, name=None):
               and upper bounds.
             - If `"log-uniform"`, points are sampled uniformly between
               `log10(lower)` and `log10(upper)`.`
```
frederikfaye: Should probably be changed to something akin `log(lower)` and `log(upper)`, where `log` has base `base`.' 

That detail aside, it may be overkill to introduce anything other than `Log10` and `Log2` (and *maybe* the natural log). On the other hand, with `LogN` one only needs one class, and can get rid of the `Log10` class. ... But I'll leave it to the regular contributors to discuss that.

liebscher:
@betatim @iaroslav-ai any thoughts? 🙂

Issue #349, opened 2017-04-10
Runaway memory use?


dhirschfeld:
I'm using `gm_minimize` for feature selection in a GBRT model. There are ~150 possible features and I'm looking for the best 20 which will give the most predictive model. My objective function thus trains a GBRT model and returns the cross-val score.

What I'm observing is that memory use goes up and down periodically , but the amplitude of the "waves" gradually increases. At the start of the optimization the memory peaked at ~35GB. After many hours and 100 iterations (of each of the 3 optimizations I'm running in parallel) the memory usage now peaks at >112GB and unsurprisingly the time for each iteration has more than doubled.

Is this expected behaviour? i.e. should the memory requirements for each iteration step continuously increase as the optimization progresses?

dhirschfeld:
![image](https://cloud.githubusercontent.com/assets/881019/24883340/ce1915b0-1e87-11e7-9745-1fb6152461ee.png)

betatim:
The quick answer is: yes. We store each fitted surrogate model created during each iteration, so memory use will keep going up.

To find out if this is indeed the cause (or if there is a leak) you could make a quick hack that only stores the last model and see what happens if you run it with an objective function that is fast to evaluate.

We should consider offering an option to store only the last N models/only ever Nth model for people who do not need to/want to inspect each surrogate in detail.

dhirschfeld:
...maybe also a spill-to-disk cache for the model history.

dhirschfeld:
Just for reference, I'm assuming you're talking about [this](https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/optimizer/optimizer.py#L272) line.

Bacoknight:
Bit of a necropost, but I've noticed a similar issue for my optimisations too, which are run for many iterations. Has anyone managed to find a proper way around this? I don't need to keep previous models as I only care for the result (for now). I can't seem to find how to discard older models.

glouppe:
Fixed by #746

PR #746, opened 2019-02-11
[MRG] Models list is now a customizable size.

Bacoknight:
I've modified the models list of the Optimizer class to be a deque (from collections). This allows a maximum list size to be set so that scripts don't get bloated as the number of iterations increases. This is meant to be a fix for issue #349.

The default value keeps the list as arbitrary in size, so as to keep the same functionality as before.

I can add these to the high-level cases (gbrt_minimize etc) if requested, and will update the Optimizer docstring if this is deemed a useful addition.

### DIFF
--- a/skopt/optimizer/base.py
+++ b/skopt/optimizer/base.py
@@ -155,6 +155,10 @@ def base_minimize(func, dimensions, base_estimator,
         `acq_optimizer` is set to "lbfgs."
         Defaults to 1 core. If `n_jobs=-1`, then number of jobs is set
         to number of cores.
+    
+    * `model_history` [int or None, default=None]
+        Keeps list of models only as long as the argument given. In the 
+        case of None, the list has no capped length.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


Bacoknight:
Instead of deque which seems to cause issues, I'll stick to a list with an if statement. Requires no importing too, and there shouldn't be too much cost in terms of time (and certainly insignificant with respect to other processes).

kitchoi:
I ran into the same issue in #349.  Changes here look promising!
You mentioned using `collections.deque` causes some issue, what were the issues you ran into?

Bacoknight:
> 
> 
> I ran into the same issue in #349. Changes here look promising!
> You mentioned using `collections.deque` causes some issue, what were the issues you ran into?

The issues were mainly with testing, whereby they were expecting empty lists rather than empty deque lists (see for example: https://travis-ci.org/scikit-optimize/scikit-optimize/jobs/491534664). I can't imagine it would be difficult to add if you'd like a deque instead.

glouppe:
Closed by #746

PR #752, opened 2019-04-03
[MRG] Fix bug with categorical spaces with different types of ca…

### LINKED ISSUE #704: Unexpected type conversion in categorical space
I am unsure if this behavior is intended or not. The just of the bug/feature is that when I create a `Space` that has only categorical dimensions, they types of the dimension's categories may not differ because some unexpected type casting occurs.

This Works:
```
from skopt import gp_minimize
from skopt.space.space import Real, Integer, Categorical, Space

space = Space([
    Categorical(name="x", categories=["1","2","3"]),
    Categorical(name="y", categories=[4,5,6]),
    Real(name="z", low=1.0, high=5.0)
])

def objective(param_list):
    x = param_list[0]
    y = param_list[1]
    z = param_list[2]

    loss = int(x) + y * z

    return loss

res = gp_minimize(objective, space, n_calls=50)

print(res["x"])
```

This does not:
```
from skopt import gp_minimize
from skopt.space.space import Real, Integer, Categorical, Space

space = Space([
    Categorical(name="x", categories=["1","2","3"]),
    Categorical(name="y", categories=[4,5,6])
])

def objective(param_list):
    x = param_list[0]
    y = param_list[1]

    loss = int(x) + y

    return loss

res = gp_minimize(objective, space, n_calls=50)

print(res["x"])
```
Stack trace from running the second snippet of code:
```
a=5
a_type=<type 'numpy.string_'>
b=4
b_type=<type 'int'>
  File "test.skopt.py", line 19, in <module>
    res = gp_minimize(objective, space, n_calls=50)
  File "<...>/scikit-optimize/skopt/optimizer/gp.py", line 228, in gp_minimize
    callback=callback, n_jobs=n_jobs)
  File "<...>/scikit-optimize/skopt/optimizer/base.py", line 247, in base_minimize
    next_x = optimizer.ask()
  File "<...>/scikit-optimize/skopt/optimizer/optimizer.py", line 322, in ask
    return self._ask()
  File "<...>/scikit-optimize/skopt/optimizer/optimizer.py", line 397, in _ask
    for xi in self.Xi])
  File "<...>/scikit-optimize/skopt/space/space.py", line 799, in distance
    distance += dim.distance(a, b)
  File "<...>/scikit-optimize/skopt/space/space.py", line 507, in distance
    " the space, not {} and {}.".format(a, b))
RuntimeError: Can only compute distance for values within the space, not 5 and 4.
```
Some thoughts:
I believe the problem stems from a type conversion in the `_tell` method in `../scikit-optimize/skopt/optimizer/optimizer.py`.
First the object `X` is created. Where the samples created in `X` have correct types.
In `_tell` from `../scikit-optimize/skopt/optimizer/optimizer.py lines 494-495`:
```
            X = self.space.transform(self.space.rvs(
                n_samples=self.n_points, random_state=self.rng))
```

Then in the same file lines `538-546` the local variable named `next_x` is defined based on some logic that manipulates the object `X` (where I suspect casting occurs).

`next_x` isthen referenced in the assignment of `self._next_x`
In `_tell` from `../scikit-optimize/skopt/optimizer/optimizer.py lines 549 - 550`:
```
    self._next_x = self.space.inverse_transform(
                    next_x.reshape((1, -1)))[0]
```

Then in the function `_ask` from `../scikit-optimize/skopt/optimizer/optimizer.py, lines 396-297`:
```
            next_x = self._next_x
            min_delta_x = min([self.space.distance(next_x, xi)
                               for xi in self.Xi])
```

Which calls `Space.distance` from `../scikit-optimize/skopt/space/space.py, lines 795-799`:
```
distance = 0.
        for a, b, dim in zip(point_a, point_b, self.dimensions):
            distance += dim.distance(a, b)

        return distance
```

Which calls `Categorical.distance` from `../scikit-optimize/skopt/space/space.py lines 503-506`:
```
        if not (a in self and b in self):
            # BEGIN CODE I INSERTED
            print("a=" + str(a) + "\n" + "a_type=" + str(type(a)))
            print("b=" + str(b) + "\n" + "b_type=" + str(type(b)))
            # END CODE I INSERTED
            raise RuntimeError("Can only compute distance for values within"
                               " the space, not {} and {}.".format(a, b))
        return 1 if a != b else 0
`` `

stefanocereda:
Closes #704 
The problem happened with categorical domains with different data types, like [['a', 'b'], [True, False]].
When the space is categorical the transformer is an identity, so the data keep their type. When stacking different types into a numpy array, they all become strings, losing their types.
The identity transformer is now able to translate everything into a string and then transform back into the original data type.
However, mixed categories like [['a', True]] are (still) not supported.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -410,7 +410,11 @@ class Categorical(Dimension):
         * `name` [str or None]:
             Name associated with dimension, e.g., "colors".
         """
-        self.categories = tuple(categories)
+        if transform == 'identity':
+            self.categories = tuple([str(c) for c in categories])
+        else:
+            self.categories = tuple(categories)
+
         self.name = name
 
         if transform is None:
@@ -423,7 +427,8 @@ class Categorical(Dimension):
             self.transformer = CategoricalEncoder()
             self.transformer.fit(self.categories)
         else:
-            self.transformer = Identity()
+            self.transformer = Identity(dtype=type(categories[0]))
+
         self.prior = prior
 
         if prior is None:
--- a/skopt/space/transformers.py
+++ b/skopt/space/transformers.py
@@ -21,7 +21,7 @@ class Identity(Transformer):
     string and the inverse transform will cast to the type defined in dtype."""
 
     def __init__(self, dtype=None):
-        super().__init__()
+        super(Identity, self).__init__()
         self.dtype = dtype
 
     def transform(self, X):
--- a/skopt/space/transformers.py
+++ b/skopt/space/transformers.py
@@ -16,13 +16,24 @@ class Transformer(object):
 
 
 class Identity(Transformer):
-    """Identity transform."""
+    """Identity transform.
+    If dtype is different from None the transform will cast everything to a
+    string and the inverse transform will cast to the type defined in dtype."""
+
+    def __init__(self, dtype=None):
+        super().__init__()
+        self.dtype = dtype
 
     def transform(self, X):
-        return X
+        if self.dtype is None:
+            return X
+        return [str(x) for x in X]
+
 
     def inverse_transform(self, Xt):
-        return Xt
+        if self.dtype is None:
+            return Xt
+        return [self.dtype(x) for x in Xt]
 
 
 class Log10(Transformer):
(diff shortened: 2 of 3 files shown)


betatim:
Merged on advice of @holgern

PR #760, opened 2019-04-29
Minor changes to parallel optimization example

JohanRensink:
Possibly due to changing functionality of Parallel and/or Optimizer


examples/parallel-optimization.ipynb
```diff
@@ -45,20 +45,20 @@
...
         "from skopt.space import Real\n",
         "from sklearn.externals.joblib import Parallel, delayed\n",
         "# example objective taken from skopt\n",
         "from skopt.benchmarks import branin\n",
         "\n",
         "optimizer = Optimizer(\n",
         "    dimensions=[Real(-5.0, 10.0), Real(0.0, 15.0)],\n",
-        "    random_state=1\n",
+        "    random_state=1,\n",
+        "    base_estimator='gp'\n",
```
betatim: (Why) do we need to explicitly specify this?

JohanRensink:
I only added this because the original import wasn’t used, and since it’s
an example, it shows where the base_estimator choice is made (so not adding
it loses information vs the original (intent), but of course it’s a style
choice, and not needed at all).

I’m totally fine wither way, the main reason for the PR is obviously the
argument in Parallel.

JR

betatim:
Fine with me. I think in retrospect the import was unused and we should have just removed it long ago :)

PR #767, opened 2019-05-24
refactor: RandomForestRegressor.__init__()

janosh:
- remove hard-coded keyword arguments, some of which are now deprecated by sklearn's underlying RandomForestRegressor and allow passing of newer ones such as `min_impurity_decrease` that have been introduced between this file's last commit (3 Nov 2017) and now
- this is particularly important for people who want to hyperoptimize these new parameters


holgern:
Will be fixed in #829

PR #772, opened 2019-06-07
[WIP] FIX deprecate fit_params at instantiation

glemaitre:
Deprecate the `fit_params` when creating the instance to be compatible with new version of scikit-learn.


betatim:
Still WIP or ready for merge?

PR #799, opened 2020-01-05
Match gridsearch masked array approach

austinwiltshire:
Passing in a list of dicts for the parameter space resulted in cv_results_ not being able to be converted to a dataframe. This was due to some parameter fit runs not always using all parameters, such as the case with sklearn's SVM's different kernels.

Using list.extend only actually added things to a list in the cv_results dictionary when the field was there, and the lists often got out of sync with each other, making the resulting dictionary of lists unable to be consumed by pandas.DataFrame

This uses the approach found in sklearn.gridsearch of masked arrays, and keeping track of the index we're actively filling in the resulting dictionary of masked arrays to keep everything aligned.

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -1,5 +1,6 @@
 from collections import defaultdict, Sized
 from functools import partial
+import six
 
 import numpy as np
 from scipy.stats import rankdata
@@ -543,6 +544,9 @@ class BayesSearchCV(BaseSearchCV):
         # convert parameters to python native types
         params = [[np.asscalar(np.array(v)) for v in p] for p in params]
 
+        assert len(params) == n_points, \
+            "The number of params we tried should equal the number of points we were asked to try"
+
         # make lists into dictionaries
         params_dict = [point_asdict(search_space, p) for p in params]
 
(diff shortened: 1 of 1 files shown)


holgern:
Hi @austinwiltshire,
I modified quite a lot and fixed the CI. Could you merge the upstream changes and try to pass the checks? That would be great.

austinwiltshire:
So I know it's been a while but I finally had some time to come back to this @holgern but I can't get master to build.

Running python 3.7.3 and getting many test failures with pytest using the instructions from the readme.md

austinwiltshire:
It was a version issue, a few things don't work with the lastest sklearn or scipy it looks like. I am now on versions closer to the min supported and tests pass.

I'm trying to figure out why this didn't work in the first place - it looks like you have a test for this very situation, what issue did I find?

austinwiltshire:
@holgern Got my sea legs again and figured things out. Let me know what you think.

austinwiltshire:
I think the same things are failing on this branch as on master. At least the masked array bit I had a fix for elsewhere - remove reference to sklearn's version and use numpy.ma instead (which is all the latest versions of sklearn pointed to anyway).

PR #806, opened 2020-01-15
[MRG] PR #675

holgern:
Evaluate and merge PR #675 with upstream

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,5 @@
+**Disclaimer**: This project is no longer actively maintained! *Help is welcome to take over*.
+
 |Logo|
 
 |Travis Status| |CircleCI Status| |binder| |gitter| |Zenodo DOI|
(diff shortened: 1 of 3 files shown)


holgern:
This PR is ready to be merged. @betatim or @glouppe can you have a look and merge if everything looks good?
Thanks.

glouppe:
I am wondering if converting a categorical into a real value is really the right thing to do for plots. It is certainly better than nothing, and I wouldn't disagree in merging this, but maybe should make it good from the start?

For plots mixing categorical and continuous axis, there are good suggestions at:
- https://seaborn.pydata.org/tutorial/categorical.html
- http://www.cs.umd.edu/hcil/trs/2016-10/2016-10.pdf

I'd rather have a second opinion before merging this one. @betatim?

holgern:
I think this PR is a first starting point and it will be possible to improve plots with categorical data based on the introduced changes.

glouppe:
@betatim what do you think?

glouppe:
I am merging this for now. Thanks Holger!

PR #819, opened 2020-01-29
[MRG] Improve sphinx doc

holgern:
The released sphinx doc can be found under:
https://scikit-optimize.github.io/beta/html/


glouppe:
Hi Holger! You contributed to produce https://scikit-optimize.readthedocs.io/en/latest/index.html a few days ago, but now I also see your activity towards having Sphinx documentation. Is your plan to drop read-the-docs? I am fine with either, but we should also consider the cost of having to maintain the various scripts for generating and deploying the documentation. Although less customisable, my feeling is that read-the-docs is maybe a bit more common and standard, hence possibly easier to use and maintain in the long term.

holgern:
Hi glouppe, the build limits are far too restricted, each project has only 15 minutes with 1Gb Ram. Thus, I must drop read-the-docs again, as I will use sphinx-gallery.
The question is should we use sphinx-gallery or nbsphinx?

Advantages sphinx-gallery:
* all examples are available as py-script and notebook
* usage of all generated plots is possible in documentation
* Links to other examples and docs is easily possible
* Images and outputs are regenerated (it is guaranteed that the current version works)
* Examples can stay in the current location
Disadvantages
* read-the-docs not possible
* Notebooks are not available in github (will be generated during the build process)

Advantages nbsphinx
* read-the-docs possible
Disadvantages
* Examples need to be moved to doc
* plots cannot be used in the documentation
* Notebooks can be viewed in github.com

I want to go withsphinx-gallery and thus drop read-the-docs.

Issue #661, opened 2018-04-20
Dask


mrocklin:
Hi all, scikit-optimize came up in a conversation between scikit-learn and dask devs last night.  I thought I'd open an issue to start conversation directly with this group to see if there are ways in which the two projects might collaborate in the future.

Here are a few links for background:

1.  https://github.com/scikit-optimize/scikit-optimize/issues/584 for a question on general parallelization
2.  http://dask-ml.readthedocs.io/en/latest/joblib.html for documentation on scaling Joblib with Dask (it seems that scikit-optimize uses joblib?
3.  https://dask.pydata.org/en/latest/futures.html for Dask's implementation of concurrent futures, which is probably the right API connection point from a dask perspective
4.  https://github.com/dask/dask-ml/issues/161 for an ongoing conversation about scaling hyper-parameter search, for which the expertise within scikit-optimize might be valuable

But more directly, is there interest in investing time to scale scikit-optimize?  If so then what are the right connection points to do this from the scikit-optimize side?  What are the constraints?  What are the problems where this would actually provide value to users, if any?  Thoughts on the joblib API vs the concurrent.futures API?

Also, please feel free to close if this is sufficiently off-topic or broad.

iaroslav-ai:
Hi @mrocklin ,
First of all I would like to say that Dask is awesome, thanks for your work! I have been using it for a while to scale my own computational experiments onto multiple machines, such as [this code for example](https://github.com/iaroslav-ai/scikit-optimize-benchmarks), where I benchmark other software against skopt. Running it on single machine would simply be too slow.

Right now the easiest way to use Dask would be by using `BayesSearchCV` class, as it uses joblib. `Optimizer` class used inside `BayesSearchCV` supports simple paralellization strategy of Bayesian Optimization called "constant liar", which works across multiple surrogates. 

One issue that we have with our paralellization approach is that it is batch based; That is, n processes / workers are started in parallel, and we wait for all of them to finish. Afaik, it is due to interfaces that joblib offers. And as @mrocklin mentioned, something like implementation of concurrent futures would be great.

In fact, @fcharras has been doing some work in this respect in joblib: https://github.com/joblib/joblib/pull/587 . If this gets through, then we can run new tasks / fittings as soon as new worker is freed, for which I believe @fcharras has already some code. I think with such approach we can still use Dask through joblib.

So to answer the questions:

[...]

PR #828, opened 2020-02-02
Dispatch tasks asynchronously with dask

fcharras:
Fix https://github.com/scikit-optimize/scikit-optimize/issues/661

Unfortunately, at the moment two tests are not passing on my local environment if the error:
```
    def check_x_in_space(x, space):
        if is_2Dlistlike(x):
            if not np.all([p in space for p in x]):
>               raise ValueError("Not all points are within the bounds of"
                                 " the space.")
E               ValueError: Not all points are within the bounds of the space.
```
I suspect that the current code might mistakenly mix points and optimizers during the dispatching ? I'll investigate this next when given more time... (any interested reviewer that get familiar with the code and find the mistake will be welcome too :grin: ).

The main points of care that I can think of are currently listed in the header of `searchcv.py`:

> TODO BEFORE CONSIDERING IF THIS IS WORTH MERGING
> 
> - Decide how to manage retrocompatibility. Retrocompatibility is a huge issue
> on this PR because:
>     1) it adds dask[distributed] as a dependency, which in turns have a lot of
>     dependencies, which could create conflicts, which users want to avoid.
>     2) it removes joblib. For some users this is transparent, but it would
>     break the workflow of people using BayesSearchCV within a joblib context
>     manager to change the backend (for e.g, dask).
>     3) Is is also complicated because the PR changes significantly the logic
>     within BayesSearchCV internals, and (although it makes it simpler and
>     closer to sklearn BaseSearchCV) a specific development is yet to be made
>     to support both backends. For now it only works with dask.
> Because of 1), dask[distributed] should be added as an extra dependency first,
> and only made mandatory if joblib.Parallel is completely dropped from
> BayesSearchCV.
> For 2), an extra argument "backend" should be added to let the user choose
> between joblib and dask (taking two values "joblib" or "dask", first defaulting
> to "joblib" and as long as "dask" is an extra dependency). A planning should be
> prepared to deprecate joblib, add dask as a mandatory dependency, make joblib
> non-default, and finally remove it.
> 3) is yet to be investigated, it is not clear wether what would be better:
> either maintaining side by side the new and the old version of BayesSearchCV or
> trying to make joblib work within the current logic.
> 
> - Compared to the old (synchronous) version, it will be harder to enforce
> determinism with this one whenever `n_jobs` > 1. Indeed, depending on the load
> of the workers, tasks can be sometimes slower, sometimes faster, and the
> optimizer might not be told the same sequences of points. I don't think this
> should be an issue outside of testing (given most estimators rely on random
> initializations anyway).
> 
> - there are a few TODOs to solve in the code. Especially, the output of the
> function _call_and_get_cfg must be structured for a better readability.

I'd add that the case `n_jobs=1` could be worked on (it shouldn't be necessary to instantiate a `dask.distributed.Client` then!).

WDYT ?


fcharras:
Oh and of course it needs performance tests too.

fcharras:
Noting that by design this also fix https://github.com/scikit-optimize/scikit-optimize/pull/799

PR #835, opened 2020-02-06
[WIP] Add initial sampling generation from latin hypercube, sobol, hammersly and halton

holgern:
See also #433 
* Add unittests for IntegerEncoder
* add Sobol, latin hypercube, hammersly and Halton for initial sampling generation
* Example and unit tests were added


betatim:
Are there benchmarks or comparisons somewhere that show the benefits of using these other samplers for the initial points?

holgern:
image
image
Benchmark on hart6 with 25 initial points and 50 call in total. gp_minimize was used.
halton shows the best results, followed by hammersly, lhs with maximin, lhs and sobol.
All  initial point generators are better than random samples.

holgern:
image
branin example with 256 retries

betatim:
Thanks for making the comparisons! Nice work.

One thing I'd consider while adding (or not) things to scikit-optimize is that it is better for users to offer less choice in options that cover (say) 80% of what users want to do.

To cater to experts wanting to do expert things (for which they are by definition a minority) we have tried to make skopt a toolkit that people can use to build their own optimisation machine by plugging the components that make the `Optimizer` class together in a different way.

So I'd always prioritise making things extendable or replaceable and keeping the amount of code and options built-in to scikit-optimize low. This means less maintenance effort, high quality of what is here, and encourages others to make packages with tools that can be used in skopt. Someone else maintaining code is a good thing because it removes burden from the skopt maintainers and keeps us honest in terms of offering a good API for extending things.

---

Another thing I wondered: for MC integration Sobol and friends are great because they cover the space more uniformly than random sampling. However an equally good thing (for MC integration) would be a precomputed grid of points. Except in MC integration you don't know how many points you will sample so you can't make a precomputed/equally spaced grid. However in skopt we decide up front how many initial points to try before switching to 'bayesian optimisation" mode. So maybe another interesting thing to add would be a sampler that doesn't pick points at random but instead uniformly distributes points on a grid across all dimensions.

holgern:
Thanks for your reply, I accidentally merged this PR to early as a mistake. 
I continued my work on this issue on PR #851 
* I removed the init_point_gen_kwargs parameter, so that it will be easier to use.
* I created a cook_initial_point_generator for advance usage
* I added a equally spaced grid layout

PR #885, opened 2020-02-27
Fix docstring: kappa is only used for LCB acquisition function

felker:
Description of `gaussian_lcb()` parameter `kappa` is the opposite of what is true. 

See, e.g., https://scikit-optimize.github.io/stable/auto_examples/exploration-vs-exploitation.html?highlight=kappa

> kappa is only used if acq_func is set to “LCB”. xi is used when acq_func is “EI” or “PI”. By default the acqusition function is set to “gp_hedge” which chooses the best of these three. Therefore I recommend not using gp_hedge when tweaking exploration/exploitation, but instead choosing “LCB”, “EI” or “PI.


holgern:
You are right, thanks!

PR #895, opened 2020-03-25
Update plots.py

AyrtonB:
Added option to specify axis for plot_objective, allowing more complex plots


holgern:
Thanks, can you fix the pep8 issues, so that the CI starts the tests?

PR #903, opened 2020-05-13
Protect against underflow warnings/errors in softmax calculations

hroff-1902:
Set temporary 'ignore' treatment for underflow conditions which can happen in the softmax calculations.
Underflow here does not cause any loss of significance of the result or erroneous results.

Underflow can happen when one of the gains is near zero and other is high, for example 10000.


hroff-1902:
^^ the Runtime underflow warning appears under conda, if you wish to reproduce

Issue #902, opened 2020-05-12
Version incompatibility with latest scikit-learn.


ytsmiling:
```python -c "import skopt"``` fails when `scikit-learn==0.23.0` is installed.

## Error messages
```
ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes' (~/.pyenv/versions/3.8.2/lib/python3.8/site-packages/sklearn/utils/fixes.py)
```

## Environment
I reproduced the problem at least in the following python versions
python: `3.6.6`, `3.8.2`

## Minimum code to reproduce
```bash
pip install scikit-learn==0.23.0 scikit-optimize==0.7.4
python -c "import skopt"
```

Innixma:
Can confirm that this breaks `autogluon<=0.0.8` with `import autogluon`

Innixma:
Issue on scikit-learn: https://github.com/scikit-learn/scikit-learn/issues/17213

rth:
A temporary workaround for people experiencing this would be to monkeypatch the missing import,
```py
from numpy.ma import MaskedArray
import sklearn.utils.fixes

sklearn.utils.fixes.MaskedArray = MaskedArray

import skopt
```

A proper fix will be included in scikit-learn 0.23.1 (or if scikit-optimize makes a release with a fix from https://github.com/scikit-optimize/scikit-optimize/pull/904 before).

magic-lantern:
I am using skopt 0.7.4 and scikit-learn 0.23.1 and looks like 0.23.1 introduced some new errors.

Here's some example code taken straight from the website at https://scikit-optimize.github.io/stable/auto_examples/sklearn-gridsearchcv-replacement.html#minimal-example:

```python
from skopt import BayesSearchCV
from sklearn.datasets import load_digits
from sklearn.svm import SVC
from sklearn.model_selection import train_test_split

X, y = load_digits(10, True)
X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=0.75, test_size=.25, random_state=0)

# log-uniform: understand as search over p = exp(x) by varying x
opt = BayesSearchCV(
    SVC(),
    {
        'C': (1e-6, 1e+6, 'log-uniform'),
        'gamma': (1e-6, 1e+1, 'log-uniform'),
        'degree': (1, 8),  # integer valued parameter
        'kernel': ['linear', 'poly', 'rbf'],  # categorical parameter
    },
    n_iter=32,
    cv=3
)

[...]

PR #904, opened 2020-05-13
[WIP] Fix ImportError for scikit-learn >= 0.23

jwmueller:
Fixes issue:  https://github.com/scikit-optimize/scikit-optimize/issues/902

Without this fix, all packages that depend on scikit-optimize are now broken unless they manually downgrade their sklearn version.
Note that the sklearn MaskedArray was anyway only needed for outdated versions of numpy: 
https://github.com/scikit-learn/scikit-learn/blob/0.22.X/sklearn/utils/fixes.py#L159


jwmueller:
All CI tests pass except for one using scikit-learn 0.21.3, but it's unclear why that one failed.  That version of scikit-learn contains the same MaskedArray code as all the other versions in the passed CI tests:   https://github.com/scikit-learn/scikit-learn/blob/0.21.3/sklearn/utils/fixes.py#L166

betatim:
The tests that fail are related to comparing two arrays that should be the same/very similar but are wildly different now. My guess is that the calculations that lead to them make use of masked arrays somewhere :-/

jwmueller:
@betatim I think there's an issue in the CI tests.  The same test fails when I just committed a "dummy test" version of the code with no change from master... 

This is the failing test:

DISTRIB="conda" PYTHON_VERSION="3.7" NUMPY_VERSION="1.16.0" SCIPY_VERSION="1.0.0" PYAML_VERSION="17.8.0" SCIKIT_LEARN_VERSION="0.21.3" MATPLOTLIB_VERSION="*" COVERAGE="true" JOBLIB_VERSION="0.13"

rth:
So the breakage is the result of https://github.com/scikit-learn/scikit-learn/pull/16725 which removed backports for numpy <1.12.0. The test that failed in the second commit did so for numpy 1.16.0, in which case,
```
from sklearn.utils.fixes import MaskedArray
```
should be exactly equivalent to,
```
from numpy.ma import MaskedArray
```
so the initial fix should indeed work.

betatim:
@jwmueller can you push the commit with your proposed change (and maybe squash the "not used anymore" ones)?

CI with no changes fails because of the problem you are trying to fix here (I think) :)

holgern:
Should we use the same dependecies as scikit-learn:

scikit-learn requires:

Python (>= 3.6)
NumPy (>= 1.13.3)
SciPy (>= 0.19.1)
joblib (>= 0.11)

betatim:
I think increasing the minimum required version of scikit-learn to 0.21 (last two releases before the current one) would be good and reduce the amount of cross dependency compatibility issues. This will also raise the minimum required version of numpy, scipy and joblib. Or put differently: I wouldn't raise those other dependencies to above what scikit-learn requires.

holgern:
@jwmueller Can you push the changes for MaskedArray again?

I will take care of the failing CI.

holgern:
The CI is fixed in pr #906 
image

Waiting for your fix on MaskedArray :)

holgern:
You can merge th upstream changes with:
https://github.com/jwmueller/scikit-optimize/compare/patch-1...scikit-optimize:master?expand=1
This should fix the CI.

jwmueller:
@holgern @betatim I've re-added the changes and now the CI test for sklearn==0.21.3 passes where it failed before.  But your newly added CI test for sklearn==0.23.0 now fails.

Please feel free to take over and modify the code yourself.  It's probably not worth you waiting for my responses just to address these 3-line code changes...

holgern:
The unit tests are broken due to changes in sklearn 0.23 on Gaussian processes regression: https://github.com/scikit-learn/scikit-learn/pull/15782

holgern:
Has been fixed in #939

PR #917, opened 2020-06-18
[MRG]Add hollow iterations early stopper

ludaavics:
Stop the optimization if the improvement over the last n iterations is below a threshold.
Similar to, but not quite the same as DeltaY stopper. For instance, if the the search has already found the true minimum, and the 2nd best is more than `delta` away, DeltaY will never kick in.


holgern:
Thanks a lot for your contribution

PR #921, opened 2020-07-07
Doc line, logspace call, set_transformer default parameter misstype, …

GCidd:
Doc line, logspace call, set_transformer default parameter misstype, use_named_args test output fix and dimension normalization change

Starting from the doc fix: 
Reading the docs for the BayesSearchCV parameters description for the acceptable input types for the search_spaces parameter, I noticed that the last acceptable patameter was on a wrong line, leading to some confusion. So, (dict, int) is returned to the previous line.

Running the pytest on the freshly downloaded dev branch, yielded in the following errors (including the fixes I applied):
In plots.py, in the plot_histogram method, np.logspace is called with the * operator for multiple arguments (np.logspace(*np.log10(dimension.bounds), bins)), which threw an error in the call. The parameter is also positioned before another one (which is wrong and if it worked otherwise, it should have been last).
In utils.py, in the use_named_args method, pytest checks if the printed values are the same with the expected ones. If pytest is run with python 2.7, the "print ("Best fitness", result.fun)" command prints the values differently (in a tuple, with the first item being the string and second the result value). This has a result of a false negative test result, specifically the result.fun value is correct, but the printed value is not exactly the same. For this reason I changed both the print commands to work in python 2.7 too.

In space/space.py, the set_transformer method in Integer and Real classes has a misstype in the default transform parameter (should be "identity" instead of "identitiy").

Lastly, with the new set_transformer method in the Dimension class (and the classes that inherit from it), in the normalize_dimensions method (in utils.py), there is not need to check for the type of dimension inside of the space object, to reinitialize the dimension with a different transform. Now, we can just call the set_transformer method of the Dimension object with the new transform type, without the need of multiple type checking.

### DIFF
--- a/skopt/plots.py
+++ b/skopt/plots.py
@@ -1,3 +1,4 @@
+# -*- encoding: UTF-8 -*-
 """Plotting functions."""
 import sys
 import numpy as np
@@ -1242,7 +1243,7 @@ def plot_histogram(result, dimension_identifier, bins=20, rotate_labels=0,
         # in which case the histogram can be plotted more easily.
         if dimension.prior == 'log-uniform':
             # Map the number of bins to a log-space for the dimension bounds.
-            bins_mapped = np.logspace(*np.log10(dimension.bounds), bins)
+            bins_mapped = np.logspace(np.log10(dimension.bounds), bins)
         else:
             # Use the original number of bins.
             bins_mapped = bins
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -53,8 +53,7 @@ class BayesSearchCV(BaseSearchCV):
         Either estimator needs to provide a ``score`` function,
         or ``scoring`` must be passed.
 
-    search_spaces : dict, list of dict or list of tuple containing
-        (dict, int).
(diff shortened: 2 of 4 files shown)


GCidd:
Please note that the failed tests have nothing to do with the commit. From what I understand they are related to:

- the matplotlib backend used in circleci (`Sphinx-Gallery relies on the matplotlib 'agg' backend to render figures and write them to files. You are currently using the qt5agg backend. Sphinx-Gallery will terminate the build now, because changing backends is not well supported by matplotlib. We advise you to move sphinx_gallery imports before any matplotlib-dependent import. Moving sphinx_gallery imports at the top of your conf.py file should fix this issue Makefile:52: recipe for target 'html-noplot' failed `) and

- the python or sklearn version used in travis-ci (`ImportError: cannot import name 'MaskedArray' from 'sklearn.utils.fixes' (/home/travis/miniconda3/envs/testenv/lib/python3.8/site-packages/sklearn/utils/fixes.py)`)

holgern:
Thanks a lot for your contribution.
The problems with sklearn 0.23 and the matplotlib backand has been fixed.

Issue #910, opened 2020-05-25
ValueError: array must not contain infs or NaNs


NaSed:
I am getting this error stems from skopt while I am sure the objective function does not generate nan or inf. I have tested the code on my friend's machine and it works but it generates the below error on my machine. DO you have any idea why it happens?

Traceback (most recent call last):
  File "/project/6004228/DR_Project/Code/Residual_v3.py", line 248, in <module>
    verbose=0)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/skopt/optimizer/gp.py", line 295, in gp_minimize
    callback=callback, n_jobs=n_jobs, model_queue_size=model_queue_size)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/skopt/optimizer/base.py", line 303, in base_minimize
    result = optimizer.tell(next_x, next_y)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/skopt/optimizer/optimizer.py", line 483, in tell
    return self._tell(x, y, fit=fit)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/skopt/optimizer/optimizer.py", line 526, in _tell
    est.fit(self.space.transform(self.Xi), self.yi)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/skopt/learning/gaussian_process/gpr.py", line 195, in fit
    super(GaussianProcessRegressor, self).fit(X, y)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/sklearn/gaussian_process/_gpr.py", line 234, in fit
    self.kernel_.bounds))]
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/sklearn/gaussian_process/_gpr.py", line 503, in _constrained_optimization
    bounds=bounds)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/scipy/optimize/_minimize.py", line 610, in minimize
    callback=callback, **options)
  File "/project/6004228/DR_Project/Code/TF220/lib/python3.6/site-packages/scipy/optimize/lbfgsb.py", line 345, in _minimize_lbfgsb
    f, g = func_and_grad(x)
[...]

PR #938, opened 2020-08-30
[WIP] Add prior value checks to Real and Integer Dimension

teffland:
The `Real` and `Integer` dimension subclasses accept invalid `prior` arguments in the constructor (e.g. `Real(... prior=None,...)`) and then silently treat these as `log-uniform` priors in `set_transform`.  This can lead to undesirable behavior, and can cause errors if the bounds are invalid in logspace. I just added value checks to their constructors so users know they've made a mistake early.

This may be related to issue #910 -- I was getting a similar cryptic error `ValueError: array must not contain infs or NaNs` when I realized that I was passing an invalid prior of `None`.  When I switched to `uniform` it now works fine.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -273,21 +273,15 @@ class Real(Dimension):
                 not in [float, np.float, np.float16, np.float32, np.float64]:
             raise ValueError("dtype must be float, np.float"
                              " got {}".format(self.dtype))
+            
+        if prior not in ("uniform", "log-uniform"):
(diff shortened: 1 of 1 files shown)


teffland:
This failed `doc-min-dependencies` check has something to do with an improper matplotlib backend? Looks like an error with the test, not my code.

teffland:
This appears to still not fix the error... thought it worked but was wrong. Will reopen when I figure it out.

holgern:
The doc-min-dependencies bug has been fixed.

Issue #931, opened 2020-08-07
Skopt fails with sklearn 23.2 due to incompatible version comparison


andersbogsnes:
In the `Space` class in `spaces.py` in version 0.7.4 (line [here](https://github.com/andersbogsnes/scikit-optimize/blob/8caf37d9f0b630de9a994539f557d34937c03804/skopt/space/space.py#L764)) - skopt expects scikit-learn to return a tuple for version - it now returns a `Version` class in scikit-learn 0.23.2. 

I see that this check is no longer present on the master branch, so I haven't submitted a pull request, a new release would fix this bug.

I'm sure you know this already, but sklearn has its own `parse_version` function, which would return a compatible comparison object. :smile:

andersbogsnes:
My dependabot traceback for reference:

```
2020-08-07T07:29:50.7795971Z .tox/py/lib/python3.8/site-packages/ml_tooling/baseclass.py:681: in bayessearch
2020-08-07T07:29:50.7796156Z     return self._cross_validated_search(
2020-08-07T07:29:50.7796571Z .tox/py/lib/python3.8/site-packages/ml_tooling/baseclass.py:510: in _cross_validated_search
2020-08-07T07:29:50.7796765Z     self.result: ResultGroup = searcher.search(
2020-08-07T07:29:50.7797226Z .tox/py/lib/python3.8/site-packages/ml_tooling/search/bayes.py:82: in search
2020-08-07T07:29:50.7797412Z     results = [
2020-08-07T07:29:50.7797816Z .tox/py/lib/python3.8/site-packages/ml_tooling/search/bayes.py:83: in <listcomp>
2020-08-07T07:29:50.7798004Z     self._step(optimizer, data, metrics, cv, n_jobs, verbose)
2020-08-07T07:29:50.7798402Z .tox/py/lib/python3.8/site-packages/ml_tooling/search/bayes.py:125: in _step
2020-08-07T07:29:50.7798563Z     params = optimizer.ask()
2020-08-07T07:29:50.7798949Z .tox/py/lib/python3.8/site-packages/skopt/optimizer/optimizer.py:332: in ask
2020-08-07T07:29:50.7799122Z     return self._ask()
2020-08-07T07:29:50.7799514Z .tox/py/lib/python3.8/site-packages/skopt/optimizer/optimizer.py:398: in _ask
2020-08-07T07:29:50.7799697Z     return self.space.rvs(random_state=self.rng)[0]
[...]

PR #939, opened 2020-09-01
[MRG] compatibility update

lucasplagwitz:
Change few lines to make it work with newest versions of scikit-learn and numpy. Short overview:

- new GP in scikit-learn >=0.23: add y_normalize=False
- new GP in scikit-learn >= 0.23:  often throws nan error -> increase initial runs
- numpy MaskArray -> numpy.ma.array
- set matplotlib backend to Agg via export for conf.py


lucasplagwitz:
The scikit-learn NaN (normalize_y=True and one point => std=0) error with GP version >= 0.23 is now mentioned in the  issue board: [scikit-learn#18318](https://github.com/scikit-learn/scikit-learn/issues/18318). In future versions the n_initial_run increasment might not be necessary anymore, but I don't think they do any harm.

holgern:
Thanks @lucasplagwitz , can you add [MRG] into the title of your pull request, when your PR is ready for review? 

As soon as you added [MRG], I will start to review. Thanks again.

lucasplagwitz:
Thanks @holgern for your hint. Add the tag immediately. In case of merge, please be so kind and make a release so we can close #931, too. I saw the instruction on your readme but am not totally sure if it should be new 0.8....
Thanks for reviewing - hope it helped to bring scikit-optimize back to just work after pip/conda install.

PR #941, opened 2020-09-03
[WIP] Fix doc build for sphinx part2

holgern:


holgern:
Merged without review as it is only a simple fix in the doc and it did not change the library itself.

PR #943, opened 2020-09-04
[MRG] Fixes GaussianProcessRegressor for sklearn >= 0.23

holgern:
Fixes changed introduced by https://github.com/scikit-learn/scikit-learn/pull/15782/files


holgern:
I will merge these changes without review, as this fix is imported for sklearn 0.23. Without this fix, all users of sklearn 0.23 will receive wrong results when using skopt.

Issue #950, opened 2020-09-13
BayesSearchCV fails if a parameter is an iterable


Yard1:
When trying to tune a categorical parameter that is an iterable (in this case, `hidden_layers` in MLPClassifier, which are tuples), BayesSearchCV fails with the following exception:

```
~/anaconda3/envs/pycaret/lib/python3.8/site-packages/skopt/searchcv.py in fit(self, X, y, groups, callback)
    690                 n_points_adjusted = min(n_iter, n_points)
    691 
--> 692                 optim_result = self._step(
    693                     X, y, search_space, optimizer,
    694                     groups=groups, n_points=n_points_adjusted

~/anaconda3/envs/pycaret/lib/python3.8/site-packages/skopt/searchcv.py in _step(self, X, y, search_space, optimizer, groups, n_points)
    566 
    567         # convert parameters to python native types
--> 568         params = [[np.array(v).item() for v in p] for p in params]
    569 
    570         # make lists into dictionaries

~/anaconda3/envs/pycaret/lib/python3.8/site-packages/skopt/searchcv.py in (.0)
    566 
    567         # convert parameters to python native types
--> 568         params = [[np.array(v).item() for v in p] for p in params]
    569 
    570         # make lists into dictionaries

~/anaconda3/envs/pycaret/lib/python3.8/site-packages/skopt/searchcv.py in (.0)
    566 
    567         # convert parameters to python native types
--> 568         params = [[np.array(v).item() for v in p] for p in params]
    569 
    570         # make lists into dictionaries

ValueError: can only convert an array of size 1 to a Python scalar
```

I imagine the fix would be to ommit this step for iterables. Will open a PR soon.

PR #952, opened 2020-09-13
[MRG] Fix exception when a categorical param is Iterable

Yard1:
If the possible categorical params are Iterables (lists, tuples, etc. - for example, `hidden_layer_sizes` in MLPClassifier), BayesSearchCV will fail with an exception. This solves this issue.

I imagine that it is not really necessary to convert the list to a numpy array in Categorical `inverse_transform`, but I kept it for consistency and in case there is some behavior that needs arrays specifically that I don't know of. Still, if possible, I'd just return the normal Python list.

Solves https://github.com/scikit-optimize/scikit-optimize/issues/950

### DIFF
--- a/skopt/searchcv.py
+++ b/skopt/searchcv.py
@@ -1,7 +1,7 @@
 try:
-    from collections.abc import Sized
+    from collections.abc import Sized, Iterable
 except ImportError:
-    from collections import Sized
+    from collections import Sized, Iterable
 from collections import defaultdict
 from functools import partial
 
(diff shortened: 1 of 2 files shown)


Yard1:
@kiudee FYI you may want to implement this for your own fork too :)

kiudee:
@Yard1 thanks for the notification. Since [bayes-skopt](https://github.com/kiudee/bayes-skopt) imports from skopt, it should automatically be fixed as soon as a new version of scikit-optimize is released.

Yard1:
@kiudee I think your current implementation overwrites the changes here, that's why I pinged you.

kiudee:
@Yard1 true, I did not see that BayesSearchCV was also changed.
I started a pull request on my side as well:
https://github.com/kiudee/bayes-skopt/pull/87

PR #974, opened 2020-12-04
[MRG] implement a surrogate using lightgbm resolves #922

### LINKED ISSUE #922: [ENH] Replace sklearn GradientBoostingRegressor by lightgbm
Current Situation
=================
The gbrt optimizer uses the sklearn gradient boosting

Enhancement
===========
Replace the sklearn gradient boosting by a lightgbm

Reasoning
=========
This would enhance the running time to a running time comparable to hyperopt (I had similar perf and running than hyperopt on a test classification problem, more extensive testing might be necessary). Moreover, it seems to be more accurate.

Implementation
==============
By setting:

```python
optimizer_kwargs={'base_estimator': LGBMQuantileRegressor(), 'acq_optimizer': "sampling"}
```

with 

```python
import numpy as np

from sklearn.base import BaseEstimator, RegressorMixin
from sklearn.ensemble import GradientBoostingRegressor
from joblib import Parallel, delayed
from lightgbm import LGBMRegressor


def _parallel_fit(regressor, X, y):
    return regressor.fit(X, y)


class LGBMQuantileRegressor(BaseEstimator, RegressorMixin):
    """Predict several quantiles with one estimator.
    This is a wrapper around `LGBMRegressor`'s quantile
    regression that allows you to predict several `quantiles` in
    one go.
    Parameters
    ----------
    quantiles : array-like
        Quantiles to predict. By default the 16, 50 and 84%
        quantiles are predicted.
    base_estimator : LGBMRegressor instance or None (default)
        Quantile regressor used to make predictions. Only instances
        of `GradientBoostingRegressor` are supported. Use this to change
        the hyper-parameters of the estimator.
    n_jobs : int, default=1
        The number of jobs to run in parallel for `fit`.
        If -1, then the number of jobs is set to the number of cores.
    random_state : int, RandomState instance, or None (default)
        Set random state to something other than None for reproducible
        results.
    """

    def __init__(self, quantiles=[0.16, 0.5, 0.84],
                 n_jobs=1, random_state=None):
        self.quantiles = quantiles
        self.random_state = random_state
        self.n_jobs = n_jobs

    def fit(self, X, y):
        """Fit one regressor for each quantile.
        Parameters
        ----------
        X : array-like, shape=(n_samples, n_features)
            Training vectors, where `n_samples` is the number of samples
            and `n_features` is the number of features.
        y : array-like, shape=(n_samples,)
            Target values (real numbers in regression)
        """
        rng = check_random_state(self.random_state)
        base_estimator = LGBMRegressor(objective='quantile')
        # The predictions for different quantiles should be sorted.
        # Therefore each of the regressors need the same seed.
        base_estimator.set_params(random_state=42)
        regressors = []
        for q in self.quantiles:
            regressor = clone(base_estimator)
            regressor.set_params(alpha=q)

            regressors.append(regressor)

        self.regressors_ = Parallel(n_jobs=self.n_jobs, backend='threading')(
            delayed(_parallel_fit)(regressor, X, y)
            for regressor in regressors)

        return self

    def predict(self, X, return_std=False, return_quantiles=False):
        """Predict.
        Predict `X` at every quantile if `return_std` is set to False.
        If `return_std` is set to True, then return the mean
        and the predicted standard deviation, which is approximated as
        the (0.84th quantile - 0.16th quantile) divided by 2.0
        Parameters
        ----------
        X : array-like, shape=(n_samples, n_features)
            where `n_samples` is the number of samples
            and `n_features` is the number of features.
        """
        predicted_quantiles = np.asarray(
            [rgr.predict(X) for rgr in self.regressors_])
        if return_quantiles:
            return predicted_quantiles.T

        elif return_std:
            std_quantiles = [0.16, 0.5, 0.84]
            is_present_mask = np.in1d(std_quantiles, self.quantiles)
            if not np.all(is_present_mask):
                raise ValueError(
                    "return_std works only if the quantiles during "
                    "instantiation include 0.16, 0.5 and 0.84")
            low = self.regressors_[self.quantiles.index(0.16)].predict(X)
            high = self.regressors_[self.quantiles.index(0.84)].predict(X)
            mean = self.regressors_[self.quantiles.index(0.5)].predict(X)
            return mean, ((high - low) / 2.0)

        # return the mean
        return self.regressors_[self.quantiles.index(0.5)].predict(X)
```

## Test

```python
from functools import partial
import time
from skopt import gp_minimize, forest_minimize, dummy_minimize, gbrt_minimize
from skopt.plots import plot_convergence

func = partial(branin, noise_level=2.0)
bounds = [(-5.0, 10.0), (0.0, 15.0)]
n_calls = 60

def run(minimizer, n_iter=5):
    return [minimizer(func, bounds, n_calls=n_calls, random_state=n)
            for n in range(n_iter)]

# Random search
tic = time.time()
dummy_res = run(dummy_minimize)
print("RND running time: {0:2.2f} s".format(time.time()-tic))

# Gaussian processes
tic = time.time()
gp_res = run(gp_minimize)
print("GP running time: {0:2.2f} s".format(time.time()-tic))

# Random forest
tic = time.time()
rf_res = run(partial(forest_minimize, base_estimator="RF"))
print("RF running time: {0:2.2f} s".format(time.time()-tic))

# Gradient boosting
tic = time.time()
gbrt_res = run(gbrt_minimize)
print("GBRT running time: {0:2.2f} s".format(time.time()-tic))

# Lightgbm
tic = time.time()
lgb_res = run(partial(gbrt_minimize, base_estimator=LGBMQuantileRegressor(), acq_optimizer= "sampling"))
print("LGB running time: {0:2.2f} s".format(time.time()-tic))

fig = plt.figure(figsize=(8, 6))
plot = plot_convergence(("dummy_minimize", dummy_res),
                        ("gp_minimize", gp_res),
                        ("forest_minimize('rf')", rf_res),
                        ("gbrt_minimize", gbrt_res),
                        ("lgb_minimize", lgb_res),
                        true_minimum=0.397887, yscale="log")

plot.legend(loc="best", prop={'size':10}, numpoints=1)
```

with outputs:

image
image

Tasks
=====
- Either replace the GradientBoostingQuantileRegressor
- Or keeping both like for ET and RF, making LGBM the default

ThomasBury:
Current Situation
=================
The gbrt optimizer uses the sklearn gradient boosting

Enhancement
===========
add a surrogate using lightgbm

Reasoning
=========
This enhances the running time to a running time comparable to hyperopt (I had similar perf and running than hyperopt on a test classification problem, more extensive testing might be necessary). Moreover, it seems to be more accurate.


skopt/optimizer/lightgbrt.py
```diff
@@ -0,0 +1,186 @@
...
+                                        n_jobs=n_jobs)
+    return base_minimize(func, dimensions, base_estimator,
+                         n_calls=n_calls, n_points=n_points,
+                         n_random_starts=n_random_starts,
+                         n_initial_points=n_initial_points,
+                         initial_point_generator=initial_point_generator,
+                         x0=x0, y0=y0, random_state=random_state, xi=xi,
+                         kappa=kappa, acq_func=acq_func, verbose=verbose,
+                         callback=callback, acq_optimizer="sampling",
+                         model_queue_size=model_queue_size)
```
kernc, line 186: I feel uneasy adding new thin implementations as API. It's just a call to `base_minimize()` with a custom `base_estimator=`:
https://github.com/scikit-optimize/scikit-optimize/blob/debaf61d4fe3fdabdfe23718b55235222b706d01/skopt/utils.py#L386-L402
That's simple enough?

ThomasBury: yes sure, the necessary element is LGBMQuantileRegressor (itself a clone of GradientBoostingRegressor). The rest can be whatever is the simplest and most convenient, it can be gbrt_minismise with the custom base_estimator LGBMQuantileRegressor and just adding it to the list of tree-based models. This introduces a dependency to lightgb but is justified given the speed up and the better accuracy. I read on some blogs that Skopt has been reported slow compared to hyperopt for instance. Which is not the case if one uses lightgbm.

kernc: What I mean is: to use lightgbm as you put it, the users needs but the code:
```py
from skopt.optimizer import base_minimize
...

objective_function = ...
dimensions = ...
base_estimator = LGBMQuantileRegressor(base_estimator=LGBMRegressor(objective='quantile'))

result = base_minimize(objective_function,
                       dimensions,
                       base_estimator,
                       acq_optimizer="sampling")
```
This, kind of, makes the extra dependency _everything but_ justified.

ThomasBury: I'm not sure to get the point because so is `gbrt_minimize` implemented using `base_minimize` and `acq_optimizer="sampling"`

Do you mean you're not comfortable to introduce a dependency to lightgbm? Because the benefits are worth it , briefly, makes things run faster and results more accurate (see the `hyperopt-comparison.py` or the modified `strategy-comparison.py)`
Or if having one more optimizer is not desired? 

In the latter case, we could get rid of `lgbrt_minimize`, keep `LGBMQuantileRegressor` and use 
`gbrt_minimize` with `base_estimator=LGBMQuantileRegressor(), acq_optimizer= "sampling"` and to avoid acquisition errors, append the `tree_estimators` list and add LGBMQuantileRegressor.

```python
def has_gradients(estimator):
    """
    Check if an estimator's ``predict`` method provides gradients.
    Parameters
    ----------
    estimator :
        sklearn BaseEstimator instance.
    """
    tree_estimators = (
            ExtraTreesRegressor, RandomForestRegressor,
            GradientBoostingQuantileRegressor
    )
```

However, not being necessary,having a wrapper (like for `GBRT`, `RF` and `ET`) could be nice, so that the user doesn't need to worry about details using `BayesSearchCV`

```python
opt = BayesSearchCV(clf,
    search_spaces=hh_space,
    scoring='f1',
    cv=StratifiedKFold(random_state=3),
    fit_params=fit_params,
    optimizer_kwargs={'base_estimator': LGBMQuantileRegressor(), 'acq_optimizer': "sampling"},
    n_iter=50,
    n_jobs=-1)
```

kernc: > not comfortable to introduce a dependency to lightgbm?
> Or if having one more optimizer is not desired?

Neither a dependency nor a new thin wrapper function, exactly.

> to worry about details

The _only_ worrying detail here seems to be `acq_optimizer`. I'd rather think about making `has_gradients()` more robust. Maybe determine heuristically on a sufficient toy example, or just flip the default. :thinking:

ThomasBury: > Neither a dependency nor a new thin wrapper function, exactly.

Is there a reason? Because lightGBM gives better and faster results than the scikit-learn GBM, so the HPO using skopt is faster and better. LightGBM is well documented, tested, maintained, has a sklearn API. Technically speaking, it could replace ET and RF as well, even if I didn't test that for skopt. It should not introduce too much technical debt. Do you want to stick to native sklearn only (and what would be the reason to give up the performance gain to not add this dependency? Too hard to maintain? Or not liking MSFT :smile: ?). 

One can think of adding one parameter to gbrt_minimize "type" = 'sklearn' or 'lgbm' (or whatever the GBM implementation as long as it has a well-documented sklearn API) to avoid one more wrapper and have only one per class of models.

kernc: > Is there a reason?

How about just preferring to not add a new project dependency and new files for something that is achieved rather trivially in [but a few lines of user code](https://github.com/scikit-optimize/scikit-optimize/pull/974#discussion_r536833229)? :raised_eyebrow:

I am sorry for your work expended thus far.

ThomasBury: It means that the user has to define `LGBMQuantileRegressor` in his/her script rather than importing it from skopt, which is not convenient. (I don't care about my work per se, if it's not useful then it's not useful, that's it ^^)

What could be the hurdle to add lightGBM as a dependency?

PR #995, opened 2021-02-08
[WIP] Introduce Unscented Bayesian Optimisation

Raziel90:
Implementation of:
- Unscented Expected Improvement;
- Unscented Optimal Intercept; 
- Unscented Bayesian Optimization interface "unscented.py"


Raziel90:
Hi Everyone,

I extended the EI and the selection phase of the next sample point using the unscented transform. 
To do so I used an external library called filterpy. 
I managed to avoid adding that dependency in the core scripts of your library (e.g. `acquisition.py`, `optimiser.py`) including it in the optimizer function file 

I can't pass the ci as the scripts don't find it installed even if I added it to the `requirements.txt`

How should I proceed? 


P.S.
For the future, one may want to think to make the structure of the library easier to extend. Possibly object-oriented.
Having to push code directly in the optimiser class may lead to breaking the library for every extension of the acquisition function or new optimisation scheme. If you are in need of people, I am available to help you with that :D

Cheers :)

Raziel90:
I also modified the only file that made the library unfriendly to python 2.7

PR #1000, opened 2021-02-11
[MRG] add ThresholdStopper()

### LINKED ISSUE #999: EarlyStopper callback given a threshold
It would be nice if we have an EarlyStopper that stops optimization if a func_val is lower (in the case of minimization) than a given threshold

I've implemented and tested it in a private repository
```
class ThresholdStopper(EarlyStopper):
    """
    Stop the optimization if the best func_vals is lower than the given threshold
    """
    def __init__(self, threshold: float) -> bool:
        super(EarlyStopper, self).__init__()
        self.threshold = threshold

    def _criterion(self, result):
        func_vals = np.sort(result.func_vals)
        best = func_vals[0]
        # True if best < threshold given by the user
        return best <= self.threshold
```

GuillaumeSimo:
* add new callback to stop optimization if the func value is lower than a given (by the user) threshold
* fixes #999
* I've implemented and tested this feature in a private repo

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -273,7 +273,8 @@ class DeadlineStopper(EarlyStopper):
 
 class ThresholdStopper(EarlyStopper):
     """
-    Stop the optimization if the best func_vals is lower than the given threshold
+    Stop the optimization if the best func_vals is lower than the given
+    threshold
     """
     def __init__(self, threshold: float) -> bool:
         super(EarlyStopper, self).__init__()
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -271,6 +271,21 @@ class DeadlineStopper(EarlyStopper):
             return None
 
 
+class ThresholdStopper(EarlyStopper):
+    """
+    Stop the optimization if the best func_vals is lower than the given threshold
+    """
+    def __init__(self, threshold: float) -> bool:
+        super(EarlyStopper, self).__init__()
+        self.threshold = threshold
+
+    def _criterion(self, result):
+        func_vals = np.sort(result.func_vals)
+        best = func_vals[0]
+        # True if best < threshold given by the user
+        return best <= self.threshold
+
+
 class CheckpointSaver(object):
     """
     Save current state after each iteration with :class:`skopt.dump`.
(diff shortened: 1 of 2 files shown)


skopt/callbacks.py
```diff
@@ -271,6 +271,22 @@ def _criterion(self, result):
...
+    """
+    Stop the optimization if the best func_vals is lower than the given
+    threshold
+    """
+    def __init__(self, threshold: float) -> bool:
+        super(EarlyStopper, self).__init__()
+        self.threshold = threshold
+
+    def _criterion(self, result):
+        func_vals = np.sort(result.func_vals)
```
kernc, line 284: The snippet below is O(n) (vs. provided O(nlogn)) complexity:
```py
return np.any(result.func_vals < self.threshold)
```

skopt/callbacks.py
```diff
@@ -271,6 +271,22 @@ def _criterion(self, result):
             return None
 
 
+class ThresholdStopper(EarlyStopper):
+    """
+    Stop the optimization if the best func_vals is lower than the given
+    threshold
+    """
+    def __init__(self, threshold: float) -> bool:
```
kernc, line 279: `__init__` doesn't return a bool?

skopt/callbacks.py
```diff
@@ -271,6 +271,22 @@ def _criterion(self, result):
             return None
 
 
+class ThresholdStopper(EarlyStopper):
+    """
+    Stop the optimization if the best func_vals is lower than the given
+    threshold
```
kernc, line 277:
```suggestion
    Stop the optimization when the objective value is lower
    than the given threshold.
```

kernc (changes requested):

skopt/callbacks.py
GuillaumeSimo: You're right
I had to do 
```python
return np.any([val < self.threshold for val in result.func_vals])
```
to avoid
```
TypeError: '<' not supported between instances of 'list' and 'float'
```

skopt/callbacks.py
GuillaumeSimo: done https://github.com/scikit-optimize/scikit-optimize/pull/1000/commits/7c50bfcff70c4cb66de1f47f1c15c451e4dc3ebe

skopt/callbacks.py
kernc: `np.min(result.func_vals) < self.threshold` also works.

kernc (approved):

Issue #1003, opened 2021-02-18
`models` not returned in `optimizer_results_` -> breaking `plot_objective`


AyrtonB:
I think this is likely another issue with the recent versions of skearn (I'm using 0.24.0) but after optimising my model using `BayesSearchCV`, I can see all of the normal scoring results and best parameters, but no models are returned with the `optimizer_results_`.

```
> opt.optimizer_results_[0].models
[]
```

The main issue for me with this is that it then breaks the visualisation of the fit as 

```
> plot_objective(opt.optimizer_results_[0])
```

Leads to the following error:

```
c:\users\user\desktop\side projects\scikit-optimize\skopt\plots.py in plot_objective(result, levels, n_points, n_samples, size, zscale, dimensions, sample_source, minimum, n_minimum_search, plot_dims, show_points, cmap)
    722             if i == j:
    723                 index, dim = plot_dims[i]
--> 724                 xi, yi = partial_dependence_1D(space, result.models[-1],
    725                                                index,
    726                                                samples=samples,

IndexError: list index out of range
```

Any help trying to resolve this would be much appreciated.

AyrtonB:
Similar issue encountered [here](https://github.com/scikit-optimize/scikit-optimize/issues/576#issuecomment-507395022)

kernc:
If you could provide a simple code snippet that someone can use to reproduce the error, that would be helpful.

wronk:
I ran into this issue with `plot_objective` well. It seems like it's tied to the `n_iter` arg. I'm still relatively new to skopt -- is there some minimum for that search below which `BayesSearchCV` doesn't make sense? Maybe it's worth including this in the `plot_objective` docs or adding warning?

For the code below, `n_iter=9` fails for me (with the same error), but a value of `10` plots as expected.

### Code example
```
from skopt import BayesSearchCV
from sklearn.datasets import load_digits
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import train_test_split
from skopt.plots import plot_objective
from matplotlib import pyplot as plt

X, y = load_digits(n_class=10, return_X_y=True)
X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=0.75, test_size=.25, random_state=0)

opt = BayesSearchCV(
    RandomForestClassifier(),
    {'n_estimators':(10, 100, 'uniform'),
     'max_features':(1, 16, 'uniform')},
[...]

PR #1005, opened 2021-02-22
skopt.callbacks.CheckpointCallback can now recycle previous function …

vnherdeiro:
…evaluations. If they exist, they will be loaded from disk and fed into the new optimization as x0, y0. Useful on expensive optimizations ran over several steps (fiddling with space, acquisition functions, more).

### DIFF
--- a/skopt/callbacks.py
+++ b/skopt/callbacks.py
@@ -274,7 +274,7 @@ class DeadlineStopper(EarlyStopper):
 class CheckpointSaver(object):
     """
     Save current state after each iteration with :class:`skopt.dump`.
-
+    Allows to re-use previously computed function evaluations.
 
     Examples
     --------
@@ -285,6 +285,13 @@ class CheckpointSaver(object):
     >>> skopt.gp_minimize(obj_fun, [(-2, 2)], n_calls=10,
     ...                   callback=[checkpoint_callback]) # doctest: +SKIP
 
+
+    When re-using stored results.
+    >>> checkpoint_callback = skopt.callbacks.CheckpointSaver("./result.pkl")
+    >>> x0,y0 = checkpoint_callback.load_previous_evaluations()
+    >>> skopt.gp_minimize(obj_fun, [(-2, 2)], n_calls=10,
+    ...                   x0=x0, y0=y0,
+    ...                   callback=[checkpoint_callback]) # doctest: +SKIP
     Parameters
     ----------
     checkpoint_path : string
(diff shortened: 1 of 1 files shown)


skopt/callbacks.py
```diff
@@ -285,6 +286,13 @@ class CheckpointSaver(object):
     >>> skopt.gp_minimize(obj_fun, [(-2, 2)], n_calls=10,
     ...                   callback=[checkpoint_callback]) # doctest: +SKIP
 
+
+    When re-using stored results.
+    >>> checkpoint_callback = skopt.callbacks.CheckpointSaver("./result.pkl")
+    >>> x0,y0 = checkpoint_callback.load_previous_evaluations()
```
kernc, line 292: As indicated in https://github.com/scikit-optimize/scikit-optimize/pull/998, we also need to warm start `base_estimator=` to continue where we had left off.

skopt/callbacks.py
```diff
@@ -304,3 +312,15 @@ def __call__(self, res):
             The optimization as a OptimizeResult object.
         """
         dump(res, self.checkpoint_path, **self.dump_options)
+
+    def load_previous_evaluations(self):
+        """
+        Loads from disk previously evaluated points.
+        Returns
+        -------
+        X, y : np.ndarray, np.ndarray
```
kernc, line 321: I personally would prefer the method name shorter and the return value a dict, such that can be used more simply:
```py
gp_minimize(..., **checkpoint_callback.load())
```

kernc (changes requested):

vnherdeiro:
@kernc I do like your way of making it more compact, and I have updated the code accordingly. I am also piping the latest state of the kernel/surrogate function space - provided it's the last value of result.models.

skopt/callbacks.py
```diff
@@ -328,5 +328,5 @@ def load(self):
         if os.path.exists(self.checkpoint_path):
             result = load(self.checkpoint_path)
             return {'x0': result.x_iters, 'y0': result.func_vals,
-                    'base_estimator': result.models[-1]}
+                    'base_estimator': result.models[-1] if result.models else None}
```
kernc, line 331: Not sure if this isn't a different bug: https://github.com/scikit-optimize/scikit-optimize/issues/1003

vnherdeiro: From what I have seen `results.model` will not be defined until your evaluation has finished the warm up stage, computation of `n_initial_points` picked by a process (typically random).

PR #1008, opened 2021-03-01
[MRG] Update sklearn-gridsearchcv-replacement.py with missing param

mathcass:
Updates an example which seems like it may be a typo. I think the example at sklearn-gridsearchcv-replacement.py may have a typo in the LinearSVC block. This should fix that example by using the Real value sampler since that's what is implied by the rest of the example.

---

The example of using BayesSearchCV seems like it's missing a function call within the LinearSVC section. As-is, it just seems like it's a tuple argument when the SVC example implies that it should be a Real variable sampling function.


mathcass:
Aha, it looks like this may be valid syntax for the constructor. I'll close this since this is correct.

PR #1020, opened 2021-04-03
[MRG] Check whether prior values are valid.

### LINKED ISSUE #1017: skopt.Space prior does not check argument validity
https://github.com/scikit-optimize/scikit-optimize/blob/master/skopt/space/space.py#L292

The "transform" part is checked for validity, but not the "prior" part. As a result, anything that isn't "uniform" will silently be treated as "log-uniform".

(But this has no effect due to other bugs, which I'll also file)

gcp:
Fixes #1017.

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -254,6 +254,9 @@ class Real(Dimension):
         if high <= low:
             raise ValueError("the lower bound {} has to be less than the"
                              " upper bound {}".format(low, high))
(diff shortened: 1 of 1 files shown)


kernc:
Why not check `prior` in the constructor?

gcp:
@kernc Sounds logical, fixed.

glouppe:
Thanks!

PR #1023, opened 2021-04-20
Fix numpy Deprecation errors

xmatthias:
This will fix some numpy depreciation errors that appear due to the usage of float.

```
  `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

It's also fixing `skopt/learning/gaussian_process/kernels.py:319: DeprecationWarning: invalid escape sequence \s` - although i'm not 100% certain how this fix will come out at the docs.
Unfortunately, i'm uncertain how to build / verify this - as it doesn't seem to be documented.

I'm hoping that CI passes, as i found no way to get `skopt/tests/test_gp_opt.py` to pass all tests (not with master, not with the modifications ...).

### DIFF
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -244,13 +244,13 @@ class Real(Dimension):
     name : str or None
         Name associated with the dimension, e.g., "learning rate".
 
-    dtype : str or dtype, default=np.float
+    dtype : str or dtype, default=np.float64
         float type which will be used in inverse_transform,
         can be float.
 
     """
     def __init__(self, low, high, prior="uniform", base=10, transform=None,
-                 name=None, dtype=np.float):
+                 name=None, dtype=float):
         if high <= low:
             raise ValueError("the lower bound {} has to be less than the"
                              " upper bound {}".format(low, high))
@@ -273,8 +273,8 @@ class Real(Dimension):
                              "or 'float64'"
                              " got {}".format(self.dtype))
         elif isinstance(self.dtype, type) and self.dtype\
-                not in [float, np.float, np.float16, np.float32, np.float64]:
-            raise ValueError("dtype must be float, np.float"
+                not in [float, np.float16, np.float32, np.float64]:
+            raise ValueError("dtype must be float, np.float64"
                              " got {}".format(self.dtype))
 
         if transform is None:
--- a/skopt/space/space.py
+++ b/skopt/space/space.py
@@ -244,7 +244,7 @@ class Real(Dimension):
     name : str or None
         Name associated with the dimension, e.g., "learning rate".
 
-    dtype : str or dtype, default=np.float64
+    dtype : str or dtype, default=float
         float type which will be used in inverse_transform,
         can be float.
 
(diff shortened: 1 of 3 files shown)


skopt/space/space.py
```diff
@@ -244,13 +244,13 @@ class Real(Dimension):
     name : str or None
         Name associated with the dimension, e.g., "learning rate".
 
-    dtype : str or dtype, default=np.float
+    dtype : str or dtype, default=np.float64
```
kernc, line 247: Why not just `float`, since that's what the user will become accustomed to seeing?

skopt/space/space.py
```diff
@@ -273,8 +273,8 @@ def __init__(self, low, high, prior="uniform", base=10, transform=None,
                              "or 'float64'"
                              " got {}".format(self.dtype))
         elif isinstance(self.dtype, type) and self.dtype\
-                not in [float, np.float, np.float16, np.float32, np.float64]:
-            raise ValueError("dtype must be float, np.float"
+                not in [float, np.float16, np.float32, np.float64]:
```
kernc, line 276: I'd promote the condition to:
```py
np.issubdtype(self.dtype, np.floating)
```
or something equivalent.

skopt/space/space.py
```diff
@@ -272,9 +272,9 @@ def __init__(self, low, high, prior="uniform", base=10, transform=None,
             raise ValueError("dtype must be 'float', 'float16', 'float32'"
                              "or 'float64'"
                              " got {}".format(self.dtype))
-        elif isinstance(self.dtype, type) and self.dtype\
-                not in [float, np.float, np.float16, np.float32, np.float64]:
-            raise ValueError("dtype must be float, np.float"
+        elif isinstance(self.dtype, type) and \
+                np.issubdtype(type(self.dtype), np.floating):
+            raise ValueError("dtype must be float, np.float64"
```
kernc, line 277: To match the previous check, the condition should probably be negated, though. :smile:

Can you also amend the error string?

skopt/space/space.py
```diff
@@ -272,9 +272,9 @@ def __init__(self, low, high, prior="uniform", base=10, transform=None,
             raise ValueError("dtype must be 'float', 'float16', 'float32'"
                              "or 'float64'"
                              " got {}".format(self.dtype))
-        elif isinstance(self.dtype, type) and self.dtype\
-                not in [float, np.float, np.float16, np.float32, np.float64]:
-            raise ValueError("dtype must be float, np.float"
+        elif isinstance(self.dtype, type) and \
+                not np.issubdtype(self.dtype, np.floating):
+            raise ValueError("dtype must be a np.floating subtype"
```
kernc, line 277:
```suggestion
            raise ValueError("dtype must be a np.floating subtype;"
```

kernc (approved):

skopt/space/space.py
xmatthias: yeah sure, sorry for the long back and forth!

PR #1040, opened 2021-06-21
[MRG] Fixed grammatical error in docs: 'less' vs 'fewer'

samuelstevens:
Just a minor grammar error that I noticed in the docs. "Fewer" is used with countable things (like evaluations) while "less" is used with amounts (like time).

Thanks for a great library!


kernc:
Thanks!

Issue #302, opened 2017-01-24
Tests use a lot of duplicate points


betatim:
Since we added the warning about re-evaluating the objective at the same point our tests are full of warnings:
```
scikit-optimize/skopt/optimizer/optimizer.py:174: UserWarning: The objective has been evaluated at this point before.
  warnings.warn("The objective has been evaluated at this point before.")
```

We should think whether we want t mute the warning or improve the tests.

glouppe:
or we might check why some points are reevaluated.

Henri-Lo:
How can you suppress the warning? I am running the very first example on the doc and I am getting the warning, perhaps you could update the example in the doc with suppressed warning?

UserWarning: The objective has been evaluated at this point before.
  warnings.warn("The objective has been evaluated "

https://github.com/scikit-optimize/scikit-optimize

TropComplique:
The same is here. A lot of annoying warnings. What to do with them?

TropComplique:
@Henri-Lo So, you can suppress the warnings by using:
```
import warnings
warnings.filterwarnings('ignore', message='The objective has been evaluated at this point before.')
```
But is it the right thing to do?

MechCoder:
The right thing to do for the example, tests or in general?

TropComplique:
Hi. 

In general. When using `skopt` for hyperparameter optimization, for example.  
Are there warnings because I'm using `skopt` in a wrong way or because `skopt` is not fully developed yet?

MechCoder:
No package is fully-developed ever :) Can you post a minimal code-snippet?

It is not too uncommon in-case you are optimizing over a space that is completely categorical for instance.

betatim:
We should try and remove the warnings from the tests by either choosing bigger spaces or filtering them where we know that it isn't a problem.

In general the warning tells you that "there might be a problem". Usually evaluating the same point multiple times isn't something you want to do, but if you have a (very) noisy objective it makes sense to do so. In my experience I usually get the warning when I get the type of one of the dimensions wrong, for example `(2, 2.1443)` as a dimension gives you an `Integer` dimension (-> #391) when really I wanted a floating point dimension.

AzraelDD:
Hi everybody,

is this still being considered? There should be (if this is possible) some sort of input flag for ``optimizer.ask()`` that forces it to avoid points that have been evaluated before. Instead it should evaluate the next "best" / "most interesting" points or maybe even some totally random points.

[...]

PR #1050, opened 2021-08-19
[WIP] Evaluate random point instead of using point twice

jnettels:
#302 discusses a long standing issue where the ask() function to "Query point or multiple points at which objective should be evaluated" will just repeatedly return the same point(s) to evaluate. In cases where the objective always returns the same results, the solver can even get stuck in a loop.

This PR aims to mitigate the issue. Instead of just warning the user about evaluating a point twice, use a new random point instead (and include this information in the warning).

This is not a good solution. Ideally, the next best point should be evaluated instead, but I do not know how to do that.

Also, in #302 a user said "Usually evaluating the same point multiple times isn't something you want to do, but if you have a (very) noisy objective it makes sense to do so." So the solution presented here might actually cause a problem for users who want the same point to be evaluated multiple times. Since this is up for debate, I am tagging it as [WIP], even though I think it is "complete" in the way that I am intending it to be.

Maybe this just stays here as a reference for users who want a very easy workaround that they can implement themselves.

### DIFF
--- a/skopt/optimizer/optimizer.py
+++ b/skopt/optimizer/optimizer.py
@@ -446,8 +446,11 @@ class Optimizer(object):
             min_delta_x = min([self.space.distance(next_x, xi)
                                for xi in self.Xi])
             if abs(min_delta_x) <= 1e-8:
-                warnings.warn("The objective has been evaluated "
-                              "at this point before.")
+                next_x_new = self.space.rvs(random_state=self.rng)[0]
+                warnings.warn("The objective has been evaluated at point {} "
+                              "before, using {} instead"
+                              .format(next_x, next_x_new))
+                next_x = next_x_new
 
             # return point computed from last call to tell()
             return next_x


skopt/optimizer/optimizer.py
```diff
@@ -446,8 +446,11 @@ def _ask(self):
             min_delta_x = min([self.space.distance(next_x, xi)
                                for xi in self.Xi])
             if abs(min_delta_x) <= 1e-8:
```
kernc, line 448: > Ideally, the next best point should be evaluated instead, but I do not know how to do that.

I guess you would have to move this check to where `self._next_x` is _set_ and recompute it there.

QuentinSoubeyran:
Since there are use case for both the old and the new behaviours, I would argue for a flag in the optimizer's `__init__` that defaults to the old behaviour. This also preserve backward compatibility.

kernc:
> will just repeatedly return the same point(s) to evaluate. In cases where the objective always returns the same results, the solver can even get stuck in a loop.

@QuentinSoubeyran What is the alleged use case for this old behavior?

QuentinSoubeyran:
> Also, in #302 a user said "Usually evaluating the same point multiple times isn't something you want to do, but if you have a (very) noisy objective it makes sense to do so."

@kernc This is what I was referring to. If such use cases are rare, defaulting to the new behaviour might be better, now that I think about it.

jnettels:
It has been a while since my first commit to this PR. I realized that this issue is still not resolved and I took another stab at it.
I also realized that we have access to the proposed next point candidates from all acquisition functions, when ``gp_hedge`` is used. So my code now does the following:
When a duplicate point is proposed as the next candidate, test if one of the acquisition functions proposed a candidate that has not been used yet. If yes, use that point instead. If no, fall back to a random point and warn about it.

This increases the chance of replacing a duplicate candidate with a useful candidate a lot. It still cannot prevent using random points as a fallback. I still do not know if it is possible to force the acquisition functions themselves to not create duplicates in the first place.
But nevertheless this increases the speed of reaching a solution for some of my tests.

I included a bool ``avoid_duplicates`` that could be used (by someone else?) to somehow make this behaviour selectable as an option for the user, but I don't know where to best put that.

PR #1063, opened 2021-09-29
[MRG] Fix GaussianProcessRegressor for scikit-learn 1.0

### LINKED ISSUE #1061: Scikit-optimize master GaussianProcessRegressor is incompatible with scikit-learn 1.0
`scikit-learn` 0.19 depreciated the `y_train_mean` attribute of `GaussianProcessRegressor`. At the time it was made private, and so scikit-optimize continued to rely on it. #467 was never merged to properly fix that.
Scikit-learn 1.0 removed the attribute, breaking scikit-optimize.

**Version**
- `python==3.9.2`
- `scikit-learn==1.0`
- `scikit-optimize==0.9.dev0` (installed with `pip install git+https://github.com/scikit-optimize/scikit-optimize.git`)

**How to reproduce**
```bash
mkdir test
cd test
python3 -m venv venv
source venv/bin/activate
pip install "scikit-learn==1.0" git+https://github.com/scikit-optimize/scikit-optimize.git
```

Write the following script to `test.py`:
```python
import scipy
import sklearn
import skopt
from sklearn import model_selection, svm
from sklearn.datasets import load_digits

print("scipy:", scipy.__version__)
print("scikit-learn:", sklearn.__version__)
print("scikit-optimize:", skopt.__version__)

X, y = load_digits(n_class=10, return_X_y=True)
X_train, X_test, y_train, y_test = model_selection.train_test_split(
    X, y, train_size=0.75, test_size=0.25, random_state=0
)

opt = skopt.BayesSearchCV(
    svm.SVC(),
    {
        "C": (1e-6, 1e6, "log-uniform"),
        "gamma": (1e-6, 1e6, "log-uniform"),
        "kernel": ["linear", "rbf"],
    },
    n_iter=32,
    cv=3,
)

opt.fit(X_train, y_train)
print("Best train score:", opt.best_score_)
print("Tests score:", opt.score(X_test, y_test))
```

and run it in the test venv:
```bash
python test.py
```

QuentinSoubeyran:
This PR fix compatibility with scikit-learn 1.0:

- Fixes #1061
- Fix tests using outdated scikit-learn API to load datasets

### DIFF
--- a/skopt/learning/gaussian_process/gpr.py
+++ b/skopt/learning/gaussian_process/gpr.py
@@ -5,7 +5,6 @@ from scipy.linalg import cho_solve
 from scipy.linalg import solve_triangular
 
 import sklearn
-from packaging.version import parse as parse_version
 from sklearn.gaussian_process import GaussianProcessRegressor as sk_GaussianProcessRegressor
 from sklearn.utils import check_array
 
@@ -14,10 +13,6 @@ from .kernels import Sum
 from .kernels import RBF
 from .kernels import WhiteKernel
 
-# Cache the version objects
-_SKLEARN_VERSION = parse_version(sklearn.__version__)
-_V_0_23 = parse_version("0.23")
-_V_0_19 = parse_version("0.19")
 
 def _param_for_white_kernel_in_Sum(kernel, kernel_str=""):
     """
@@ -229,10 +224,10 @@ class GaussianProcessRegressor(sk_GaussianProcessRegressor):
         self.K_inv_ = L_inv.dot(L_inv.T)
 
         # Fix deprecation warning #462
-        if _SKLEARN_VERSION >= _V_0_23:
+        if sklearn.__version__ >= "0.23":
             self.y_train_std_ = self._y_train_std
             self.y_train_mean_ = self._y_train_mean
-        elif _SKLEARN_VERSION >= _V_0_19:
+        elif sklearn.__version__ >= "0.19":
             self.y_train_mean_ = self._y_train_mean
             self.y_train_std_ = 1
         else:
--- a/skopt/learning/gaussian_process/gpr.py
+++ b/skopt/learning/gaussian_process/gpr.py
@@ -229,7 +229,6 @@ class GaussianProcessRegressor(sk_GaussianProcessRegressor):
         self.K_inv_ = L_inv.dot(L_inv.T)
 
         # Fix deprecation warning #462
-
         if _SKLEARN_VERSION >= _V_0_23:
             self.y_train_std_ = self._y_train_std
             self.y_train_mean_ = self._y_train_mean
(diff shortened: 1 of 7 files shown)


skopt/learning/gaussian_process/gpr.py
```diff
@@ -224,10 +229,10 @@ def fit(self, X, y):
         self.K_inv_ = L_inv.dot(L_inv.T)
 
         # Fix deprecation warning #462
-        if int(sklearn.__version__[2:4]) >= 23:
+        if _SKLEARN_VERSION >= _V_0_23:
             self.y_train_std_ = self._y_train_std
             self.y_train_mean_ = self._y_train_mean
-        elif int(sklearn.__version__[2:4]) >= 19:
+        elif _SKLEARN_VERSION >= _V_0_19:
```
kernc, line 235: If you rather simply do:
```py
if sklearn.__version__ >= '0.23':
    ...
if sklearn.__version__ >= '0.19':
    ...
```
we won't require a dependency on package [_packaging_](https://pypi.org/project/packaging/).

QuentinSoubeyran: I'm reluctant to use string comparison instead of proper version comparison, since bad version comparison is the source of #1061 .

I incorrectly though `scikit-optimize` indirectly depended on `packaging` already, but this is not the case (`packaging` is required by `pytest`).

Apparently both [`scipy`](https://github.com/scipy/scipy/blob/master/scipy/_lib/_pep440.py) and [`scikit-learn`](https://github.com/scikit-learn/scikit-learn/pull/19826) chose to vendor `packaging.parse_version` for this purpose. Maybe this is the best route ? Or maybe we can use their vendored version, since `scikit-optimize` depends on both ? I think this will increase the minimum supported `scikit-learn` version, I have a vague memory that older versions didn't vendor it.

kernc: I _don't_ think this is the best route. Extra dependencies for a two-line fix rarely are.

Do you see a problem with the above proposal? It's a common idiom. Previously it was implemented incorrectly.

QuentinSoubeyran: I think scipy and scikit-learn copy the code of the single module they need to avoid the dependency.
I don't really see a problem with the above idiom, I'm just not sure if alphabetical ordering matches version ordering. Since it is a common idiom, it should be robust enough, I'll go with that.

kernc: sklearn 0.2 and 0.20 are far enough apart for us to not care. :smile: 
It should be good enough for this release. Which is kinda urgent.

QuentinSoubeyran: That's fair, let's not overcomplicate that

kernc:
Many thanks!

Issue #1065, opened 2021-09-29
integer lower_bound and upper_bound result in ignoring "prior", and bug if are not of the same type


Abdelgha-4:
In skopt.Optimizer [docs](https://scikit-optimize.github.io/stable/modules/generated/skopt.Optimizer.html) it said:

> - a (lower_bound, upper_bound, "prior") tuple (for Real dimensions), ...

however if lower_bound and upper_bound are integers it generate an Integer dimension
a reproduceable example:
```
from skopt import BayesSearchCV
from sklearn.datasets import load_iris
from sklearn.svm import SVC

X, y = load_iris(return_X_y=True)

searchcv = BayesSearchCV(
    SVC(gamma='scale'),
    search_spaces={'C': (1, 100, 'log-uniform')},
    n_iter=10,
    cv=5,
    scoring='f1_macro',
    random_state=42
)
searchcv.fit(X, y)
print(searchcv.cv_results_['param_C'])
```
it shows a list of integer values.
**Another bug:**
with the same code above, when replacing `(1, 100, 'log-uniform')` with `(1.1, 100, 'log-uniform')`, it produce a `ValueError: All values shouldbe greater than 0.041393`

Abdelgha-4:
> I believe this is intended
Ok then I'll just suggest a better documentation of how to specify the dimensions, because I ran an hour long search before discovering this behavior, and I think it'll help other users to avoid this misunderstanding.

QuentinSoubeyran:
The other bugs mentioned are indeed bugs. I also think list should always lead to `Categorical` object. What do you think @kernc @glouppe ?
See #1067

PR #1067, opened 2021-09-30
[WIP] improved check_dimension type inference, added warnings, tests

QuentinSoubeyran:
#1065 found that list could be inferred to dimensions other than `Categorical`, and that mixing `int` and `float` in the bounds produced incorrect results.

This PR fixes `skopt.space.check_dimension` to actually follow its own documentation:
- `list` always lead to `Categorical` object
- `tuple` lead to `Integer` or `Real` object, as inferred by the bound types
- numpy arrays are treated as lists, and documented


QuentinSoubeyran:
@kernc @glouppe This is ready for review

skopt/space/space.py
```diff
@@ -55,13 +55,19 @@ def check_dimension(dimension, transform=None):
         Search space Dimension.
         Each search dimension can be defined either as
 
-        - a `(lower_bound, upper_bound)` tuple (for `Real` or `Integer`
-          dimensions),
-        - a `(lower_bound, upper_bound, "prior")` tuple (for `Real`
-          dimensions),
-        - as a list of categories (for `Categorical` dimensions), or
-        - an instance of a `Dimension` object (`Real`, `Integer` or
-          `Categorical`).
```
kernc, line 64: Where did this disappear?

skopt/space/space.py
```diff
@@ -55,13 +55,19 @@ def check_dimension(dimension, transform=None):
...
+        - a list or numpy array of choices, for `Categorical` dimension
+        - a 2-, 3- or 4-tuple, for `Real` and `Integer` dimensions, of
+          the form ``(low, high [, prior [, base]])`` (values in square
+          brackets are optional). If both ``low`` and ``high`` are integral
+          numbers (as per the `number.Integral`), a `Integer` dimension is
+          returned, else a `Real` dimension is returned.
+
+          NOTE: For a transitionary period, tuple, list and array currently all
```
kernc, line 65: [`.. note::` in ReST](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html).

skopt/space/space.py
```diff
@@ -84,11 +90,44 @@ def check_dimension(dimension, transform=None):
...
+                      "upcoming versions of scikit-optimize, it will be "
+                      f"inferred to {new_dim}. See the documentation of the "
+                      "check_dimension function for the upcoming API.")
+    return old_dim
+
+
+def _check_dimension(dimension, transform=None):
+    if isinstance(dimension, Dimension):
+        return dimension
+    elif isinstance(dimension, (list, np.ndarray)):
```
kernc, line 109: I don't like that a tuple of strings is no longer supported. Why does it need to be a list?

skopt/space/space.py
```diff
@@ -84,11 +90,44 @@ def check_dimension(dimension, transform=None):
...
+        return Categorical(dimension, transform=transform)
+    elif isinstance(dimension, tuple) and 1 <= len(dimension) <= 4:
+        low, high, *args = dimension
+        if (isinstance(low, numbers.Integral)
+                and isinstance(high, numbers.Integral)):
+            return Integer(int(low), int(high), *args, transform=transform)
+        elif isinstance(low, numbers.Real) and isinstance(high, numbers.Real):
+            return Real(float(low), float(high), *args, transform=transform)
```
kernc, line 117: `low` and `high` are already `int` (or `float`). Why cast?

skopt/space/space.py
```diff
@@ -99,41 +138,47 @@ def check_dimension(dimension, transform=None):
...
-                for d in dimension]):
+        if any(isinstance(d, (str, bool)) or isinstance(d, np.bool_)
+                for d in dimension):
             return Categorical(dimension, transform=transform)
-        elif all([isinstance(dim, numbers.Integral) for dim in dimension]):
-            return Integer(*dimension, transform=transform)
-        elif any([isinstance(dim, numbers.Real) for dim in dimension]):
-            return Real(*dimension, transform=transform)
+        elif all(isinstance(dim, numbers.Integral) for dim in dimension):
+            return Integer(*map(int, dimension), transform=transform)
```
kernc, line 145: Likewise, why this `map()`?

skopt/space/space.py
QuentinSoubeyran: The `float` cast is because one of the boundaries might be an `int`, and a `Real` dimension asks for two `float` in its doc.
The `int` cast was to support other integral types like `bool` or `np.int32`.

Since `int` is not a sublcass of `float`, I think the `float` cast is necessary to respect `Real`'s interface. I'm not too sure about the `int` cast, I mostly did it for symmetry.

skopt/space/space.py
QuentinSoubeyran: see above

skopt/space/space.py
QuentinSoubeyran: oups, my bad !

skopt/space/space.py
QuentinSoubeyran: The idea was to tighten the interface to prevent user errors leading to valid, but unwanted, dimension objects. This is most surely inspired by my uses of type-checkers in python, as the above makes the signature of `check_dimension` not depend on the actual values contained in the passed sequence:
- `list` becomes a `Categorical`
- `tuple[int, int]` (and the 3- and 4-long forms) become `Integer`
- `tuple[int, float]`, `tuple[float, int]` and `tuple[float, float]` (and corresponding 3 and 4-long forms) become `Real`

It was also to avoid an incorrect tuple silently becoming a `Categorical`, e.g.
- `(0, 1, "uniform", base)` where `base="10"` was no parsed from a data source, silently returning a `Categorical` instead of the intended `Integer`
- `(0, 1, *params)` where `params=(5, "log-uniform", 2)` because the semantic/content of `params` is wrong, silently falling back to a `Categorical`
- `[True, False]` being inferred to an `Integer` (though the old implementation has a dedicated branch to avoid that).

We could make it so that `tuple`s that fail the `Integer`/`Real` inference default to a `Categorical` as before. This would also allow to use any `Iterable`, like so:
```python
if isinstance(dimension, tuple) and 2 <= len(dimension) <= 4:
    low, high, *args = dimension
    if (not args or isinstance(args[0], str)) and (len(args) < 2 or isinstance(args[1], int)):
        if isinstance(low, numbers.Integral) and isinstance(high,  numbers.Integral):
              return Integer(...)
        elif isinstance(low, numbers.Real) and isinstance(high, numbers.Real):
            return Real(...)
# not an else, fallback
if isinstance(dimension, Iterable):
    return Categorical(...)
raise ValueError(...)
```

Do you prefer that other type-tightness ?

QuentinSoubeyran:
@kernc I made tuples fall back to `Categorical` dimensions, with a warning if the tuple seems to be wrong. Is this better ?

PR #1088, opened 2021-10-29
[MRG] Fix dimension mismatch with gp_minimize in Matern kernel when nu=0.5

mariegold:
When running `gp_minimize` on a GP using a Matern kernel with `nu=0.5`, there is a shape mismatch with `x`: expects `(n,)` but `gp_minimize` passes `(1, n)`. This is a problem when applying the mask on line 127.


zbirobin:
This commit is valid and useful.

PR #1123, opened 2022-08-30
Replace deprecated np.int type, causing warnings.

preritdas:
This small change corrects the following warning: 

DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use 
`int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.

There were only two instances of `np.int` in the `space/transformers.py` module, both of which have been changed to `np.int64`; of course, this can be another precision type instead, ex. `np.int32` or `np.int16`. The two instances of `np.int` corrected are shown below.

https://github.com/preritdas/scikit-optimize/blob/bbb7a1da30f7c0ef3763a80fe63142cb8baa7ed9/skopt/space/transformers.py#L262-L275

### DIFF
--- a/skopt/learning/forest.py
+++ b/skopt/learning/forest.py
@@ -435,7 +435,7 @@ class ExtraTreesRegressor(_sk_ExtraTreesRegressor):
         mean = super(ExtraTreesRegressor, self).predict(X)
 
         if return_std:
-            if self.criterion != "mse":
+            if self.criterion not in {"mse", "friedman_mse"}:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


preritdas:
It seems a circlici test is failing with an error caused by a too-early version of Python, 3.7. 

RuntimeError: Scikit-learn requires Python 3.8 or later. The current Python version is 3.7.13 installed in /home/circleci/miniconda/envs/testenv/bin/python.

PR #1139, opened 2023-01-20
[MRG] Remove deprecated np.int type

MalcolmMielle:
np.int was fully removed in numpy 1.24.

See release notes for 1.20: https://numpy.org/doc/stable/release/1.20.0-notes.html


BartWesthoff:
Hi man great work. I was looking for this issue and it was solved. Very strange that this package is not supported anymore as it seems.

MalcolmMielle:
There is actually other PR solving this issue already. I feel like this package is not maintained anymore, which is too bad.

PR #1146, opened 2023-02-12
Move the metadata into `PEP 621`-compliant `pyproject.toml`.

KOLANICH:
Requires #1145

Also this commit gets rid of the most build-time dependencies (module is no longer imported when the wheel is being built, instead modern versions of `setupttols` can read vars using static analysis. Requires a newer `setuptools`. `distutils` support is dropped.


KOLANICH:
For now we don't delete `setup.py`, but it should be done. It may require changes in CI scripts, I have no time to debug that, so leave it up to the maintainers of this repo.

CI fails because it seems the CI scripts got rotten. It is not up to me to fix them.

`setup.cfg` is kept to store prefs for other tools.

PR #1148, opened 2023-02-14
`np.int` -> `int`

valtron:
`np.int is int` and it was deprecated in numpy 1.20: https://numpy.org/doc/1.20/release/1.20.0-notes.html#deprecations


Csaba-emnify:
Would highly appreciate this change

shariharan99:
We would love this fix too!

PR #1150, opened 2023-02-16
[MRG] Change ExtraTreesRegressor criterion from 'mse' to 'squared_error'

jotoak:
Changes are made to be compatible with scikit-learn version 1.0. Version bump of both scipy and numpy is also done. This pull request also requires [#1148](https://github.com/scikit-optimize/scikit-optimize/pull/1148)

### DIFF
--- a/skopt/learning/forest.py
+++ b/skopt/learning/forest.py
@@ -27,7 +27,7 @@ def _return_std(X, trees, predictions, min_variance):
     -------
     std : array-like, shape=(n_samples,)
         Standard deviation of `y` at `X`. If criterion
-        is set to "mse", then `std[i] ~= std(y | X[i])`.
+        is set to "squared_error", then `std[i] ~= std(y | X[i])`.
 
     """
     # This derives std(y | x) as described in 4.3.2 of arXiv:1211.0906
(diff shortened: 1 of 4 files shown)


skopt/learning/forest.py
```diff
@@ -194,7 +194,7 @@ class RandomForestRegressor(_sk_RandomForestRegressor):
     .. [1] L. Breiman, "Random Forests", Machine Learning, 45(1), 5-32, 2001.
 
     """
-    def __init__(self, n_estimators=10, criterion='mse', max_depth=None,
+    def __init__(self, n_estimators=10, criterion='squared_error', max_depth=None,
                  min_samples_split=2, min_samples_leaf=1,
```
ShamoX: To respect the 80 line width restriction, I propose to put `max_depth` on the next line.

```suggestion
    def __init__(self, n_estimators=10, criterion='squared_error',
                 max_depth=None, min_samples_split=2, min_samples_leaf=1,
```

ShamoX:
It worked for me to use this patch. I simply used @jotoak repository as source of the package (requirements.txt file of my project):

```requirements.txt
scikit-optimize @ git+https://github.com/jotoak/scikit-optimize.git@master
```

For reference, the error I had occurred after the initialization steps and was:

```
sklearn.utils._param_validation.InvalidParameterError: The 'criterion' parameter of ExtraTreesRegressor must be a str among {'poisson', 'absolute_error', 'squared_error', 'friedman_mse'}. Got 'mse' instead.
```

I am running this on an Apple Silicon M2, and so I couldn't get back to an old version of sklearn.

skopt/learning/forest.py
jotoak: Looks good

PR #1156, opened 2023-03-10
Implement MES and PVRS acquisition functions.

gcp:
Add MES and PVRS information theory based acquisition functions, based on the papers and implementations in bayes-skopt (bask).

Extend the gp_hedge to include Max-value Entropy Search.

Change optimizer selections to not use F-BFGS-B for the new functions without gradients.

### DIFF
--- a/skopt/acquisition.py
+++ b/skopt/acquisition.py
@@ -2,6 +2,8 @@ import numpy as np
 import warnings
 
 from scipy.stats import norm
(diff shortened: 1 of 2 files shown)


gcp:
The implementations are based on the ones in [bayes-skopt](https://github.com/kiudee/bayes-skopt), the original author @kiudee has agreed to re-license the relevant code under BSD 3-Clause to ease merging here.

PR #1185, opened 2023-10-25
Update transformers.py

sergsb:
fix numpy int error -> int32


aroyphillips:
I think you can just change it to int instead of np.int32. Not sure if that will fix those tests, but that's the official equivalency according to numpy: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

Issue #1117, opened 2022-07-13
New release?


tvdboom:
@kernc Hi. I see there are many PRs that would greatly improve the library ready to merge to `master`. Any idea when this will happen and a new release will be launched? I think many people and other libraries would benefit from this. I could help if that's needed. Thanks

kernc:
For a start, getting something like https://github.com/scikit-optimize/scikit-optimize/pull/1074 in and finally have a working CI again would be great. One of these days I might just get round to it.

PR #1074, opened 2021-10-04
Move CI to GitHub Actions

kernc:
Fixes https://github.com/scikit-optimize/scikit-optimize/issues/1060
Closes https://github.com/scikit-optimize/scikit-optimize/pull/1072
Closes  https://github.com/scikit-optimize/scikit-optimize/issues/1068
Closes https://github.com/scikit-optimize/scikit-optimize/pull/1069

This took a while, but it's often better late than never. :sweat_smile:

The PR looks huge, but it's namely about:
* migrating the whole of CI onto GitHub Actions (purging TravisCI and CircleCI machinery so we don't have dead code lying around),
* improving setup.py to aid in releasing (specifying dependencies in one single place, version control by version control etc.),
* cleaning the code to conform to (hereby finally active!) linting,
* some documentation consolidation and improvements (README.rst, CONTRIBUTING.md, ISSUE_TEMPLATE.md ...) to help new users.

I tried my best to simplify, yet continue @holgern's work—what happened, man, have you died?! :D

It's not too pretty, but it's what I can offer. The commits have been roughly squashed by function. The PR is at this point certainly ready for review. You can see all the checks passing on a [PR to my fork, here](https://github.com/kernc/scikit-optimize/pull/3).

Issue #847, opened 2020-02-12
New maintainers


betatim:
Hello all,

scikit-optimize has seen a lot of new contributions recently from @holgern. He contacted the current maintainers (@glouppe, @iaroslav-ai and @MechCoder) asking if he could help out with maintenance.

Welcome Holger!

The guidelines for [interacting](CONDUCT.md) and [contributing](CONTRIBUTING.md) remain the same.

A main task for maintaining this project is around growing the number of people who actively contribute to the maintenance of the project in order to keep the level of quality high. This is a hard task! Depending on what part of the history of the project you look at the past maintainers did a good or a bad job at this. I look forward to seeing how new maintainers tackle this challenge and keep growing the community.

As part of this I would be happy to see more people creating PRs and increasing their involvement so that new people can join Holger in maintaining this project.

kernc:
I would like to nominate myself as a new maintainer. :raised_hand_with_fingers_splayed: 

It's a real shame if not wholly unacceptable to see a such a well-rounded drop-in library stall unmaintained, accumulating untriaged issues even to the point of [incompatibilities with its dependencies](https://github.com/scikit-optimize/scikit-optimize/issues/981) and [downgrades as the recommended course of action](https://github.com/scikit-optimize/scikit-optimize/issues/978#issuecomment-751972201)!

Having just recently evaluated several options in the Python hyperparameter optimization space, I found scikit-optimize the one to go with for one of my projects, so I have some direct interest in it remaining maintained and useful.

If you'll consider my application, please see examples of my [reasoning](https://github.com/scikit-optimize/scikit-optimize/pull/974), [work](https://github.com/scikit-optimize/scikit-optimize/pull/971), and [interaction](https://github.com/scikit-optimize/scikit-optimize/search?q=commenter%3Akernc&type=issues) already in the issue tracker, with more available obviously in other projects I work on.

As a new maintainer, I'd begin by:
* [ ] porting CI to GitHub Actions (can't have CI broken or taking hours to complete),
* [ ] triaging existing issues and PRs (assigning labels like you guys have till late 2017), _tidying up_ and hopefully making everyone again somewhat more interested in contributing.

Thanks for your consideration.

cc: @scikit-optimize, @betatim, @glouppe, @iaroslav-ai, @MechCoder, @holgern

betatim:
Hi @kernc 👋 

thanks for taking the time to find, evaluate and then start helping out here!

[...]

PR #1191, opened 2023-11-11
finish CI move to GHA

shimanek:
Just poking around trying to finish #1074 and maybe make progress toward #1117.  With this it seems like GitHub actions based CI is in a working state; publishing workflow obviously untested and so far remains unchanged from the draft by @kernc.  Relevant to recent discussion around #847.  Is this viable?  What's left? 

Passing checks here: https://github.com/shimanek/scikit-optimize/pull/4


.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+        if: env.TEST_WITH_COVERAGE == 'true'
+        run: |
+          make test-coverage
+          bash <(curl -s https://codecov.io/bash)
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: package
+          path: dist
```
xmatthias, line 80: I don't think this is correct.
At least one of the above steps MUST build the package, otherwise the dist directory is empty.


something around the following should work ...

```
      - name: Build package
        if: env.TEST_WITH_COVERAGE == 'true'
        run: |
          pip install --upgrade build
          python -m build --sdist --wheel --outdir dist/
```

PR #1192, opened 2023-11-24
Improved Bayesian Optimization documentation

bjiro:
Improved clarity on Bayesian Optimization presentation, so that transition from scikit-learn to scikit-optimize is smoother.


bjiro:
unsure what causes the ci to fail

